PR #29, opened 2025-04-16
Build as pip package

christian-byrne:
Changes structure to be installable as pip package.

Part of https://github.com/Comfy-Org/ComfyUI_frontend/issues/3478

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -2,10 +2,18 @@ name: Publish to PyPI
 
 on:
   workflow_dispatch:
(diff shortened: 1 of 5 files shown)


.github/workflows/publish.yml
```diff
@@ -0,0 +1,32 @@
+name: Publish to PyPI
+
+on:
+  workflow_dispatch:
```
christian-byrne, line 4: Let's run on PR if pyproject.toml has changed

huchenlei (approved):

PR #55, opened 2025-05-20
Add SDVN Example Workflow

phamhungd:
Some sample workflows using my SDVN Comfy Node set.


comfyui-wiki:
Thanks for your PR, but ComfyUI native template only allows using built-in nodes. In that case, users can directly run the workflow without any other custom nodes.

And the workflow should contain a model link as well. Please read the readme.

comfyui-wiki:
You can update the workflows in your custom node. Workflow, workflows, example, examples, these folder names can be detected by ComfyUI and directly displayed in the templates.

Since we can't make sure all the custom node authors can keep updating their custom nodes, keeping the workflow in the custom node folder is the recommended way.

PR #73, opened 2025-06-02
[fix] standardize model metadata format across all templates

### LINKED ISSUE #72: [Bug]: Missing model warning not always shown
### Frontend Version

Frontend version: 1.18.9. ComfyUI on master.

### Expected Behavior

When I load comfy ui for the first time, I see this warning: 

Image

If I don't check the "don't show this again" box and reload the page, I expect to see this error again.

### Actual Behavior

This error modal is no longer displayed. 

I didn't download the model on the first run, and on subsequent runs instead of seeing this helpful error I just saw this:

Image 

It's not obvious how to fix this error from this modal (took me a while to realize I did a fresh git clone and forgot to copy over the models). Ideally, we should show the "missing models" error again or something similar. 

Aside: It's also not obvious where v1-5-pruned-emaonly-fp16.safetensors came from if it wasn't already in `models/checkpoints` - maybe it was a hardcoded checkpoint name for the default workflow? 

### Steps to Reproduce

see above

### Debug Logs

```powershell
nothing special
```

### Browser Logs

Running on localhost:8188
App.vue:45 ComfyUI Front-end version: 1.18.9
:8188/favicon.ico:1 
            
            
           Failed to load resource: the server responded with a status of 404 (Not Found)Understand this error
:8188/api/userdata?dir=workflows&recurse=true&split=false&full_info=true:1 
            
            
           Failed to load resource: the server responded with a status of 404 (Not Found)Understand this error
:8188/api/userdata/comfy.templates.json:1 
            
            
           Failed to load resource: the server responded with a status of 404 (Not Found)Understand this error

### Setting JSON

{
    "Comfy.TutorialCompleted": true
}

### What browsers do you use to access the UI ?

Google Chrome

### Other Information

_No response_

┆Issue is synchronized with this [Notion page](https://www.notion.so/Issue-3811-Bug-Missing-model-warning-not-always-shown-1ec6d73d365081c6955cd2515c9e8ab2) by [Unito](https://www.unito.io)

christian-byrne:
Converts 22 workflow templates from inconsistent top-level models arrays to the standardized node properties format, resolving the model warning modal display inconsistency. Also enhances the validation script to prevent future format violations and detect missing widget values.

Fixes #72

### DIFF
--- a/scripts/validate_templates.py
+++ b/scripts/validate_templates.py
@@ -158,6 +158,107 @@ def check_required_thumbnails(index_data: List[Dict], templates_dir: Path) -> Tu
     return len(errors) == 0, errors
 
 
+def find_model_loader_nodes(template_data: Dict, models: List[Dict]) -> List[Tuple[str, str, str]]:
+    """Find nodes that use models from the models array in their widget values."""
+    model_nodes = []
+    model_names = {model.get('name', '') for model in models}
+    
+    # Look in the nodes array
+    nodes = template_data.get('nodes', [])
+    for node in nodes:
+        if isinstance(node, dict):
+            node_id = str(node.get('id', 'unknown'))
+            node_type = node.get('type', 'unknown')
+            widgets_values = node.get('widgets_values', [])
+            
+            # Check if any widget value matches a model name
+            for widget_value in widgets_values:
+                if isinstance(widget_value, str) and widget_value in model_names:
+                    model_nodes.append((node_id, node_type, widget_value))
+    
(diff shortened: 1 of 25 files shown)


comfyui-wiki:
For the recently updated workflows, I've been using my html tool to update the model links in the widget_values. I think the workflows I've newly updated shouldn't need any adjustments, right?

christian-byrne:
I added a test to ensure all workflows use the new format only, so all should be correct, otherwise the test will fail.

PR #101, opened 2025-06-27
Update the Flux.1 Kontext Dev basic template

comfyui-wiki:
Changed the default boolean value in widgets_values from true to false in flux_kontext_dev_basic.json. This may affect the initial state or behavior of the related widget.

### DIFF
--- a/templates/flux_kontext_dev_basic.json
+++ b/templates/flux_kontext_dev_basic.json
@@ -319,7 +319,7 @@
       },
       "widgets_values": [
         "ComfyUI_00018_.png [output]",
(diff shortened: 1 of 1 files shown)


christian-byrne:
Is it because the behavior is unexpected if you have not read what the button does first?

comfyui-wiki:
> Is it because the behavior is unexpected if you have not read what the button does first?

Yes, so I think I should change it to the default setting and add a note beside it. I will update it in the next release

comfyui-wiki:
Will merge after adding the note.

PR #180, opened 2025-08-22
Fix: `hidream_i1_fast.json` download URL mismatch

triestpa:
Simple typo fix - the `hidream_i1_fast.json` template has a duplicated t5xxxl model link included where there is supposed to be a llama 3.1 8b download link.


Robert904:
This error is also present in the "full" flavor of hidream
https://github.com/Comfy-Org/workflow_templates/blob/d1d4432d167ece7ac75352779be6c7e19d5f583b/templates/hidream_i1_full.json#L286

PR #181, opened 2025-08-25
Update i18n and Qwen image controlnet template

comfyui-wiki:
-  Update i18n index for new template modal
- Added multiple Qwen image controlnet templates
- Removed Kontext Grouped template


comfyui-wiki:
Still can't commit the fix successfully, emm, but the model check went well anyway.

PR #211, opened 2025-09-28
fix: update huggingface URL for WanAnimate_relight_lora_fp16.safetensors 

EmmanuelMr18:
The path for `WanAnimate_relight_lora_fp16.safetensors` changed in [this commit](https://huggingface.co/Kijai/WanVideo_comfy/commit/c047e06c231d7d978fa9ce775274a2afec58e418).

Current behavior without fixing it:

https://github.com/user-attachments/assets/4ce771fd-e44f-4b52-a895-5cf6754471a3

The big diff block is because it's updating the reference that exists in the markdown note. In reality, it is just this change:
CleanShot 2025-09-28 at 12 48 13


EmmanuelMr18:
Just to show that now works, and even the model size is correctly populated in the missing models UI.

CleanShot 2025-09-28 at 12 52 42

comfyui-wiki:
Thanks for the PR. Kijai changed the model path a few days ago, and I didn't notice that.

PR #246, opened 2025-10-30
Qwen T2I: switch LoRA to 8-step and update note

benceruleanlu:
See https://github.com/Comfy-Org/desktop/issues/1387

This removes references of qwen image edit 4 steps with lightning 8 steps.

The missing model modal did not show for this lora because references with 4 step and 8 step were mixed.

The diff does not like the markdown note's widget value, here is the richer diff:

image


benceruleanlu:
Unsure about the CI, seems unrelated

comfyui-wiki:
> Unsure about the CI, seems unrelated
The CI checks whether all input assets have been uploaded. Since I'm working alone on this repo, I didn't fix it. Haha

PR #249, opened 2025-11-02
[feat] Consolidate API templates into GENERATION TYPE categories

christian-byrne:
This PR migrates all 71 Partner Node (API) templates from CLOSED SOURCE MODELS categories into their corresponding GENERATION TYPE categories (Image, Video, 3D Model, Audio, LLMs). Each API template now has OpenSource: false field added, enabling the frontend to filter and display them in the new Partner Nodes virtual category. The CLOSED SOURCE MODELS category group has been removed entirely. Version bumped from 0.2.9 to 0.3.0 for PyPI release. This coordinates with frontend PR #6542 which implements the Partner Nodes filtering UI.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -4,7 +4,7 @@ build-backend = "setuptools.build_meta"
 
 [project]
 name = "comfyui_workflow_templates"
-version = "0.2.9"
+version = "0.3.0"
 description = "ComfyUI workflow templates package"
 readme = "README.md"
 requires-python = ">=3.9"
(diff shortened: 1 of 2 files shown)


comfyui-wiki:
I can do this on my end after the virtual category is supported in core. Also need to update the other languages' index files as well

comfyui-wiki:
Already moved all the API templates under the GENERATION TYPE (now called MODELS), so I'll close this PR.

PR #258, opened 2025-11-05
Validate model metadata within subgraphs; add iter_all_nodes helper

benceruleanlu:
This pull request improves the validation script to correctly handle models referenced inside subgraphs and tightens model metadata checks.

Changes
- Add `iter_all_nodes` helper to return a flat list of all nodes, including those defined under `definitions.subgraphs[].nodes`.
- Use the helper to:
  - Suggest appropriate model loader nodes when a template incorrectly uses top-level `models`.
  - Validate that each model listed in a node's `properties.models` has a corresponding entry in that node's `widgets_values`.
- Improve type annotations and error messages for clarity.

Why
Some workflows place loader nodes inside subgraphs. Previously, validation only considered top-level `nodes`, which could miss these cases.

Notes
- Only `scripts/validate_templates.py` is modified.
- Local run confirmed no regressions; the validation script runs cleanly.

If you prefer a different branch name, title, or PR wording, I can update accordingly.

### DIFF
--- a/scripts/validate_templates.py
+++ b/scripts/validate_templates.py
@@ -158,30 +158,49 @@ def check_required_thumbnails(index_data: List[Dict], templates_dir: Path) -> Tu
     return len(errors) == 0, errors
 
 
-def find_model_loader_nodes(template_data: Dict, models: List[Dict]) -> List[Tuple[str, str, str]]:
-    """Find nodes that use models from the models array in their widget values."""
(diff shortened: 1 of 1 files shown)


christian-byrne (approved):

christian-byrne:
Should we run it against the original bugged workflow to verify it works?

PR #331, opened 2025-11-26
Fix Hunyuan3D 2.1 description

waltercool:
Seems like Hunyuan3D 2.1 description is shown as 2.0.

Attaching commit renaming it as 2.1.


comfyui-wiki:
Thanks :)

PR #623, opened 2026-03-04
Add conditioning steering to z_image_turbo template

danielholanda:
# Description

This PR injects positive SFW conditioning into the generation pipeline via `ConditioningCombine`, reducing NSFW outputs without modifying the user-visible prompt.

New nodes (CLIPTextEncode + ConditioningCombine) are collapsed by default inside the subgraph to keep the UI clean.


comfyui-wiki:
TBH, I'm open to it since it's an open-source model, so I want to keep it more or less as it is.

PR #652, opened 2026-03-10
Add PostHog analytics to template site.

robinjhuang:
Adds PostHog event tracking to the site with separate project keys for staging vs production. Tracks run_button_clicked, template_viewed, download/share clicks, search, and filter actions using object_verb naming convention. Events sent to ph.comfy.org.

I added the prod project key to `.github/workflows/deploy-site.yml`


site/src/lib/posthog.ts
```diff
@@ -0,0 +1,98 @@
+import posthog from 'posthog-js';
+
+const POSTHOG_STAGING_KEY = 'phc_I86RrsKajBkKfL6z4ABf1uxnLtvO1NLZvAEOGUujSDD';
```
huntcsg: Can we do this at build time? And is that better?

PR #731, opened 2026-03-24
feat: exclude cloud-only templates from pip package distributions

comfyui-wiki:
## Problem

Templates that are either cloud-only or require custom nodes are useless to pip package / local users:
- **Cloud-only** (`includeOnDistributions: ["cloud"]`): the frontend already hides them at runtime, but they still ship in the wheel
- **Requires custom nodes** (`requiresCustomNodes` non-empty): pip users have no mechanism to auto-install custom nodes, so these templates would be broken out of the box

Both categories waste package space and pollute the local template browser.

## Solution

Filter them out **during sync** in `scripts/sync_bundles.py`, so published packages never contain them.

### What's excluded

| Reason | Count |
|--------|------:|
| Cloud-only (`includeOnDistributions: ["cloud"]`) | 11 |
| Requires custom nodes (`requiresCustomNodes` non-empty) | 62 |
| **Total excluded** | **73** |
| **Distributed to pip** | **346** (source: 419) |

Templates with `includeOnDistributions: ["local"]`/`["desktop"]` are kept — they're explicitly for local users.

### How it works

1. **`manifest.json`** — `build_manifest()` skips excluded templates before hashing assets, so the manifest and disk stay in sync (prevents `FileNotFoundError` at runtime)
2. **Workflow JSONs & thumbnails** — assets for excluded templates are not copied to package directories
3. **`index*.json` files** (primary + all 10 locale variants) — rewritten on the fly with excluded entries removed, preserving correct category structure
4. **Source files unchanged** — `templates/index.json` remains the source of truth; filtering only happens at sync/package-build time

### CI is automatic

`sync_bundles.py` is called by the `version-check` workflow on every PR touching `templates/**` or `bundles.json`, and the updated `manifest.json` is auto-committed back to the PR branch. No manual maintenance needed.

### Escape hatch — `--no-filter`

To revert the filtering without a code change, pass `--no-filter` to `sync_bundles.py`. The workflow has a comment showing exactly what to change:

```yaml
# To temporarily disable it and ship all templates, add --no-filter:
#   python scripts/sync_bundles.py --no-filter
python scripts/sync_bundles.py   # filtering ON by default
```

## Test plan

- [ ] Run `python scripts/sync_bundles.py` — output logs 73 excluded templates
- [ ] Verify `packages/*/src/*/templates/gsc_*.json` do not exist after sync
- [ ] Verify `manifest.json` does not contain any excluded template IDs
- [ ] Verify `packages/media_other/src/.../templates/index.json` has 346 templates (not 419)
- [ ] Verify no entries with `includeOnDistributions: ["cloud"]` remain in any distributed `index*.json`
- [ ] Verify templates with `includeOnDistributions: ["local"]` are still present
- [ ] Run `python scripts/sync_bundles.py --no-filter` — all 419 templates synced, manifest has 419 entries
- [ ] Confirm `packages/core` tests pass (`pytest packages/core`)

### DIFF
--- a/packages/core/tests/test_templates_repo.py
+++ b/packages/core/tests/test_templates_repo.py
@@ -1,6 +1,5 @@
 import importlib.util
 import json
-import sys
 from pathlib import Path
 
 REPO_ROOT = Path(__file__).resolve().parents[3]
@@ -66,7 +65,12 @@ def test_manifest_includes_all_template_assets():
         and path.suffix.lower() in SUPPORTED_EXTENSIONS
         # Exclude assets belonging to pip-filtered templates.
         # Asset filenames always start with the template name (e.g. "foo.json", "foo-1.webp").
-        and not any(path.name == ex + path.suffix or path.name.startswith(ex + "-") or path.name.startswith(ex + ".") for ex in excluded)
+        and not any(
+            path.name == ex + path.suffix
+            or path.name.startswith(ex + "-")
+            or path.name.startswith(ex + ".")
+            for ex in excluded
+        )
     }
     missing = sorted(files_on_disk - manifest_assets)
     assert not missing, f"Assets missing from manifest: {missing}"
--- a/scripts/sync_bundles.py
+++ b/scripts/sync_bundles.py
@@ -21,6 +21,7 @@ import hashlib
 import json
 import shutil
 from pathlib import Path
+from typing import Optional
 
 # Distribution values that are considered "local" (non-cloud).
 # A template is cloud-only when its includeOnDistributions is non-empty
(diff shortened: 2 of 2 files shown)


comfyui-wiki:
## Follow-up fixes (latest commit: e24b83e)

Two CI failures were caused by existing checks not knowing about the new pip-exclusion filter:

### 1. `validate` — orphaned template check
The shell loop in `validate-manifests.yml` flagged all 73 filtered-out templates as "orphaned" because they're still on disk but absent from the manifest. Fixed by calling the new `scripts/list_pip_excluded.py` helper to get the excluded names and skipping them in the loop.

### 2. `build-and-test` — `test_manifest_aligns_with_bundles_config`
The test asserted `manifest.keys() == bundles.json.keys()`, which broke because the manifest now intentionally omits pip-excluded templates. Fixed by importing `get_pip_excluded_template_names` from `sync_bundles.py` and removing excluded names from the expected set before comparing. Same fix applied to `test_manifest_includes_all_template_assets`.

### Additional improvements in this commit
- `sync_bundles.py`: `get_pip_excluded_template_names()` is now called **once** in `main()` and passed down — no more double reads of `index.json`
- `index_data_stems` hardcoded set replaced with a dynamic `glob("index.*.json")` scan — new locales are picked up automatically, `index.schema.json` is correctly excluded
- `scripts/list_pip_excluded.py`: standalone script used by CI (avoids embedding multi-line Python in YAML)

PR #775, opened 2026-04-10
fix(ci): reduce preview cron to 45min to avoid 429 rate limits

dante01yoon:
## Summary
- Changes preview cron schedule from every 15 minutes to every 45 minutes
- The 15-minute interval caused Vercel 429 rate limit errors on preview deployments due to the volume of matrix deploys (main prod + main test + labeled PRs)
- Production cron is unaffected — only preview environment hit the rate limit

Ref: https://github.com/Comfy-Org/workflow_templates/actions/runs/24222145508/job/70715913115

## Test plan
- [ ] Verify cron triggers every 45 minutes after merge
- [ ] Confirm no more 429 errors in preview cron runs

### DIFF
--- a/.github/workflows/preview-cron.yml
+++ b/.github/workflows/preview-cron.yml
@@ -2,7 +2,7 @@ name: Preview Cron
 
 on:
   schedule:
-    - cron: '*/15 * * * *'
+    - cron: '*/45 * * * *'
   workflow_dispatch:
 
 permissions:
--- a/.github/workflows/preview-cron.yml
+++ b/.github/workflows/preview-cron.yml
@@ -2,7 +2,10 @@ name: Preview Cron
 
 on:
   schedule:
-    - cron: '*/45 * * * *'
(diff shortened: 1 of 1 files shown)


Myestery (dismissed):

Myestery:
Note: */45 doesn't mean "every 45 minutes from last run" — it means "every minute divisible by 45", which gives you two hits per hour (0 and 45), not a true 45-minute interval. If you want a consistent 45-min gap, you'd need something like 0,45 * * * *, which is effectively the same here, but the distinction matters for values that don't divide 60 evenl

dante01yoon:
> Note: */45 doesn't mean "every 45 minutes from last run" — it means "every minute divisible by 45", which gives you two hits per hour (0 and 45), not a true 45-minute interval. If you want a consistent 45-min gap, you'd need something like 0,45 * * * *, which is effectively the same here, but the distinction matters for values that don't divide 60 evenl

applied good catch

PR #789, opened 2026-04-14
Fix search index duplicate ID crash and make prebuild failures fatal

dante01yoon:
## Summary
- **Search index duplicate ID fix**: Use `shareId` instead of `name` as MiniSearch document ID. The hub API can return multiple workflows with the same `name` from different users, causing `MiniSearch: duplicate ID` crash (e.g. `image_chrono_edit_14B`). `shareId` is the actual unique identifier.
- **Prebuild failure handling**: Make Phase 2 failures fatal (`process.exit(1)`) instead of non-fatal warnings. Previously a failed `build-search-index` would let the deploy continue with stale/missing data.
- **Search URL routing**: Store `slug` (`{name}-{shareId}`) in the search index so search results link to the correct workflow detail page.
- **Cron interval**: Change production rebuild cron to once per day at midnight UTC (`0 0 * * *`).
- **Preview cron interval**: Change preview rebuild cron to once per day at midnight UTC (`0 0 * * *`).

Fixes: https://github.com/Comfy-Org/workflow_templates/actions/runs/24384123386/job/71213859091

## Test plan
- [ ] Verify search index builds without duplicate ID errors
- [ ] Verify search results link to correct workflow pages (`/workflows/{name}-{shareId}/`)
- [ ] Verify prebuild exits with code 1 when any Phase 2 task fails
- [ ] Verify production cron schedule is `0 0 * * *`
- [ ] Verify preview cron schedule is `0 0 * * *`

### DIFF
--- a/site/scripts/lib/search/build-index.ts
+++ b/site/scripts/lib/search/build-index.ts
@@ -47,6 +47,11 @@ const MEDIA_TYPE_LABELS: Record<string, string> = {
   '3d': '3D',
 };
 
+/**
+ * Fetch workflow index entries from the hub API.
+ * Returns `null` when no API URL is configured (local dev).
+ * Throws when the API is configured but returns an error or empty response.
+ */
 async function fetchIndexEntries(): Promise<IndexEntry[] | null> {
   const apiUrl = (process.env.PUBLIC_HUB_API_URL || '').replace(/\/$/, '');
   if (!apiUrl) return null;
--- a/site/scripts/lib/search/build-index.ts
+++ b/site/scripts/lib/search/build-index.ts
@@ -20,6 +20,8 @@ interface SearchDocument {
   username: string;
   creatorName: string;
   // Stored fields for display (not searched)
+  name: string;
+  slug: string;
   thumbnail: string;
   usage: number;
   tagsArray: string[];
@@ -106,10 +108,15 @@ export async function buildSearchIndex(): Promise<void> {
       const creatorName = displayNames.get(username) || username;
       const tags = data.tags || [];
       const models = data.models || [];
-      const name = data.name || data.shareId || '';
+      const shareId = data.shareId || '';
+      const name = data.name || shareId;
+      // Use shareId as the unique ID; name can be duplicated across users
+      const id = shareId || name;
+      // URL slug: "{name}-{shareId}" for hub entries, "{name}" for legacy
+      const slug = shareId ? `${name}-${shareId}` : name;
 
       documents.push({
-        id: name,
+        id,
         title: data.title || name,
         description: data.description || '',
         tags: tags.map(tagSearchText).join(' '),
@@ -118,6 +125,8 @@ export async function buildSearchIndex(): Promise<void> {
         mediaTypeLabel: MEDIA_TYPE_LABELS[data.mediaType || ''] || data.mediaType || 'image',
         username,
         creatorName,
+        name,
+        slug,
         thumbnail: data.thumbnailUrl || '',
         usage: 0,
         tagsArray: tags.map(tagDisplayName),
@@ -137,9 +146,14 @@ export async function buildSearchIndex(): Promise<void> {
       const models: string[] = data.models || [];
       const thumbnails: string[] = data.thumbnails || [];
 
+      const shareId = data.shareId || '';
+      const name = data.name || shareId;
+      const id = shareId || name;
+      const slug = shareId ? `${name}-${shareId}` : name;
+
       documents.push({
-        id: data.name,
-        title: data.title || data.name,
+        id,
+        title: data.title || name,
         description: data.description || '',
         tags: tags.map(tagSearchText).join(' '),
         models: models.join(' '),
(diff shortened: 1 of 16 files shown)


site/src/components/hub/SearchPopover.vue
```diff
@@ -414,7 +414,7 @@ function activateItem(index: number) {
       if (creator) window.location.href = getCreatorUrl(creator.username);
     } else {
       const wf = displayedWorkflows.value[index - sugTotal - matchedCreators.value.length];
-      if (wf) window.location.href = getTemplateUrl(wf.id);
+      if (wf) window.location.href = getTemplateUrl(wf.slug);
```
christian-byrne, line 417: question: Popular workflow links use `wf.name` (here and line 728) while search result links were updated to use `wf.slug` (`{name}-{shareId}`). `SerializedTemplate` has `name` and `shareId` but no computed `slug` field. Does `extractShareId()` in the route handler handle both URL formats, or should popular workflows also use a slug?

site/src/lib/hub-api.ts
```diff
@@ -460,6 +460,32 @@ export function extractShareId(urlSegment: string): string | null {
...
+    return entries.map((e) => serializeIndexEntry(e, profiles));
+  } catch (err) {
+    if (import.meta.env.PUBLIC_HUB_API_URL) {
+      throw new Error(`Hub API failed during build: ${err}`);
+    }
+    console.warn('No hub API configured, using content collection for local build');
+    const templates = await getCollection();
+    return templates
+      .filter((tmpl) => !tmpl.id.includes('/'))
+      .sort((a, b) => ((b.data as any).usage || 0) - ((a.data as any).usage || 0))
```
christian-byrne: issue: (non-blocking) `(b.data as any).usage` -- the `as any` is unnecessary here. The `data` parameter is typed as `Parameters<typeof serializeCollectionEntry>[0]` which already includes `usage?: number` (line 329). Can be `.sort((a, b) => (b.data.usage || 0) - (a.data.usage || 0))`.

site/src/lib/hub-api.ts
```diff
@@ -460,6 +460,32 @@ export function extractShareId(urlSegment: string): string | null {
...
+): Promise<SerializedTemplate[]> {
+  const profiles = await getProfileCache();
+  try {
+    const entries = await listWorkflowIndex();
+    return entries.map((e) => serializeIndexEntry(e, profiles));
+  } catch (err) {
+    if (import.meta.env.PUBLIC_HUB_API_URL) {
+      throw new Error(`Hub API failed during build: ${err}`);
+    }
+    console.warn('No hub API configured, using content collection for local build');
```
christian-byrne: suggestion: (non-blocking) The fallback `console.warn` says "No hub API configured" but the catch could fire for other reasons. Consider: `console.warn('Hub API error, falling back to content collection:', err)` to preserve diagnostics.

site/scripts/lib/search/build-index.ts
```diff
@@ -48,22 +50,21 @@ const MEDIA_TYPE_LABELS: Record<string, string> = {
 async function fetchIndexEntries(): Promise<IndexEntry[] | null> {
```
christian-byrne, line 55: suggestion: (non-blocking) `fetchIndexEntries()` returns `null` when no API URL is set but throws on empty response or HTTP error. A brief JSDoc noting "returns null = no API configured; throws = API broken" would help future maintainers.

site/src/pages/[locale]/workflows/category/[type].astro
```diff
@@ -48,24 +48,8 @@ const categoryName = t(categoryTranslationKeys[mediaType], typedLocale);
 const title = `${categoryName} Workflows - ${t('meta.title', typedLocale)}`;
 const description = t('meta.description', typedLocale);
```
Comment (christian-byrne) [inline site/src/pages/[locale]/workflows/category/[type].astro:49]: nitpick: (non-blocking) `allSerialized.filter((t) => ...)` -- `t` shadows the imported `t()` i18n function. Consider `tmpl` as the parameter name.

site/src/components/hub/SearchPopover.vue
dante01yoon: Good catch — popular workflows were using `wf.name` which hits the legacy 301 redirect path. Fixed both line 426 and line 728 to use `\`${wf.name}-${wf.shareId}\`` so they link directly to the canonical URL without the redirect hop.

site/src/lib/hub-api.ts
dante01yoon: Fixed — removed the `as any` casts. The `data` parameter type already includes `usage?: number` via `Parameters<typeof serializeCollectionEntry>[0]`.

site/src/lib/hub-api.ts
dante01yoon: Good point — updated to `console.warn('Hub API error, falling back to content collection:', err)` so the actual error is preserved in the logs.

site/scripts/lib/search/build-index.ts
dante01yoon: Added JSDoc clarifying the contract: returns `null` when no API URL is configured, throws when the API is configured but returns an error or empty response.

site/src/pages/[locale]/workflows/category/[type].astro
Comment (dante01yoon) [reply site/src/pages/[locale]/workflows/category/[type].astro:49]: Fixed — renamed to `tmpl` in both the locale and non-locale category pages.

christian-byrne (approved):
LGTM

PR #792, opened 2026-04-14
Update ERNIE-Image templates

alexisrolland:
@comfyui-wiki update workflow examples:

- Added Load CLIP to use the PE model
- Added conversion nodes for width, height, and concatenate them with the existing prompt
- Set prompt enhancement to true by default
- Updated the Markdown notes
- Also updated the seed value for `image_ernie_image.json` because it had a bad one


comfyui-wiki:
The workflow needs to embed the model link, so I have updated the workflow on my end. Thanks for the PR :)

PR #812, opened 2026-04-21
fix(site): render video previews in Comfy Hub search (FE-235)

dante01yoon:
## Summary

Search popover thumbnails went blank for templates whose primary asset is a `.mp4`/`.mov`. `HubWorkflowCard` already handled this via `getVideoFrameUrl()` + `<video>` fallback — `SearchPopover` had drifted.

Reported in Slack by Denys: https://comfy-organization.slack.com/archives/C0A4XMHANP3/p1776786515961529
Linear: [FE-235](https://linear.app/comfyorg/issue/FE-235/search-results-missing-workflow-previews-on-comfy-hub)

## Root cause

`site/src/components/hub/SearchPopover.vue` had two broken thumbnail render paths for video files:

1. **Popular Workflows** — `getPrimaryThumb()` returned `null` for video files, and the template had no `<video>` fallback, so the tile rendered empty.
2. **Workflow search results** — rendered `` directly (only audio was special-cased), so the browser showed a broken image.

Templates affected include `Seedance 2.0: Text to Video`, `Reference to Video`, `First-Last-Frame to Video` — their primary thumbnails live on `comfy-hub-assets.comfy.org`, which isn't on the `engcomfy.com` Cloudflare CDN, so `getVideoFrameUrl()` returns `null` and there's no poster frame available.

## Changes

- **`SearchPopover.vue`**: import `getVideoFrameUrl` and add `getImageThumb` / `videoThumbUrl` helpers. Both the Popular Workflows and Workflow search results sections now prefer the Cloudflare poster frame when available, and fall back to `<video preload="metadata">` (first-frame render) otherwise. Mirrors `HubWorkflowCard`'s pattern.
- **`media-utils.ts`**: strip `?query` / `#hash` and lowercase before the extension check so `clip.mp4?token=abc` and `CLIP.MP4` classify as video. Pure hardening — no regressions for image/audio callers.
- **`tests/unit/media-utils.test.ts`**: new — covers signed URLs, uppercase, hash fragments, and the "video name only in query string" negative case.

No behavior change for image/audio thumbnails. No new dependencies.

## Test plan

- [x] `pnpm test` — 89/89 passing (added 9 new `media-utils` cases)
- [x] `pnpm run check` (astro / vue-tsc) — 0 errors / 0 warnings from this change
- [x] `pnpm exec prettier --check` — clean on changed files
- [ ] Manual browser check: open Comfy Hub search, confirm video templates render a preview in both the Popular Workflows list and the Workflow search results list
- [ ] Visual-regression / e2e CI


site/src/components/hub/SearchPopover.vue
```diff
@@ -505,14 +506,30 @@ function formatUsage(usage: number): string {
...
+}
+
+// Returns the full video URL when the thumbnail is a video file, otherwise null.
+// Used to drive a <video> fallback for non-CDN videos where the browser
+// renders the first frame for free via preload="metadata".
+function videoThumbUrl(file: string | undefined | null): string | null {
+  if (!file || !isVideoFile(file)) return null;
+  return resolveThumbUrl(file);
+}
+
```
christian-byrne: Why not use jsdoc method comments?

christian-byrne (dismissed):
LGTM with single non-blocking nit

site/tests/unit/media-utils.test.ts
```diff
@@ -0,0 +1,53 @@
...
+  it('ignores extension case', () => {
+    expect(isVideoFile('CLIP.MP4')).toBe(true);
+    expect(isVideoFile('Clip.Mov')).toBe(true);
+  });
+
+  it('ignores query strings and hash fragments on signed URLs', () => {
+    expect(isVideoFile('https://cdn.example.com/clip.mp4?token=abc')).toBe(true);
+    expect(isVideoFile('https://cdn.example.com/clip.mp4#t=5')).toBe(true);
+    expect(isVideoFile('https://cdn.example.com/clip.mp4?x=1#t=5')).toBe(true);
+  });
```
DrJKL: Want to parameterize these with `it.for`?

site/src/components/hub/SearchPopover.vue
DrJKL: I'm generally opposed to comments, but if we're going to have block comments I agree with @christian-byrne that they should use the standard.

dante01yoon: I agree JSdoc would be more pretty

site/tests/unit/media-utils.test.ts
dante01yoon: done

site/tests/unit/media-utils.test.ts
```diff
@@ -0,0 +1,41 @@
...
+    ['clip.mov', true],
+    ['CLIP.MP4', true],
+    ['Clip.Mov', true],
+    ['https://cdn.example.com/clip.mp4?token=abc', true],
+    ['https://cdn.example.com/clip.mp4#t=5', true],
+    ['https://cdn.example.com/clip.mp4?x=1#t=5', true],
+    ['https://cdn.example.com/image.webp?fallback=clip.mp4', false],
+    ['thumb.webp', false],
+    ['song.mp3', false],
+  ] as [string, boolean][])('isVideoFile(%s) → %s', ([filename, expected]) => {
```
DrJKL, line 16: Or does it.for take a generic?
```suggestion
  ] satisfies [string, boolean][])('isVideoFile(%s) → %s', ([filename, expected]) => {
```

site/tests/unit/media-utils.test.ts
DrJKL, line 1: Nice :-D

Issue #786, opened 2026-04-14
Remove dead /api/vitals beacon — 66K wasted 404s/day


christian-byrne:
## Problem

`site/src/scripts/vitals.ts` fires `navigator.sendBeacon("/api/vitals")` on every page load (6 Web Vitals metrics per visit: CLS, FCP, FID, INP, LCP, TTFB), but **no API route handler exists** to receive the data. Every beacon results in a 404.

This generates ~66K wasted requests/day (~11K page visits × 6 metrics). These requests hit the comfy-router Cloudflare Worker, get routed to Framer (catch-all `/*` rule), and 404 there.

## Root Cause

- `site/src/layouts/BaseLayout.astro` imports `../scripts/vitals` on every page
- `site/src/scripts/vitals.ts` calls `navigator.sendBeacon("/api/vitals", body)` in production
- No corresponding `site/src/pages/api/vitals.ts` endpoint was ever created
- The `web-vitals` package is listed in `site/package.json` but the data is never collected

## This Data Is Already Collected Elsewhere

`@vercel/analytics` (via `<Analytics />` in `BaseLayout.astro`) **already collects Core Web Vitals automatically** and sends them to `vitals.vercel-insights.com`. The custom `vitals.ts` script is fully redundant — the Vercel dashboard already has real-user CLS, FCP, INP, LCP, and TTFB data.

Lighthouse CI (in `site-ci.yml` / `lighthouserc.json`) handles CI-time performance budgets separately and is unrelated to this script.

## Fix

Remove the dead code:
1. Delete `site/src/scripts/vitals.ts`
2. Remove the `import("../scripts/vitals")` from `site/src/layouts/BaseLayout.astro`
3. Remove `web-vitals` from `site/package.json`
4. Update `site/docs/design-integration-guide.md` — remove `vitals.ts` from the "Protected Components" table and the preservation checklist

## Context

Discovered during a security audit of the Cloudflare setup. See [comfy-router](https://github.com/Comfy-Org/comfy-router) for the Worker routing logic — `/api/vitals` does not match any Vercel route prefix so it falls through to Framer.

PR #828, opened 2026-04-29
fix(site): guard workflow URLs against undefined slugs (FE-223)

dante01yoon:
## Problem

PostHog shows two related bugs producing **~2,015 broken pageviews/week** on `www.comfy.org` / `comfy.org`, with the broken URL traffic flat-to-rising over the past 14 days.

| URL pattern | 7d views | What's undefined |
|---|---|---|
| `/workflows/undefined/` | 1,771 | both `name` and `shareId` |
| `/workflows/video_wan2_2_14B_i2v-/` | 135 | only `shareId` (note trailing dash) |
| `/workflows/api_seedance2_0_r2v-/` | 52 | only `shareId` |
| `/workflows/api_grok_image_edit-/` | 31 | only `shareId` |
| `/workflows/api_seedance2_0_t2v-/` | 22 | only `shareId` |
| 2 more `*-/` slugs | 4 | only `shareId` |

Google indexed `/workflows/undefined/` and is referring **570 views/wk** — the page returns HTTP 200 with `<title>undefined (@undefined) - ComfyUI Workflow Templates</title>`.

## Root causes

Five places, layered:

1. `src/components/hub/HubWorkflowCard.vue:97-101` — when `props.name` was undefined, the template literal stringified it to literal `"undefined"`, and `handleCardClick` hard-navigated users to `/workflows/undefined/`.
2. `src/pages/workflows/[slug].astro:39-55` — `getStaticPaths` built a static page even when `entry.shareId` (or `entry.name`) was missing, producing slugs like `<name>-/` with empty suffix. Astro pre-rendered these as indexable HTML returning 200.
3. `src/pages/workflows/[username].astro` and `src/pages/[locale]/workflows/[slug].astro` — once the static page fix above stops generating the bad files, requests fall through to the on-demand creator-profile route. That route accepted any slug, fell back to `displayName = slugParam`, and rendered a profile page titled `<slug> (@<slug>)`. The existing redirect guard `!displayName && serialized.length === 0` never fired because `displayName` always falls back to the slug itself.
4. `src/components/hub/SearchPopover.vue:746-750, 1109` — the navbar search island built popular-result hrefs by interpolating `wf.name`/`wf.shareId` directly, and the badge-only mapped hits omitted the `slug` field entirely. A malformed index row could still produce `/workflows/undefined-<id>/`, `/workflows/<name>-undefined/`, or `/workflows/<name>-/` from the search popover. _(Surfaced by Codex adversarial review.)_
5. `scripts/lib/search/build-index.ts` — the prebuilt MiniSearch index used `name = data.name || shareId` and `slug = shareId ? \`${name}-${shareId}\` : name`, so a hub entry missing `name` was silently rebuilt as `<shareId>-<shareId>` and a name-only entry with empty `data.name` produced an empty slug. _(Surfaced by Codex adversarial review.)_

## Fix

- New `workflowDetailPath(name, shareId?, locale?)` helper in `src/lib/routes.ts` returning `null` when `name` is missing, plus a sibling `workflowDetailSlug(name, shareId?)` returning just the slug component for callers (search index, badge-only hits) that don't need the full URL.
- `HubWorkflowCard.vue` uses the helper; click handler bails out when null; `cursor-pointer` is conditional so cards without a target URL aren't visually clickable.
- `[slug].astro` `getStaticPaths` skips entries missing `name` or `shareId` (with `console.warn`); the content-collection fallback also filters by `name`.
- `[username].astro` and `[locale]/workflows/[slug].astro` track `profileFound` explicitly and redirect to the gallery when the slug isn't a detail page, isn't a known profile, and has no workflows. Creator-grid filters now also drop rows missing `name`/`shareId` so cards never render with a broken target.
- `SearchPopover.vue`: added `shareId` to the local `SearchTemplate` interface (was missing despite being read at runtime); routed popular-workflow hrefs and click handlers through `workflowDetailPath`; filtered `popularWorkflows` to drop rows lacking either field; fixed the badge-only `displayedWorkflows` mapping to attach a real `slug` (the missing field was its own latent bug).
- `build-index.ts`: skip hub entries lacking `name` or `shareId` before adding to the search index (with `logger.warn`); content-collection fallback drops nameless rows. Strict input contract means the prebuilt index can never contain `undefined-...` or empty slugs.

The strategy on shareId-only entries is "filter everywhere consistently" rather than "preserve via fallback slug" — PostHog data shows zero shareId-only URL traffic in production, so a name-less workflow has no human-readable display anyway. Filtering at every entry point (cards, search index, grids, static paths) guarantees a malformed row can't surface as a broken link from any UI surface.

## Behavioural impact after deploy

- New static-build run drops indexable `/workflows/<name>-/` pages → those URLs become real 404s, which Google will eventually drop from the index. A Google Search Console "Removals" request can speed this up.
- Card clicks for malformed entries become no-ops instead of `/workflows/undefined/` navigation.
- Bogus slugs that bleed through to the on-demand creator-profile route now redirect to `/workflows/` instead of rendering as fake profiles.
- Search popover popular-workflows rows and badge-only hits never render undefined-slug hrefs.
- Search index size shrinks slightly (skipped hub entries logged as warnings during build).

### Unit tests

- `pnpm test` — **106 unit tests pass** (8 in `tests/unit/routes.test.ts` covering both helpers, all branches, including the null-on-missing-name guard and shareId-only edge cases).
- Prettier clean on touched files.
- ESLint clean on touched files (5 pre-existing repo-wide errors are unchanged versus `main`).
- `pnpm check` — 0 errors, 0 warnings.

### Vercel preview probes (commit `b0b1f13`, preview `workflow-templates-8tvpu4abo-comfyui.vercel.app`)

| URL | Before | After | Expected | Notes |
|---|---|---|---|---|
| `/workflows/undefined/` | 200 | 302 → `/workflows/` | 302 | ✅ Bug fixed |
| `/workflows/video_wan2_2_14B_i2v-/` | 200 | 302 → `/workflows/` | 302 | ✅ Bug fixed |
| `/workflows/api_grok_image_edit-/` | 200 | 302 → `/workflows/` | 302 | ✅ Bug fixed |
| `/workflows/api_seedance2_0_r2v-/` | 200 | 302 → `/workflows/` | 302 | ✅ Bug fixed |
| `/workflows/api_seedance2_0_t2v-/` | 200 | 302 → `/workflows/` | 302 | ✅ Bug fixed |
| `/workflows/comfyui/` (real creator) | 200 | 200 | 200 | ✅ Real creator profile preserved |
| `/workflows/` (gallery) | 200 | 200 | 200 | ✅ Gallery loads |
| `/workflows/api_seedance2_0_r2v-64f4db9e3e33/` (real detail) | 200 | 302 → `/workflows/` | 200 | ⚠ Preview Hub API returned no workflow entries at build time, so `[slug].astro` couldn't generate static detail pages. Falls through to `[username].astro`'s pre-existing detail-catch redirect (line 64, unchanged by this PR). Confirmed prod still 200s these URLs. Production build will populate detail pages from the real Hub API and serve them statically — preview false negative on this row only. |

### Search index audit (commit `b0b1f13`)

Pulled the live preview's `/workflows/search-index.json` and audited:

```
total docs: 420
bad slug (empty/undefined): 0
bad name (empty/undefined): 0
slug ending in -: 0
```

Zero malformed slugs reach the client search index, confirming the `build-index.ts` filter and that the SearchPopover popular/badge-only paths can never receive undefined-slug rows.

### Production validation (post-deploy)

Plan to re-run the same probe set against `www.comfy.org` after merge + deploy. Expected:

```bash
curl -o /dev/null -w "%{http_code}\n" https://www.comfy.org/workflows/undefined/
# expect: 302 (currently 200)

curl -o /dev/null -w "%{http_code}\n" https://www.comfy.org/workflows/video_wan2_2_14B_i2v-/
# expect: 302 (currently 200)

curl -o /dev/null -w "%{http_code}\n" https://www.comfy.org/workflows/api_seedance2_0_r2v-64f4db9e3e33/
# expect: 200 (must still serve)
```

PostHog query 24h post-deploy:
```sql
SELECT toDate(timestamp) AS day, properties.$pathname AS path, count() AS views
FROM events
WHERE event = '$pageview'
  AND timestamp >= now() - interval 3 day
  AND (properties.$pathname = '/workflows/undefined/' OR properties.$pathname LIKE '/workflows/%-/')
GROUP BY day, path ORDER BY day DESC, views DESC
```
Expected: `/workflows/undefined/` and `*-/` views drop toward 0 by next day.

## Notes

This bug requires malformed upstream data to repro in a browser, so no clean before/after browser screenshot — the unit tests + preview probes + search-index audit cover the behaviour. Google Search Console "Removals" request for `/workflows/undefined/` after deploy will speed up de-indexing.

Codex adversarial review pass on commit `00a520d` flagged two additional leaks (SearchPopover URL builders and shareId-only routing inconsistency), addressed in commit `b0b1f13` and re-verified against the new preview.

Fixes FE-223

### DIFF
--- a/site/scripts/lib/search/build-index.ts
+++ b/site/scripts/lib/search/build-index.ts
@@ -57,9 +57,7 @@ async function fetchIndexEntries(): Promise<IndexEntry[] | null> {
   if (!apiUrl) return null;
 
   const approvedOnly = process.env.PUBLIC_APPROVED_ONLY === 'true';
-  const statuses = approvedOnly
-    ? 'approved'
-    : 'pending,approved,rejected,deprecated';
+  const statuses = approvedOnly ? 'approved' : 'pending,approved,rejected,deprecated';
   const res = await fetch(`${apiUrl}/api/hub/workflows/index?status=${statuses}`);
 
   if (!res.ok) {
@@ -107,17 +105,23 @@ export async function buildSearchIndex(): Promise<void> {
       }
     }
 
+    let skipped = 0;
     for (const data of hubEntries) {
+      // Drop entries that can't form a clean detail URL. Without this, the
+      // search index emits slugs like "undefined-<id>" or "<name>-undefined"
+      // and the popover renders /workflows/undefined/ links (FE-223).
+      if (!data.name || !data.shareId) {
+        skipped++;
+        continue;
+      }
       const username = data.username || 'ComfyUI';
       const creatorName = displayNames.get(username) || username;
       const tags = data.tags || [];
       const models = data.models || [];
-      const shareId = data.shareId || '';
-      const name = data.name || shareId;
+      const { name, shareId } = data;
       // Use shareId as the unique ID; name can be duplicated across users
-      const id = shareId || name;
-      // URL slug: "{name}-{shareId}" for hub entries, "{name}" for legacy
-      const slug = shareId ? `${name}-${shareId}` : name;
+      const id = shareId;
+      const slug = `${name}-${shareId}`;
 
       documents.push({
         id,
@@ -136,22 +140,33 @@ export async function buildSearchIndex(): Promise<void> {
         tagsArray: tags.map(tagDisplayName),
       });
     }
+    if (skipped > 0) {
+      logger.warn(`Skipped ${skipped} hub entries missing name or shareId — see FE-223`);
+    }
   } else {
     // No PUBLIC_HUB_API_URL — local/offline build, use content collection
-    const files = fs.readdirSync(CONTENT_DIR).filter((f) => f.endsWith('.json') && !f.includes('/'));
-    logger.warn(`No hub API configured — building search index from content collection (${files.length} files)`);
+    const files = fs
+      .readdirSync(CONTENT_DIR)
+      .filter((f) => f.endsWith('.json') && !f.includes('/'));
+    logger.warn(
+      `No hub API configured — building search index from content collection (${files.length} files)`
+    );
 
     for (const file of files) {
       const filePath = path.join(CONTENT_DIR, file);
       const data = JSON.parse(fs.readFileSync(filePath, 'utf-8'));
+
+      // Legacy content-collection templates always have a name but may omit
+      // shareId. Drop nameless rows so we never index an empty slug (FE-223).
+      if (!data.name) continue;
 
       const username = data.username || 'ComfyUI';
       const tags: string[] = data.tags || [];
       const models: string[] = data.models || [];
       const thumbnails: string[] = data.thumbnails || [];
 
-      const shareId = data.shareId || '';
-      const name = data.name || shareId;
+      const shareId: string = data.shareId || '';
+      const name: string = data.name;
       const id = shareId || name;
       const slug = shareId ? `${name}-${shareId}` : name;
 
@@ -229,5 +244,7 @@ export async function buildSearchIndex(): Promise<void> {
 
   const sizeKB = (Buffer.byteLength(serialized) / 1024).toFixed(1);
   const duration = ((Date.now() - startTime) / 1000).toFixed(2);
-  logger.info(`Search index written to public/workflows/search-index.json (${sizeKB} KB, ${duration}s)`);
+  logger.info(
+    `Search index written to public/workflows/search-index.json (${sizeKB} KB, ${duration}s)`
+  );
 }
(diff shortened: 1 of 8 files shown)


dante01yoon:
Closing in favor of #830, which carries the same code fix but with an accurate root-cause description.

The original PR description here blamed `HubWorkflowCard.vue`'s template literal — that code path is theoretically reachable but Hub API never sends `name=undefined` in production, so it doesn't fire. PostHog session replay (`019ddaec-7f57-7ce8-a551-b17d132be449`) pinpointed the actual active leak: `SearchPopover.vue`'s badge-only `displayedWorkflows` mapping is missing the `slug` field, and `getTemplateUrl(undefined)` produces `/workflows/undefined/`. Timeline matches PR #786 (slug propagation, missed the badge-only branch) → PR #789 (deployed it) → first event at 2026-04-16 03:32 UTC, 10 minutes after #789 merged.

The corrected diagnosis and verification plan live on #830 and Linear FE-223. Closing this PR for review-history clarity.

PR #856, opened 2026-05-12
feat(site): hide free tier behind cloudFreeTier feature flag

christian-byrne:

---

## Summary

Free-tier signups are paused after credit abuse. Hide free-tier promotional copy on templates.comfy.org via a new build-time JSON feature flag, mirroring the pattern in `comfyui_frontend/apps/website` ([Comfy-Org/ComfyUI_frontend#12170](https://github.com/Comfy-Org/ComfyUI_frontend/pull/12170)).

## Feature flag

- `site/src/config/featureFlags.json` — single `cloudFreeTier` boolean (default `false`).
- `site/src/config/featureFlags.ts` — typed loader using ES2024 `import … with { type: 'json' }`. Result is `Object.freeze`d.
- `site/src/config/featureFlags.schema.json` — JSON schema for editor autocomplete + validation.

Flag name mirrors `remoteConfig.new_free_tier_subscriptions` in the main Vue app so the two systems can be unified later.

## What changes when `cloudFreeTier = false`

| Surface | Flag ON (today) | Flag OFF (this PR's default) |
|---|---|---|
| Hub navbar CTA (`HubNavbar.astro`) | "Try Comfy Cloud for free" | "Try Comfy Cloud" |
| Template-detail Try-on-Cloud CTA subline (`TemplateDetailPage.astro`) | "✦ Get Free Credits" (en) + 10 localized variants | Subline `<span>` omitted entirely |

The detail-page subline gate is at the component level, so all 11 locales (en, zh, zh-TW, ja, ko, es, fr, ru, tr, ar, pt-BR) are covered without touching `i18n/ui.ts`.

## What I intentionally left alone

- `meta.description` "Browse free, ready-to-use ComfyUI workflows" — describes the open-source templates, not the cloud free tier.
- SEO titles like "Free AI Generation Workflows" / "Free Templates" — same reason.
- `knowledge/models/vidu.md` "free tier available" mention — internal docs, not user-facing.

If you want any of these stripped too, easy follow-up.

## Verification

- `pnpm test` — 112 tests pass (4 new for the flag loader).
- `pnpm check` (astro check) — 0 errors, 0 warnings, 10 pre-existing hints unrelated to this PR.
- `pnpm lint` — my files clean. The 5 pre-existing errors (`SerializedTemplate` unused in 4 page files, `prefer-rest-params` in `BaseLayout.astro`) are not in files I changed.
- `pnpm exec prettier --check` on changed files — clean.
- `pnpm exec astro build` — succeeds in both flag states. Built HTML verified via grep: flag-off has zero "Get Free Credits" / "for free" occurrences; flag-on shows them in the expected spots.
- Live Playwright preview confirmed navbar reads "Try Comfy Cloud" and template detail CTA has no subline when flag is off.

## Re-enabling free tier

Flip `cloudFreeTier` to `true` in `featureFlags.json` and rebuild. No other changes required.

## Related

- Companion PR in the marketing site: [Comfy-Org/ComfyUI_frontend#12170](https://github.com/Comfy-Org/ComfyUI_frontend/pull/12170).
- Mirrors `remoteConfig.new_free_tier_subscriptions` in the main app for eventual unification.

## Screenshots

templates.comfy.org workflow hub with cloudFreeTier=false: navbar CTA reads 'Try Comfy Cloud' (no 'for free').

templates.comfy.org template detail page with cloudFreeTier=false: Try-on-Cloud CTA has no '✦ Get Free Credits' subline.


site/tests/unit/feature-flags.test.ts
```diff
@@ -0,0 +1,21 @@
...
+  it('mirrors the cloudFreeTier value from featureFlags.json', () => {
+    expect(featureFlags.cloudFreeTier).toBe(rawFlags.cloudFreeTier);
+  });
+
+  it('exposes cloudFreeTier as a boolean', () => {
+    expect(typeof featureFlags.cloudFreeTier).toBe('boolean');
+  });
+
+  it('defaults cloudFreeTier to false while signups are paused', () => {
+    expect(featureFlags.cloudFreeTier).toBe(false);
```
dante01yoon: nit-pick (non-blocking): this assertion ties the test suite to the flag's current value, so the documented re-enable path ("flip `cloudFreeTier` to `true` in `featureFlags.json` and rebuild") will fail this test until someone also updates the expectation here. The first test already verifies the loader mirrors JSON. Consider dropping this case or rephrasing it as a comment-only reminder, so flipping the flag is a one-file change.

site/src/config/featureFlags.ts
```diff
@@ -0,0 +1,9 @@
+import rawFlags from './featureFlags.json' with { type: 'json' };
+
+interface FeatureFlags {
+  readonly cloudFreeTier: boolean;
+}
+
+const { cloudFreeTier } = rawFlags;
```
dante01yoon: suggestion (non-blocking): the explicit destructure-then-reconstruct means any new flag added to `featureFlags.json` won't be exposed on `featureFlags` until this file is also updated. That's arguably a feature (explicit allowlist matching the `FeatureFlags` interface), but if the intent is for the JSON to be the source of truth, `Object.freeze({ ...rawFlags } satisfies FeatureFlags)` would catch divergence at the type level without a manual destructure. Either way is reasonable — flagging in case the intent is JSON-as-source-of-truth.

dante01yoon (dismissed):
Approving. Clean, well-scoped feature flag with the gate placed at the component level so all 11 locales are covered without touching `i18n/ui.ts`. Verification (build in both flag states, grep of built HTML, Playwright preview) is thorough.

A couple of non-blocking notes left inline. Confirmed:

- Hub navbar CTA: when flag is off, the bare `Try Comfy Cloud` button still has only one centered span; `flex flex-col` works fine.
- Detail-page CTA: hiding the subline `<span>` leaves the bold `cta.tryCloud` span centered inside the same `flex flex-col py-2` container — visually consistent with the screenshots.
- `TryOnCloudButton.astro` (template card variant) doesn't carry free-tier copy, so it correctly stays out of the gate.
- Pre-existing `meta.description` ("free, ready-to-use ComfyUI workflows") and `knowledge/models/vidu.md` mention of "free tier" are intentionally left alone per PR description — both describe the templates / model rather than the cloud free-tier subscription.
- The hub navbar string remains hardcoded English, but that matches main's prior behavior — no regression.

christian-byrne:
Refactored it more to put incremental feature flag snapshoting into the CI/build.

site/src/utils/featureFlags.ci.ts
```diff
@@ -0,0 +1,90 @@
...
+
+import type { FetchOutcome } from './featureFlags';
+
+let hasReported = false;
+
+export function resetFeatureFlagsReporterForTests(): void {
+  hasReported = false;
+}
+
+export function reportFeatureFlagsOutcome(outcome: FetchOutcome): void {
```
dante01yoon, line 11: question (non-blocking): `reportFeatureFlagsOutcome` is exported here but I don't see any importer in the PR — not from astro.config.mjs, not from `.github/workflows/*.yml`, not from any other script. So as it stands, this 90-line CI reporter is dead code at the PR's head.

Similarly, `fetchFeatureFlagsForBuild` in `featureFlags.ts` is only invoked from the manual `refresh-feature-flags-snapshot.ts` script and the unit tests — not from any build step.

Is the CI workflow that calls these (e.g., a `site-ci.yml` step that runs `feature-flags:refresh-snapshot` and pipes the outcome into `reportFeatureFlagsOutcome`) landing in a follow-up PR? If so, totally fine — flag it in the description so reviewers know the runtime path will be wired up next. If not, the runtime story is currently "snapshot updates require a manual `pnpm feature-flags:refresh-snapshot` + commit + PR + merge", which is a meaningfully different cadence than what the PR body implies and would also leave the reporter file unused.

site/src/config/featureFlags.ts
```diff
@@ -0,0 +1,9 @@
+import snapshot from '../data/feature-flags.snapshot.json' with { type: 'json' };
+
+interface FeatureFlags {
+  readonly cloudFreeTier: boolean;
+}
+
+export const featureFlags: FeatureFlags = Object.freeze({
+  cloudFreeTier: snapshot.flags.cloudFreeTier,
```
dante01yoon, line 8: question (related to the comment on `featureFlags.ci.ts`): the synchronous consumer reads from the committed `feature-flags.snapshot.json` directly, so the in-prod flag state at any given build is whatever was last committed to that file. That's fine, but unless the refresh script gets wired into CI (or a deploy hook), this isn't really a build-time feature flag system in the dynamic sense — it's effectively a typed boolean constant that happens to live in JSON. Not a blocker, but worth being explicit in the PR description about the operational model so reviewers don't expect remote-flip behavior on the next build.

dante01yoon (commented):
The refactor in `0f174ad` cleanly addresses both of my prior non-blocking notes — the test no longer pins a hardcoded `toBe(false)`, and `featureFlags.ts` derives from a typed snapshot (schema-validated against the `/features` response). The fetcher's retry/timeout/fallback logic and the unit coverage are solid.

That said, before approving I'd like to confirm one thing about the wiring, because right now most of the new code isn't reachable from the build pipeline.

This is a question, not a request-changes — happy to approve once it's clarified or once the missing wiring lands.

PR #949, opened 2026-06-18
fix(seo): default canonical/sitemap origin to apex comfy.org

christian-byrne:
## Problem / Goal

The canonical SEO origin defaulted to `https://www.comfy.org` in `astro.config.mjs` (`site` → sitemap `<loc>`s, `siteOrigin` → custom sitemap pages) and `src/config/site.ts` (`DEFAULT_ORIGIN` → canonical `<link>` tags + structured data). When `PUBLIC_SITE_ORIGIN` is unset, the build emits `www.comfy.org` URLs. The comfy-router canonicalizes everything to apex `comfy.org` (and now normalizes workflow pages to trailing-slash form, Comfy-Org/comfy-router#31), so any `www` URL in the sitemap or a canonical tag becomes a 301 — the sitemap would advertise redirecting URLs, an SEO anti-pattern. The unified `/sitemap-index.xml` only rewrites the **index** `<loc>`s to apex; the page `<loc>`s inside child sitemaps are proxied raw, so they depend entirely on what this app emits.

## Proposed Solution

Make the apex origin **declarative / IaC** instead of relying on a manually-set Vercel dashboard env var:

1. **`site/vercel.json` → `build.env.PUBLIC_SITE_ORIGIN = "https://comfy.org"`** — committed source of truth, applied during the Astro build. No dashboard step, reproducible per-deploy. (`build.env` is valid per the `openapi.vercel.sh/vercel.json` schema.)
2. **Code default → `https://comfy.org`** in `site.ts` + `astro.config.mjs` — fallback for local dev / CI / any non-Vercel build (these don't get `vercel.json` build env).

Trailing-slash is already correct here (Astro emits `/workflows/{slug}/`); this closes the host-origin gap.

## Acceptance Criteria

- [x] `PUBLIC_SITE_ORIGIN` pinned to apex in `site/vercel.json` (`build.env`).
- [x] `site`, `siteOrigin`, `DEFAULT_ORIGIN` default to `https://comfy.org`.
- [x] No test/CI asserts the old `www` default (`seo-smoke-test.yml` asserts apex).
- [x] `Lint & Format` + `build-and-test` green. (`preview` job fails on an **unrelated** Hub API 403 in `build-search-index` — secret/permission issue, not from this change.)
- [ ] After deploy, `seo-smoke-test` continues to pass (canonical = `https://comfy.org/`).

[...]

PR #950, opened 2026-06-18
Feat/comfyhub redesign

MaanilVerma:
## Summary

End-to-end visual redesign of the ComfyHub site to the new Figma direction — a real marketing navbar, a much stronger search, a single browse toolbar replacing the sidebar/drawer filters, a featured carousel, and reworked cards and detail page. SEO and i18n are preserved across all 11 locales; the whole hub now reads as one consistent unit instead of a basic filtered grid.

Figma Link
- https://www.figma.com/design/11vkE4FAn4plEYpawd57zS/Comfy----Website-Design?node-id=5825-351202&m=dev
- https://www.figma.com/design/11vkE4FAn4plEYpawd57zS/Comfy----Website-Design?node-id=3724-38299&m=dev

## Changes

**What**

- **Marketing navbar**, ported from the current comfy.org nav so it matches upstream exactly: a desktop mega-menu (multi-column dropdowns with a featured image card per section, "NEW" badges, animated CTA pill) and a mobile slide-in sheet with drill-down sub-menus and a back button. Keeps the hub-only "Top creators" shortcut. Built on reka-ui navigation-menu/sheet primitives.
- **Search that stops silently failing.** It used to intermittently return nothing and miss obvious matches. Index and query now share one tokenizer config so they can't drift; a failed first index fetch no longer poisons the cache (it retries and shows a clean empty state); the tokenizer splits hyphen *and* underscore terms so `flux_image-to-video` is found by "flux"/"image"/"video"; common abbreviations expand at query time ("txt2img" → text to image, "i2v" → image to video, "cn" → controlnet); fuzziness scales with term length so short queries stay tight and long typos still match; multi-word queries match all terms first and only fall back to any-term when that's empty. Real usage counts now feed ranking, so popular workflows surface.
- **Browse toolbar** replaces the sidebar and the mobile filter drawer with one component (tabs + sort + facet filters), so filtering is the same on every screen size.
- **Featured carousel** — a usage-ranked hero on the landing page.
- **Workflow card** — 4:3 thumbnail with the title and provider logo overlaid on a scrim, a creator line, a circular CTA, and a tag row that scrolls with chevrons that only appear when the tags overflow. Author-name clipping is fixed, cards are a bit wider, and "Load more" matches Figma.
- **Detail page** — a breadcrumb trail in place of the back button, an expandable description whose full text is always in the SSR DOM (so crawlers see it while it's collapsed), an accordion FAQ, and a "View all workflows" related grid.
- **SEO** — FAQ and breadcrumb JSON-LD are escaped before injection (no script-breakout) and applied consistently across the workflow/category/tag/model pages; breadcrumb labels are localized and mirror the visible trail; collection pages carry an ItemList count.
- **Search a11y** — wrapped in `<form role="search">`, Enter activates the highlighted (or first) result with no page reload, and a live region announces the result count.

**Breaking**

None. Protected infra is restyled around, never removed — SEOHead/HreflangTags, canonical `{name}-{shareId}` links, `getCloudCtaUrl()`/UTM params, and `<Analytics />` are all intact, and the new JSON-LD matches what's rendered.

## Review Focus

- **Search is the riskiest area** — see the AND→OR fallback and the index-fetch recovery; both change behavior users will notice. The shared tokenizer config is the one file to read for why index and query stay in sync.
- **Navbar is a faithful port**, not a fresh design — it intentionally mirrors the upstream comfy.org nav (mega-menu structure, mobile sheet, featured cards) so it stays in sync; the only deliberate addition is the hub's "Top creators" section in the mobile sheet.
- **Filter migration** — the browse toolbar fully replaces the removed sidebar and mobile filter drawer; worth confirming facet-count and active-filter parity.
- Island architecture held throughout — cross-island state via shared composables, no custom-event bridges, islands get plain data + locale and don't import i18n.

## Testing

Search is the part most likely to regress silently, so it has the most coverage; the rest is verified by the type/lint gate plus manual passes against the Figma.

- `tests/unit/search-config.test.ts` — tokenizer sub-parts, synonym expansion, length-aware fuzzy thresholds, and the AND→OR fallback (including the title-over-description ranking).
- `tests/unit/use-facets.test.ts` — facet derivation and dedup from the template set.
- `tests/unit/featured.test.ts` / `tests/unit/provider-logos.test.ts` / `tests/unit/routes.test.ts` — usage-ranked selection, word-boundary logo matching, and canonical slug composition.
- Gate: `pnpm run check` (0 errors), `pnpm test` (168 passing), `pnpm run lint`, `pnpm run build` all green.
- Manual (`pnpm run dev`): search `txt2img` / `i2v` / `cn` and `flux`; block the first `search-index.json` request once and search again to confirm it recovers; open a desktop dropdown and the mobile sheet drill-down; scroll a card's tag row via the chevrons; spot-check a `/zh/…` and the RTL `/ar/…` page for i18n + dir.

Screen Recording

https://github.com/user-attachments/assets/e9e21f1a-379f-4c73-8344-d6de81f88c08


site/src/pages/[locale]/workflows/[slug].astro
```diff
@@ -205,6 +221,10 @@ const structuredData = !isDetailPage
       inLanguage: locale,
     }
   : null;
+
+// "View all workflows" grid for the detail page — full index minus current.
+const detailRelated: SerializedTemplate[] =
+  isDetailPage && detailData ? await listRelatedWorkflows(detailData.name, profiles) : [];
```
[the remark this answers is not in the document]
Comment (MaanilVerma) [reply site/src/pages/[locale]/workflows/[slug].astro:194]: listRelatedWorkflows() already wraps its whole body in try/catch → [], so the unguarded await can't 500 the page.

site/src/components/hub/nav/HeaderMainMobile.vue
```diff
@@ -0,0 +1,166 @@
...
+        :close-label="t('nav.close', locale)"
+      >
+        <SheetHeader class="sr-only">
+          <SheetTitle>{{ t('nav.menu', locale) }}</SheetTitle>
+          <SheetDescription>{{ t('nav.menu', locale) }}</SheetDescription>
+        </SheetHeader>
+
+        <a
+          :href="navRoutes.home"
+          class="focus-visible:border-primary-comfy-yellow focus-visible:ring-primary-comfy-yellow/50 inline-flex w-auto shrink-0 focus-visible:ring-3"
```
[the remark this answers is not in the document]
MaanilVerma: navRoutes.home is https://comfy.org/ — an absolute URL to the separate comfy.org marketing site, intentionally not localized (documented in nav-routes.ts).

site/src/lib/search-config.ts
```diff
@@ -0,0 +1,120 @@
...
+
+/**
+ * Expand known abbreviations in a raw query before searching. Whole-token
+ * matches only (so "i2v" expands but "vi2vid" doesn't). Returns the original
+ * query when nothing matches.
+ */
+export function expandQuery(query: string): string {
+  return query
+    .split(/\s+/)
+    .map((word) => QUERY_SYNONYMS[word.toLowerCase()] ?? word)
```
dante01yoon: issue: Abbreviation expansion **replaces** the literal token instead of augmenting it, regressing searches for workflows whose title contains the abbreviation verbatim.

`expandQuery('t2v')` returns `'text to video'` and drops the `t2v` token; combined with `combineWith: 'AND'` (line 117) the query now requires `text AND to AND video`. Real templates `Wan2.1 Alpha T2V` and `Squish Anything with LTX-2 I2V and LoRA` (both in `templates/index.json`) carry the literal abbreviation in the title but none of the expanded words, so they're now **unfindable** by `t2v`/`i2v` — they matched before this PR (no expansion). The OR fallback at line 119 doesn't rescue them either, since `text`/`to`/`video` don't appear.

Consider keeping the literal as an alternative — e.g. add the original (unexpanded) query as a final fallback in `searchWithFallback`, or expand so the abbreviation OR the phrase can match — rather than substituting. Note no unit test covers a literal-abbreviation document, so the suite stays green through this regression.

site/src/lib/github.ts
```diff
@@ -0,0 +1,74 @@
...
+  return count.toString();
+}
+
+function readGitHubStarsOverride(): number | undefined {
+  const rawCount = import.meta.env.WEBSITE_GITHUB_STARS_OVERRIDE;
+  if (rawCount === undefined || rawCount === '') return undefined;
+
+  const count = Number(rawCount);
+  if (!Number.isSafeInteger(count) || count < 0) {
+    throw new Error('WEBSITE_GITHUB_STARS_OVERRIDE must be a non-negative integer');
```
dante01yoon: issue: This `throw` contradicts the file header's contract ("Returns null on any error so the badge can be hidden gracefully") and is reached from `readGitHubStarsOverride()` at line 14 — *before* the try/catch in `doFetch`.

The only caller, `HubNavbar.astro:13` (`const rawStars = await fetchGitHubStars(...)`), is un-guarded, and `HubNavbar` renders on every `workflows/*` route. So a single malformed env value (`WEBSITE_GITHUB_STARS_OVERRIDE=abc` / `1.5` / `-1`) throws during static build and fails the entire site build, instead of degrading to a hidden badge.

Suggest returning `undefined` (fall through to the live fetch) or `null` on an invalid override rather than throwing, to keep the documented graceful-degradation behavior.

site/src/pages/workflows/model/[name].astro
```diff
@@ -62,20 +66,7 @@ const structuredData = {
...
-    {
-      '@type': 'ListItem',
-      position: 2,
-      name: t('labels.models', locale),
-      item: absoluteUrl(localizeUrl('/workflows/model/', locale)),
-    },
-    { '@type': 'ListItem', position: 3, name: modelName },
-  ],
-};
+const breadcrumbStructuredData = buildWorkflowBreadcrumb(locale, { name: modelName });
```
Comment (dante01yoon) [inline site/src/pages/workflows/model/[name].astro]: issue: The model-page breadcrumb regressed its middle crumb. It used to be `name: t('labels.models')` ("Models") → `item: absoluteUrl(localizeUrl('/workflows/model/', locale))`. `buildWorkflowBreadcrumb(locale, { name: modelName })` falls back to the **default** section `t('breadcrumb.workflows')` ("Workflows") → `/workflows/` (see `structured-data.ts:46-48`).

So the model listing's structured-data trail changed from `Home → Models → {model}` to `Home → Workflows → {model}`, losing the section-specific label and URL. Same in the localized variant `[locale]/workflows/model/[name].astro:64`. The PR description says breadcrumbs are "applied consistently across … model pages," which this contradicts.

If intentional, please confirm. Otherwise pass a `section` override, e.g. `buildWorkflowBreadcrumb(locale, { name: modelName }, { name: t('labels.models', locale), item: absoluteUrl(localizeUrl('/workflows/model/', locale)) })`.

site/src/composables/useCurrentPath.ts
```diff
@@ -0,0 +1,31 @@
...
+    document.removeEventListener('astro:page-load', update);
+    window.removeEventListener('popstate', update);
+  });
+
+  return currentPath;
+}
+
+/** True when a nav href points at the current page (ignores hash/query, skips externals). */
+export function isHrefActive(href: string, currentPath: string): boolean {
+  if (!href || !currentPath || href.startsWith('http')) return false;
```
dante01yoon: question: `isHrefActive` short-circuits to `false` for any href starting with `http`, but every entry in `config/nav-routes.ts` is an absolute `https://comfy.org/...` URL — so `isHrefActive`/`isNavItemActive` (`HeaderMainDesktop.vue:19-22`) can never return true and no nav item ever receives the active highlight. The `astro:page-load` listener (lines 12/17) also never fires, since this site has no `ClientRouter`/ViewTransitions (full page loads), so `currentPath` only updates once on mount.

Is active-state highlighting expected to work here? As written it's inert client-side reactivity. If the dropdowns are intentionally all off-site, consider dropping the active-state machinery to avoid shipping dead code; if not, the comparison needs to normalize to pathnames.

site/src/styles/design-tokens.css
```diff
@@ -38,17 +38,23 @@
   font-display: swap;
 }
 
+/* Comfy brand palette — declared in @theme so they emit both Tailwind color
+   utilities (bg-primary-comfy-*, …) and :root vars for the derived tokens below. */
+@theme {
+  --color-primary-comfy-yellow: #f2ff59;
```
dante01yoon, line 44: suggestion (primary — per the Workflow-Hub-Consolidation RFC and the #hub Slack thread): the team agreed the Hub will be migrated into the main comfy.org website and design components will be shared. To enable a **soft landing** with minimal churn at migration time, it would be worth aligning this PR's design layer with the **website's** existing design tokens, rather than (re)defining a parallel Comfy palette here — and likewise the matching additions in `global.css` (`--font-display`/`--font-body`/`--color-nav-dropdown`, the `ppformula-optical-center*` utilities).

Could you check whether the website's token names/values can be reused as-is, so migration is mostly a move rather than a re-theme? Concretely, a few new components still hardcode off-scale values (`text-[10px]`/`text-[12px]` in `FeaturedCarousel.vue`, `BrowseToolbar.vue`, `GitHubStarBadge.vue`) that should map to shared type tokens. Happy to pair on which token set is canonical.

site/src/components/hub/BrowseToolbar.vue
```diff
@@ -0,0 +1,258 @@
...
+  <div class="flex items-center justify-between gap-2">
+    <TabsRoot
+      :model-value="store.activeTab.value"
+      class="flex min-w-0 overflow-x-auto scrollbar-hide"
+      @update:model-value="store.setTab($event as HubTab)"
+    >
+      <TabsList
+        class="inline-flex items-center gap-1 rounded-xl border border-white/15 bg-white/8 p-1"
+      >
+        <TabsTrigger
```
dante01yoon, line 101: suggestion (a11y): On mobile (`<sm`) an inactive tab renders icon-only — the icon at line 107 is `aria-hidden="true"` and the label `<span>` is hidden until active (line 111), so inactive `TabsTrigger`s have an empty accessible name. Screen readers / voice-control announce them as unnamed buttons. Add `:aria-label="labels[tab.labelKey]"` to the `TabsTrigger` so the name is always present regardless of viewport.

dante01yoon (commented):
## Strict review — ComfyHub redesign (read-only, not approving per request)

Reviewed at head `37f0634`. I focused on the areas the description flags as risky (search, navbar, filter migration, structured-data escaping) and on the **design-system alignment** raised in the Hub-consolidation RFC thread. Verified against the actual current files (note: the earlier CodeRabbit pass references `MobileMenu.vue`/`SiteNav.vue`/`NavDesktopLink.vue`, which no longer exist — they were refactored into `HeaderMain*.vue`, so several of those comments are stale).

### Primary ask (strategic, from the RFC / Slack thread)
The team agreed the Hub will be migrated into the main comfy.org website with shared design components. To get a **soft landing** (minimal churn at migration time), this PR should reuse the **website's** design tokens rather than defining a parallel Comfy palette/typography here. Left as an inline `suggestion` on `design-tokens.css`. This is the highest-leverage change to make before merge — happy to pair on which token set is canonical.

### Blocking-class issues (inline)
- `issue` Search abbreviation expansion **replaces** the literal token, regressing exact-abbreviation title matches (`Wan2.1 Alpha T2V`, `LTX-2 I2V` become unfindable by `t2v`/`i2v`). — `search-config.ts:64`
- `issue` GitHub stars override `throw` violates the file's own "returns null on any error" contract and can crash the whole site build on a bad env value, because the only caller is un-guarded. — `github.ts:70`
- `issue` Model-page breadcrumb regressed (`Models → /workflows/model/` became `Workflows → /workflows/`), contradicting the "consistent across model pages" claim. — `workflows/model/[name].astro:69` (+ localized variant)

### Questions / suggestions (inline)
- `question` Active-nav highlighting is inert: every `nav-routes` href is an absolute `http(s)` URL, so `isHrefActive` always returns false. — `useCurrentPath.ts:26`
- `suggestion` a11y: mobile icon-only inactive tabs have no accessible name. — `BrowseToolbar.vue:101`

### What checked out well (verified, no action)
- Search **does not** drift index/query config (shared `tokenize`/`SEARCH_FIELDS`/`STORE_FIELDS`), and the index-fetch cache **does** recover from a failed first fetch (`indexPromise` is cleared on error) — both core claims hold.
- `serializeJsonLdForScript` is a correct escaper (`<`/`>`/`&`/U+2028/U+2029 → `\uXXXX`) and every **changed** inline JSON-LD site (FAQ, breadcrumb, softwareApp) routes through it; `listRelatedWorkflows` is wrapped in try/catch → `[]`.
- `useFacets` dedups values per-template (`new Set(field(t))`) — fixes latent count inflation vs. the old drawer.
- `provider-logos.ts` uses word-boundary regex matching — fixes the old `includes()` substring bug (e.g. `Swan` no longer matches `wan`).
- `ExpandableText` keeps full text in SSR DOM while collapsed (crawler-visible), disconnects its `ResizeObserver` on unmount, and honors `prefers-reduced-motion`; `FeaturedCarousel` destroys Embla/autoplay on unmount, handles RTL + 0 slides; scroll-lock ref-counts and restores correctly on unmount.

### Non-blocking notes (not inline)
- **CI:** `Build Site` and `build-and-test` (168 tests) pass. `preview` fails only on a missing `VERCEL_TOKEN` (fork PR — secrets unavailable; the build itself completed). `SEO Audit` fails on a missing `dist/` in its own job. `Lint & Format` is red but dominated by **pre-existing** parse errors in files this PR doesn't touch (`BaseLayout.astro` `arguments`/`is:inline`, and the `[username].astro` line-15 HTML comment that is unchanged from `main`); 0 new ESLint errors are introduced. Worth confirming none of these gate the merge.
- **Pre-existing / out of scope:** the dual-mode badge union bug at `HubBrowse.vue:63-67` (selecting both `app`+`nodeGraph` shows only apps) that CodeRabbit flagged is **byte-for-byte unchanged from `main`** — not introduced here, but a good follow-up.
- **Escaping scope gap:** the co-located TechArticle/SoftwareApplication graph (same `data.title`/`data.description`) still serializes raw via the unchanged `SEOHead.astro` `JSON.stringify`, so the "title/description-derived JSON-LD is escaped" invariant isn't fully true post-PR. Pre-existing path, but consider migrating SEOHead to the new escaper too.
- **nit** Dead code: `mobileDrawerOpen`/`toggleMobileDrawer`/`closeMobileDrawer` in `useHubStore.ts` have no consumers after `MobileFilterDrawer.vue` deletion. `formatStarCount` renders `1000K` in the 999.5k–1M window.

Not approving — strict review only, as requested.

site/src/lib/search-config.ts
MaanilVerma: Search — what we offer now, end to end @dante01yoon 

1. **Exact match wins**
Type t2v, wan 2.7, or flux → the workflow whose name/title literally contains it ranks on top.

2. **Shorthand resolves**
t2i, txt2img, i2v, v2v, s2v, cn, and spelling variants (txt2image, img2video…) → find the right workflows; exact matches first, related ones below.

3. **Versions match any way you type them**
wan 2.7 = wan2.7 (same results); 2.5 never wrongly matches 3.5.

4. **Multi-word stays precise**
qwen image edit → the few Qwen edit workflows, not the whole catalog. Falls back to broader matches only if nothing matches all words.

5. **Typos forgiven**
conrtol → Control, upscalr → Upscale (≈2 edits on real words; numbers stay exact).

https://github.com/user-attachments/assets/7381a9e0-e181-4d7a-8606-365dc80b9573

site/src/styles/design-tokens.css
MaanilVerma: **What I did now:**
- The core brand colors were already a perfect match, so no re-theming needed.
- Lined up everything else with the website: the fonts, the text-centering helper, the dropdown color, and the tiny font sizes (swapped hardcoded values for the shared ones).

**What I'll do as a separate follow-up:**
- The Hub still relies on a layer of "naming shortcuts" (like page, brand, content) that the main site doesn't have yet. So if we just copy the components over today, some styling would come through blank. _( Need to verify if they are still used or not )_

@dante01yoon

site/src/components/hub/FeaturedCarousel.vue
```diff
@@ -0,0 +1,298 @@
...
+// and isn't reconfigured if the OS preference changes mid-session (acceptable —
+// the carousel stays fully usable via the manual prev/next controls).
+const reducedMotion = usePreferredReducedMotion();
+const autoplayEnabled = reducedMotion.value !== 'reduce';
+
+const [emblaRef, emblaApi] = emblaCarouselVue(
+  {
+    loop: true,
+    align: 'center',
+    direction: props.locale === 'ar' ? 'rtl' : 'ltr',
```
dante01yoon: suggestion (non-blocking): RTL is hardcoded to `locale === 'ar'`. That's correct today since `ar` is the only `dir: 'rtl'` locale, but the site already has an `isRTL(locale)` helper, so this silently won't pick up a future RTL locale (he/fa/ur). Since islands intentionally don't import i18n, the cleanest fit is to resolve an `isRtl` boolean in the Astro page (like the labels) and pass it as a prop, mirroring how `featuredLabel`/`prevLabel` are already passed in.

site/src/components/hub/FeaturedCarousel.vue
```diff
@@ -0,0 +1,298 @@
...
+          :key="slide.key"
+          class="relative h-full min-w-0 flex-[0_0_100%]"
+        >
+          <video
+            v-if="slide.videoUrl && !videoFailed[slide.key]"
+            :src="slide.videoUrl"
+            :poster="slide.posterUrl || undefined"
+            class="h-full w-full object-cover"
+            :preload="index === 0 ? 'auto' : 'metadata'"
+            autoplay
```
dante01yoon: suggestion (non-blocking): the `reducedMotion`/`autoplayEnabled` gate only disables Embla's slide auto-advance — each slide's `<video autoplay loop>` still plays regardless, so a `prefers-reduced-motion` user still gets continuously looping motion. Consider gating the video `autoplay` on the same check. Separately, every video slide (up to `FEATURED_COUNT`) mounts with `autoplay`, so they all play at once; only autoplaying the in-view slide would save bandwidth/CPU on mobile.

site/src/components/hub/TagScroller.vue
```diff
@@ -0,0 +1,86 @@
...
+
+const scroller = useTemplateRef<HTMLElement>('scroller');
+const canPrev = ref(false);
+const canNext = ref(false);
+
+function syncEdges() {
+  const el = scroller.value;
+  if (!el) return;
+  canPrev.value = el.scrollLeft > 1;
+  canNext.value = el.scrollLeft < el.scrollWidth - el.clientWidth - 1;
```
dante01yoon: nit (non-blocking): the prev/next edge detection assumes LTR. In the RTL (`ar`) build, `scrollLeft` is 0-at-start and goes negative (Chrome/FF), so `scrollLeft > 1` (canPrev) and `scrollLeft < scrollWidth - clientWidth - 1` (canNext) can be inverted, making the chevrons enable/disable backwards relative to their icons. Low priority since it's the one RTL locale, but worth normalizing if you keep the chevron affordance.

dante01yoon (approved):
Re-reviewed at head `13fba4f` (read-only). Strong, well-documented redesign. I re-verified the substantive CodeRabbit findings against the current head and they are all resolved, and I confirmed the red CI checks are fork-PR infrastructure failures rather than code problems. Findings below are non-blocking nits/suggestions, so approving.

**CI red checks are infra, not code** — worth noting so they aren't mistaken for blockers:
- `Lint & Format`: ESLint and Prettier both passed; the job only failed at the *Comment on PR* step with `403 Resource not accessible by integration` (read-only token on a fork PR).
- `SEO Audit`: Sitemap Validation and SEO Audit both passed; same 403 on the comment step.
- `preview`: the Vercel deploy step needs `VERCEL_TOKEN`, which isn't exposed to fork PRs.
- The real quality gates — `Build Site`, `build-and-test`, CodeRabbit, Socket Security, CLA — are green.

**CodeRabbit findings re-verified as resolved on this head:**
- JSON-LD XSS: every workflow route page now serializes via `serializeJsonLdForScript` (escapes `<`/`>`/`&`/U+2028-9); the only remaining raw `JSON.stringify` is in unchanged `SEOHead.astro`, out of scope.
- `github.ts` inflight cache now clears via `.finally(() => inflight.delete(key))` — no permanent caching of failures.
- `provider-logos.ts` uses word-boundary regex, so `Swan AI` no longer matches `Wan`.
- `useFacets.ts` dedups per-template values with `new Set(field(t))` before counting.
- `listRelatedWorkflows` already wraps its fetch in try/catch returning `[]`, so the detail page can't crash on a Hub API failure.
- The nav was rewritten (`MobileMenu.vue`/`NavDesktopLink.vue`/`SiteNav.vue` -> `HeaderMain*`/`NavColumn`), so CodeRabbit's nav comments are stale/moot.
- The dual-mode (`app`+`nodeGraph`) first-match-wins filter CodeRabbit flagged is *unchanged pre-existing code* and currently unreachable (the discovery panel that offers a second mode chip hides once any badge is active), so it isn't introduced here.

**Verified:** all `t('...')` keys referenced by new code resolve in `ui.ts` (a missing key would render the literal key); FAQ JSON-LD and the visible `FaqAccordion` share the same `faqItems` source; the breadcrumb JSON-LD mirrors the visible `Breadcrumbs.astro` trail in label and order.

**Not run:** static review + CI log analysis only — I did not run the app or re-run the build/tests locally, and did not verify visual/Figma fidelity.

site/src/pages/workflows/model/[name].astro
```diff
@@ -62,20 +67,14 @@ const structuredData = {
...
-    },
-    { '@type': 'ListItem', position: 3, name: modelName },
-  ],
-};
+const breadcrumbStructuredData = buildWorkflowBreadcrumb(
+  locale,
+  { name: modelName },
+  {
+    name: t('labels.models', locale),
+    item: absoluteUrl(localizeUrl('/workflows/model/', locale)),
```
Comment (dante01yoon) [inline site/src/pages/workflows/model/[name].astro:75]: suggestion (non-blocking, follow-up to my earlier breadcrumb finding): restoring the **Models** crumb is exactly right and resolves the regression. One thing I missed in the original review — this crumb's `item` URL `/workflows/model/` doesn't resolve to a page: there's only `[name].astro` here (no `index.astro`) and no redirect, so the bare path is a 404. Note this is **not introduced by this PR** — `main` pointed the same crumb at `/workflows/model/` too, so it's a pre-existing latent issue, and there's no user-facing broken link (only the JSON-LD `BreadcrumbList`, no visible trail on model pages).

Still, Google expects breadcrumb `item` URLs to resolve, so while we're touching this line it'd be cheap to either drop the `item` (schema.org allows a name-only crumb — same as the leaf crumbs on the category/tag pages) or add a `/workflows/model/` index. Same line in the localized variant `[locale]/workflows/model/[name].astro:75`.

dante01yoon (commented):
## Re-review at `13fba4f` — all 6 prior findings verified resolved ✅

Thanks for the fast turnaround, @MaanilVerma. I re-reviewed the follow-up commits since my last pass (`37f0634..13fba4f`) and adversarially verified each fix (rebuilt the MiniSearch index with the current shared config to test queries empirically, ran the unit suites, and grepped for dangling references/tokens). Every prior finding is genuinely resolved:

| Prior finding | Fix | Verified |
|---|---|---|
| `issue` search abbrev regression | `searchWorkflows` now searches the **literal** query first, then appends the expansion (deduped); `expandQuery` returns `null` when nothing expands | ✅ `t2v`→ returns `Wan2.1 Alpha T2V` literal **and** the expansion, literal ranked first; new test asserts ordering; 58/58 pass |
| `issue` GitHub override throw | `console.warn` + `return undefined` (falls through to live fetch) | ✅ every invalid input returns instead of throwing; new `github.test.ts` 4/4 pass; contract upheld |
| `issue` Models breadcrumb | section override restores `Models → /workflows/model/` in both default + localized pages | ✅ localized for all 11 locales, no raw-key leak (see one non-blocking note below) |
| `question` inert active-nav | `useCurrentPath.ts` deleted, all consumers cleaned | ✅ zero dangling refs; `NavigationMenuTrigger.active` is optional so the contract holds |
| `suggestion` design tokens | `--text-2xs/3xs`, `--color-site-dropdown`, `ppformula-text-center` tokens added; `text-[10px]`/`[12px]` removed | ✅ no dangling token refs; `0.625rem`=10px exact (no visual regression); good step toward the migration soft-landing |
| `suggestion` tab a11y | `:aria-label="labels[tab.labelKey]"` on `TabsTrigger` | ✅ localized name now always present |

Not approving (review-only, as before). Two small **non-blocking** follow-ups inline — neither blocks merge.

<details><summary>Verification notes</summary>

- Build/lint sanity sweep of the follow-up commits found no dangling imports, unused-var lint failures, or undefined token classes.
- The named search templates (`Wan2.1 Alpha T2V`, `LTX-2 I2V`) are only *findable* if present in the hub-API search index (it indexes hub workflows, not `templates/index.json`) — the fix is correct for any indexed doc carrying a literal abbreviation regardless.
</details>

site/src/lib/search-config.ts
```diff
@@ -0,0 +1,167 @@
...
+}
+
+/** Per-term fuzziness: 0.3 (≈2 edits, forgives "conrtol" → "control"), but exact
+ *  for ≤3-char terms (noise) and any term with a digit (so "2.5" ≠ "3.5"). */
+export function termFuzziness(term: string): number | false {
+  return term.length <= 3 || /\d/.test(term) ? false : 0.3;
+}
+
+/** Prefix matching + per-term fuzziness; AND by default so multi-word stays precise. */
+export function searchOptions(combineWith: 'AND' | 'OR' = 'AND') {
```
dante01yoon, line 67: question (non-blocking): the rewritten `searchOptions` drops the per-field `boost` map the old version had (`title:3, models:2, tags:2, creatorName:1.5, mediaType:1, description:0.5`), so ranking now leans entirely on MiniSearch's BM25 field-length normalization. In the fixtures I tested, ordering didn't flip (BM25 still favors short fields like title), so this isn't a correctness bug — but title/model matches are no longer *explicitly* weighted above description matches. Was dropping `boost` intentional, or collateral from the rewrite? If unintentional, re-adding `boost` to the returned options object would restore the prior relevance tuning.

Unrelated nit on the same file (line 46): the version regex `^([a-z][a-z.]*?)(\d+(?:\.\d+)*)$` lets the name group keep a trailing dot, so `flux.1` tokenizes to `flux.` (with the dot) rather than `flux`. Harmless today thanks to prefix matching, but the emitted sub-token is dirty.

MaanilVerma: Good catch! the boost map shouldn't have been dropped. The intent  was that a title match ranks above an incidental mention in description/tags; I was relying on BM25 field-length normalization to deliver that (titles are short, so title hits already float up), which holds in the fixtures but isn't guaranteed when a long title competes with a short description. That's fragile and implicit. Restoring an explicit title boost so the ranking intent is in the code, not a side effect of field lengths.

site/src/pages/workflows/model/[name].astro
Comment (MaanilVerma) [reply site/src/pages/workflows/model/[name].astro:75]: @dante01yoon This is actually addressed in a follow-up PR that comes right after this one is merged (and I don't think we autodeploy). In that follow-up PR, we've introduced /models as its own page along with several SEO-related updates, so this is already taken care of there.

dante01yoon:
🚀 **Preview deployed** for this PR (head `2fcc956`):

https://workflow-templates-h3x07jaod-comfyui.vercel.app

Deployed manually by a maintainer as a one-off preview. Note: built without the preview API env vars, so data-driven sections may render empty — the comfyhub redesign UI itself is fully viewable.

Going forward, fork PRs will get automatic previews once a maintainer adds the `preview` label (see #965).

dante01yoon:
Verify actual CI passed - failures happen due to token issue

MaanilVerma:
Knit Picks Updated  (It start from this [Commit ID ](https://github.com/Comfy-Org/workflow_templates/pull/950/commits/c6798732e1e6f4d062981c116e48023d0a39fb48))

What we fixed:
1. Author name now lines up neatly with the avatar on cards and the featured banner.
2. Card tags now show only full, readable tags that fit, then a "+N" chip — no more clipped "Vi…" stubs.
   Hovering "+N" lists the remaining tags. Tags read in normal Title Case; the old scroll arrows are gone.
3. Cards with no tags show a single fallback chip (Image / Video / Audio / 3D) so every card looks balanced.
4. The featured banner is now fully clickable — clicking the title or empty area opens the workflow
   (the creator and tag links still go to their own pages).
5. Navbar items no longer flash a white box on keyboard focus; they show a clean yellow focus ring instead.
6. Card hover: the corner button starts as a chevron and smoothly reveals "TRY NOW" on hover.
7. Search bar moved out of the header into the page, above the filters — it now sticks to the top once you
   scroll past it and drops back into place on scroll up. (Other pages keep the search in the header as before.)
8. The filter toolbar no longer sticks (it scrolls away normally); only the search bar pins.
9. Multiple model logos now show as a small overlapping stack (max 3) on the top-right of a card;
   hovering one brings it to the front.
10. Removed the big circle handle from the before/after comparison slider on cards.

Under the hood:
- Search dropdown badge labels (creators / categories / models) are now vertically centered.
- Added a reusable tooltip + a "fit + overflow" tag component, with unit and e2e test coverage.
- Removed duplicated types and dead scroll-shadow code; tightened comments.

dante01yoon:
🚀 **Preview deployed:** https://workflow-templates-9xscoj11g-comfyui.vercel.app

Manual prebuilt Vercel deploy of this PR's head (`0d6ec85`), built with `SKIP_AI_GENERATION=true` (content from synced collection). Home and `/workflows` verified 200.

site/src/components/ui/button-pill/index.ts
```diff
@@ -0,0 +1,148 @@
...
+);
+
+/**
+ * Reveal label: animates a grid column 0fr→1fr on hover (pill or ancestor
+ * `group/pill-trigger`). Grid, not max-width, so the label's box height — and
+ * thus its baseline/glyphs — is never clipped. Only rendered for reveal pills.
+ */
+export const buttonPillLabelVariants = cva('grid transition-[grid-template-columns] duration-500', {
+  variants: {
+    open: {
```
dante01yoon: nit-pick (non-blocking): `buttonPillLabelVariants` is only ever called as `buttonPillLabelVariants()` in ButtonPill.vue, so `open` always resolves to its `false` default and the `open: true` branch (`grid-cols-[1fr]`) is unreachable. Either drive it from a prop (e.g. an always-open / `expanded` state) or drop the variant and inline the hover-reveal class. Not blocking.

site/tests/e2e.spec.ts
```diff
@@ -216,48 +214,35 @@ test.describe('Search Filter "+X more" Expansion', () => {
...
 
-    // Skip if no container overflows at this viewport width
-    test.skip(!overflowingContainer, 'No tag container overflows at 375px width');
-
-    // Get the last tag link — it should be partially or fully clipped
-    const lastTag = overflowingContainer!.locator('a').last();
-    await expect(lastTag).toBeAttached();
+    const firstTag = rows.locator('a.tag-link').first();
+    test.skip((await firstTag.count()) === 0, 'No card has tags at this viewport');
+    await Promise.all([page.waitForNavigation(), firstTag.click()]);
```
dante01yoon: nit-pick (non-blocking): `page.waitForNavigation()` is deprecated (astro check flags it as ts(6387)). Since the assertion right below already matches on the URL, `await Promise.all([page.waitForURL(/\/workflows\/tag\/.+\//), firstTag.click()])` would be the non-deprecated equivalent. Carried over from the old test, so not a regression — just a cleanup.

dante01yoon (approved):
Re-review of the Knit-Picks round since my prior approval (`13fba4f1` → `0d6ec856`). The incremental changes are clean polish — TagScroller→TagRow fit+overflow, stacked model logos, ButtonPill `reveal` try-now, sticky in-body search, RTL/reduced-motion fixes, restored field boost, focus-ring fix.

Verified locally on the PR head:
- `astro check`: 0 errors / 0 warnings.
- vitest: 213 passing (both changed suites — `search-config` +30, `provider-logos` +29 — green).
- No dangling refs after the renames/moves: `TagScroller`, the old `tag-pills` testid, and the local `CreatorEntry` export are all gone, with `CreatorEntry` now sourced from `hub-api` and consumed only by `SearchPopover`.
- FeaturedCarousel "fully clickable" is sound: a real full-slide `<a class="absolute inset-0 z-10">` sits behind the now-`pointer-events-none` overlay, and the creator/tag children re-enable `pointer-events-auto` with `@click.stop`.
- ButtonPill `reveal` self-hover works (`group/button-pill` is on the cva base) and the card path uses `group/pill-trigger`.

Note on red checks: Lint & Format, SEO Audit, and preview fail only at their final "Comment on PR"/deploy step with `403 Resource not accessible by integration` (fork-PR token), not on real checks — `ESLint ✅`, `Prettier ✅`, `Sitemap ✅`, `SEO Audit ✅` all passed. Matches the maintainer note.

Approving. Two non-blocking nits inline.

dante01yoon (approved):
Re-review of the latest round since my prior approval (`0d6ec856` → `727d5c61`). 5 files; both of my earlier nits are resolved and the navbar wordmark fix verified.

**Nit follow-ups (both addressed):**
- `buttonPillLabelVariants` removed and inlined into ButtonPill.vue. The inline class set is byte-for-byte equivalent to the old `buttonPillLabelVariants()` default (`open: false`) output — same six utilities, only reordered — so the reveal animation is unchanged. No leftover references anywhere.
- `page.waitForNavigation()` → `page.waitForURL(/\/workflows\/tag\/.+\//)` using the same regex as the assertion below. `astro check` deprecation hints dropped 5 → 4, confirming the ts(6387) flag is gone.

**Wordmark (SVG rebuild + HeaderMain sizing) — rendered to verify:**
- The SVG renders cleanly as "Comfy HUB" with no doubled/overlapping glyphs (I'd flagged a possible overlap between the original path and the new `scale(0.6115)` group — render confirms it's a single clean lockup).
- Intrinsic 240×48 → at `h-10` (40px) the image is 200px wide; new container `max-w-56` (224px) fits it without clipping, where the old `w-44` (176px) was cutting the HUB badge. Sizing math and the rendered in-context capture both confirm no clip.

**Validation on `727d5c61`:** `astro check` 0 errors / 0 warnings; vitest 213 passing; CI `ESLint ✅` / `Prettier ✅` (Lint & Format / SEO Audit / preview still red only on the fork-token `403 Resource not accessible by integration` comment/deploy step — not real failures).

No new findings. Approving.

MaanilVerma:
Few Polishing Updates

1. Detail page (TemplateDetailPage.astro)
- Sidebar now sticky while video + description scroll past it (pins below the search header)
- Long description moved below the video in the right column (was full-width)
- "Created by" date smaller than the artist name (text-base → text-xs)
- CTA buttons (Try on Comfy Cloud / Download Workflow) restyled to the design-system variants (brand-solid / pill-outline, rounded-2xl)
- Removed noise comments; analytics hooks kept intact

2. Hub controls (FeaturedCarousel.vue, button-pill/index.ts)
- Carousel arrows: glass at rest, yellow on hover (both match — yellow is rollover only)
- Card icon button: single background, glass-at-rest badge → solid yellow "Try now" pill on hover

3. Small polish
- "Node graphs" → "Node Graphs" tab label
- Bottom padding under the search bar to match the top spacing (non-home pages)
- "Load more" label optically centered (ppformula nudge)

dante01yoon:
Preview deploy for this PR (Astro site in `site/`):

**https://workflow-templates-bnun00qin-comfyui.vercel.app**

- Built from PR head `be4ed1f` and deployed prebuilt to the `workflow-templates` Vercel project as a preview.
- Templates synced en-only; hub data loads live from `cloud.comfy.org` with the content-collection fallback. `/` redirects to `/workflows/`.
- Ad-hoc manual deploy (not the label-gated CI preview) — ping me to redeploy after new commits.

site/src/components/template-detail/TemplateDetailPage.astro
```diff
@@ -88,134 +88,47 @@ function formatRelativeDate(dateStr: string): string {
...
             {
               showCloudCta && (
                 <a
                   href={getCloudCtaUrl(cloudTemplateId, 'hero')}
                   target="_blank"
                   rel="noopener noreferrer"
-                  class="run-cloud-btn flex flex-col items-center justify-center py-2 bg-brand rounded-full text-sm text-page hover:opacity-90 transition-opacity"
+                  class={cn(
+                    buttonVariants({ variant: 'brand-solid' }),
+                    'run-cloud-btn flex flex-col py-2 transition-opacity'
```
dante01yoon: suggestion (non-blocking): `buttonVariants({ variant: 'brand-solid' })` pulls in the default size (`h-9` -> fixed 36px), but this CTA is `flex flex-col` and renders a second line (`cta.tryCloudSubline`) when `featureFlags.cloudFreeTier` is on. Two stacked lines (~36px of text) + `py-2` (16px) don't fit in 36px, so the subline would overflow below the brand box and into the `gap-3` above the download button. The previous classes had no fixed height and grew to fit.

It's masked today because `cloudFreeTier` is `false` in `feature-flags.snapshot.json`, so this is latent -- but it would break the hero CTA when the flag flips. Adding `min-h-10` (or `h-auto`) to the local override on this line keeps the single-line look and survives the two-line case.

dante01yoon (approved):
Re-review of the delta since my last approval (`59b989b3` → `be4ed1ff`). The new work is the six "Few Polishing Updates" commits plus two clean merges from `main` (CLA #974, SEO origin #949 — verified the merge introduces no evil-merge edits to feature files).

The changes are all design polish and check out:
- `TemplateDetailPage` flex → 2-col grid (sticky sidebar `row-span-2`, media + description stacked in col 2). Structure is balanced; mobile order (media → sidebar → description) is preserved.
- CTAs repointed to `buttonVariants` — `brand-solid` and `pill-outline` both exist in `button/index.ts`.
- Carousel arrows unified to glass-at-rest / yellow-on-hover; `button-pill` reveal variant refactor; `ppformula-text-center-sm` is defined in `global.css`; `Node graphs` → `Node Graphs` (en only).

Real CI is green — `build-and-test` and `Build Site` pass, and the red `Lint & Format` job actually ran ESLint OK / Prettier OK and only failed on the 403 PR-comment step (known fork-token issue).

Approving. One non-blocking suggestion inline on the hero CTA height — it's gated behind a flag that's currently off, so no production impact today.

MaanilVerma:
**What changed from previous approve.** 

Docks the inline hub search flush beneath the sticky navbar so the two read as one continuous `bg-page` surface on scroll.

**Test:** scroll the hub index — search pins directly under the navbar with no gap/overlap at mobile and lg.

dante01yoon (approved):

PR #965, opened 2026-06-24
ci: deploy fork-PR previews via label-gated workflow_run

dante01yoon:
## Problem

Preview deploys fail for PRs opened from forks (e.g. #950). On `pull_request` events from a fork, GitHub withholds repository secrets, so the Vercel step runs with an empty token:

```
Error: You defined "--token", but it's missing a value
```

The build itself passes — only the deploy step fails, surfacing a red X to external contributors.

## Approach

Split the preview into build (untrusted, no secrets) and deploy (trusted, secrets) so the Vercel token never touches fork-authored build code:

- **`preview-site.yml`** (`pull_request`) now only builds the Astro site and uploads the prebuilt `.vercel/output` + PR number as an artifact. No secrets, no deploy. Added `labeled` to the trigger types so applying the label re-runs the build and chains a deploy.
- **`deploy-preview.yml`** (`workflow_run`) runs in the base-repo context (secrets available), is gated by the maintainer-only **`preview`** label, downloads the prebuilt artifact, and runs `vercel deploy --prebuilt`. It posts the preview URL as a sticky PR comment.

Because the deploy job only ships an already-built artifact (no `pnpm build` of fork code runs with the token in scope), this avoids the secret-exfiltration risk of `pull_request_target`/`issue_comment` patterns.

## Maintainer usage

Add the **`preview`** label to a fork PR → the build re-runs → a preview is deployed and the URL is commented. Remove + re-add the label to redeploy after reviewing new commits.

## Notes

- `PUBLIC_HUB_API_URL` / `PUBLIC_COMFY_CLOUD_URL` are baked at build time and remain unset for fork builds (pre-existing). If correct preview endpoints are needed for fork PRs, move these non-secret public values to repo Variables. Out of scope here.
- `workflow_run` workflows don't appear as PR status checks, so contributors no longer see a red X for the deploy.
- actionlint passes on both files.

### DIFF
--- a/.github/workflows/deploy-preview.yml
+++ b/.github/workflows/deploy-preview.yml
@@ -1,0 +1,104 @@
+name: Deploy Preview
+
+on:
+  workflow_run:
+    workflows: ['Preview Site']
+    types: [completed]
+
+concurrency:
+  group: deploy-preview-${{ github.event.workflow_run.head_repository.full_name }}-${{ github.event.workflow_run.head_branch }}
+  cancel-in-progress: true
+
+permissions:
+  contents: read
+  actions: read
+  pull-requests: write
+
+jobs:
+  deploy:
+    if: >
+      github.event.workflow_run.event == 'pull_request' &&
+      github.event.workflow_run.conclusion == 'success'
+    runs-on: ubuntu-latest
+    steps:
+      - name: Download prebuilt output
+        uses: actions/download-artifact@v4
+        with:
+          name: vercel-preview-output
+          path: artifact
+          run-id: ${{ github.event.workflow_run.id }}
+          github-token: ${{ secrets.GITHUB_TOKEN }}
+
+      - name: Resolve PR number
+        id: pr
+        run: |
+          number="$(tr -d '[:space:]' < artifact/preview-meta/pr-number)"
+          if [ -z "$number" ]; then
+            echo "No PR number in artifact; nothing to deploy."
+            echo "found=false" >> "$GITHUB_OUTPUT"
+          else
+            echo "found=true" >> "$GITHUB_OUTPUT"
+            echo "number=$number" >> "$GITHUB_OUTPUT"
+          fi
+
+      - name: Check for preview label
+        id: gate
+        if: steps.pr.outputs.found == 'true'
+        env:
+          GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+        run: |
+          labels="$(gh pr view "${{ steps.pr.outputs.number }}" --repo "${{ github.repository }}" --json labels --jq '.labels[].name')"
+          if printf '%s\n' "$labels" | grep -qx 'preview'; then
+            echo "approved=true" >> "$GITHUB_OUTPUT"
+          else
+            echo "approved=false" >> "$GITHUB_OUTPUT"
+            echo "PR #${{ steps.pr.outputs.number }} has no 'preview' label — skipping deploy."
+          fi
+
+      - name: Deploy preview to Vercel
+        id: deploy
+        if: steps.gate.outputs.approved == 'true'
+        env:
+          VERCEL_TOKEN: ${{ secrets.VERCEL_TOKEN }}
+          VERCEL_ORG_ID: ${{ secrets.VERCEL_ORG_ID }}
+          VERCEL_PROJECT_ID: ${{ secrets.VERCEL_PROJECT_ID }}
+        run: |
+          stdout_file="$RUNNER_TEMP/vercel-preview.stdout"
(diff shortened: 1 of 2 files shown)


.github/workflows/preview-site.yml
```diff
@@ -2,6 +2,7 @@ name: Preview Site
 
 on:
   pull_request:
+    types: [opened, synchronize, reopened, labeled]
```
christian-byrne, line 5: issue: `labeled` event type is functionally dead when combined with the `paths: ['site/**']` filter. GitHub evaluates event types and paths as a conjunction, so adding the `preview` label to a fork PR that has no `site/**` changes will not trigger this workflow. The "remove + re-add the label to redeploy" workaround described in the PR description also fails for the same reason -- the only working path for a first deploy is a new push (`synchronize`) after the label is present.

Since the label gate is correctly enforced at deploy-time in `deploy-preview.yml`, the `labeled` type here is not needed. Suggest removing it, or moving label-triggered builds to a separate workflow file without a `paths` filter.

.github/workflows/deploy-preview.yml
```diff
@@ -0,0 +1,104 @@
...
+on:
+  workflow_run:
+    workflows: ['Preview Site']
+    types: [completed]
+
+concurrency:
+  group: deploy-preview-${{ github.event.workflow_run.head_repository.full_name }}-${{ github.event.workflow_run.head_branch }}
+  cancel-in-progress: true
+
+permissions:
```
christian-byrne, line 12: issue: `cancel-in-progress: true` on a deploy workflow can leave a broken Vercel preview if a second push arrives while `vercel deploy` is mid-flight -- the runner is killed after the upload has started but before it completes, and Vercel may surface a stale or errored deployment. `deploy-site.yml` uses `cancel-in-progress: false` for exactly this reason. Deploy workflows should queue rather than cancel.

.github/workflows/deploy-preview.yml
```diff
@@ -0,0 +1,104 @@
...
+            echo "No PR number in artifact; nothing to deploy."
+            echo "found=false" >> "$GITHUB_OUTPUT"
+          else
+            echo "found=true" >> "$GITHUB_OUTPUT"
+            echo "number=$number" >> "$GITHUB_OUTPUT"
+          fi
+
+      - name: Check for preview label
+        id: gate
+        if: steps.pr.outputs.found == 'true'
```
christian-byrne, line 46: issue: label approval can be consumed by a later, unreviewed commit. Sequence: (1) maintainer adds `preview` label to commit A; (2) fork author immediately pushes commit B; (3) `Preview Site` runs on commit B and uploads a new artifact; (4) this workflow fires, checks the label -- still present -- and deploys commit B's content. The label granted for commit A now covers commit B without maintainer re-review.

Two options: (a) add a `pull_request: [synchronize]` workflow that strips the `preview` label on every new push, forcing re-approval per commit; or (b) record the approved SHA when labeling and verify `github.event.workflow_run.head_sha` matches at deploy time.

.github/workflows/deploy-preview.yml
```diff
@@ -0,0 +1,104 @@
...
+          name: vercel-preview-output
+          path: artifact
+          run-id: ${{ github.event.workflow_run.id }}
+          github-token: ${{ secrets.GITHUB_TOKEN }}
+
+      - name: Resolve PR number
+        id: pr
+        run: |
+          number="$(tr -d '[:space:]' < artifact/preview-meta/pr-number)"
+          if [ -z "$number" ]; then
```
christian-byrne, line 36: nitpick: (non-blocking) `tr -d '[:space:]'` strips whitespace but does not validate the result is a non-negative integer before writing to `GITHUB_OUTPUT` and passing to `gh pr view`. In practice `github.event.pull_request.number` is always a positive integer, but a defensive check costs nothing: `if [[ ! "$number" =~ ^[0-9]+$ ]]; then echo "Invalid PR number" >&2; exit 1; fi`

.github/workflows/deploy-preview.yml
```diff
@@ -0,0 +1,104 @@
...
+    types: [completed]
+
+concurrency:
+  group: deploy-preview-${{ github.event.workflow_run.head_repository.full_name }}-${{ github.event.workflow_run.head_branch }}
+  cancel-in-progress: true
+
+permissions:
+  contents: read
+  actions: read
+  pull-requests: write
```
christian-byrne, line 15: nitpick: (non-blocking) `contents: read` is listed but this workflow has no `actions/checkout` step and reads no repo contents. `actions/download-artifact` with `run-id` uses `actions: read` (already present). The existing `workflow_run` workflows in this repo (`spellcheck-comment.yml`, `link-check-comment.yml`) omit `contents: read` and work correctly.

PR #972, opened 2026-06-26
Track workflow search context

benceruleanlu:
Summary:
- Enrich hub:search_performed with result_count, surface, and filters_applied.
- Keep telemetry on the existing 200ms debounced search path.
- Watch derived filter values so in-place badge additions rerun scoped search telemetry.
- Add unit coverage for payload shape and zero-result searches.

Tests:
- pnpm run test -- tests/unit/posthog.test.ts
- pnpm run check
- pnpm exec eslint src/lib/posthog.ts tests/unit/posthog.test.ts --max-warnings 0
- pnpm exec prettier --check src/lib/posthog.ts src/components/hub/SearchPopover.vue tests/unit/posthog.test.ts
- git diff --check

Notes:
- Full pnpm run lint and pnpm run format:check still fail on unrelated pre-existing files in the fresh clone.

Fixes GTM-139


site/src/lib/posthog.ts
```diff
@@ -27,7 +27,14 @@ export function initPostHog(): void {
  * - past tense verbs
  * - e.g. hub:run_button_clicked, hub:template_viewed
  */
-type EventProperties = Record<string, string | number | boolean | undefined>;
+type EventPropertyValue = string | number | boolean | string[] | undefined;
+type EventProperties = Record<string, EventPropertyValue>;
+
+export interface SearchPerformedProperties {
```
pythongosssss: nit: not used outside this class

site/src/components/hub/SearchPopover.vue
```diff
@@ -195,11 +198,18 @@ watchDebounced(
...
     isSearching.value = true;
+    const filtersApplied = [...activeFilters.value];
     try {
-      searchResults.value = await searchIndex(trimmed, {
+      const results = await searchIndex(trimmed, {
         allowedNames: badgeFilteredNames.value ?? undefined,
       });
-      trackSearchPerformed(trimmed);
+      searchResults.value = results;
+      const resultCount = results.workflows.length + matchedCreators.value.length;
```
pythongosssss: Do we want workflows + creators in the same value?

site/src/components/hub/SearchPopover.vue
```diff
@@ -79,6 +79,9 @@ const visibleBadgesDesktop = computed(() => store.filterBadges.value.slice(0, MA
 const overflowCountDesktop = computed(() =>
   Math.max(0, store.filterBadges.value.length - MAX_BADGES_DESKTOP)
 );
+const activeFilters = computed(() =>
```
pythongosssss: is the actual fix here to change addBadge in the store to resassign so the watch triggers on adds

site/src/lib/posthog.ts
benceruleanlu: Removed export

site/src/components/hub/SearchPopover.vue
benceruleanlu: I split it out to workflow_result_count and creator_result_count

site/src/components/hub/SearchPopover.vue
benceruleanlu: Indeed, switched to         filterBadges.value = [...filterBadges.value, badge]

pythongosssss (approved):

PR #974, opened 2026-06-26
Adding back the CLA

alexisrolland:
@comfyui-wiki I am adding back the CLA GitHub action workflow:

- It has been updated to add new `allowlist` to avoid PR getting block by bots which cannot sign.
- We should add it back first without the branch protection rule. This way we can test if it runs smoothly without blocking PRs.


comfyui-wiki (approved):

comfyui-wiki:
Approved

PR #980, opened 2026-06-29
Fix SEO smoke test workflow routes

benceruleanlu:
## Summary
- update the SEO smoke test from the old `/templates` routes to the current `/workflows` canonical fixture
- check `sitemap-workflows-index.xml` instead of the removed templates sitemap
- update served-by assertions for the current Vercel routing headers

## Validation
- ran the smoke-test curl checks locally against `comfy.org` and `workflow-templates.vercel.app`
- `git diff --check`

### DIFF
--- a/.github/workflows/seo-smoke-test.yml
+++ b/.github/workflows/seo-smoke-test.yml
@@ -11,31 +11,35 @@ on:
 jobs:
   seo-checks:
     runs-on: ubuntu-latest
+    env:
+      COMFY_ORIGIN: https://comfy.org
+      VERCEL_ORIGIN: https://workflow-templates.vercel.app
+      WORKFLOW_FIXTURE_PATH: /workflows/flux_schnell-75fb83052003/
     steps:
       - name: Check canonical URLs point to comfy.org
         run: |
           echo "Checking canonical URLs..."
-          BODY=$(curl -s --max-time 15 "https://comfy.org/templates/flux_schnell")
-          if echo "$BODY" | grep -q 'rel="canonical" href="https://comfy.org/'; then
+          BODY=$(curl -fsSL --max-time 15 "$COMFY_ORIGIN$WORKFLOW_FIXTURE_PATH")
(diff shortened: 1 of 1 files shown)


deepme987 (approved):
Approving. Found this while chasing why `SEO Smoke Test` is red on `main` — it has been red on every commit I sampled back to 2026-08-28, and this PR is the fix.

**Verified against the live site today.** I ran both versions' commands. On `main`, **all five steps fail**, not just the first — CI only ever shows step 1 because it exits there:

| step | asserts | actual on `main` |
| --- | --- | --- |
| 1 canonical | `curl -s` (no `-L`) `/templates/flux_schnell` | 301, empty body → grep fails |
| 2 `/templates/` is 200 | `= 200` | **301** |
| 3 sitemap | `sitemap-templates-index.xml` 200 | **404** (`sitemap-workflows-index.xml` is the live one) |
| 4 vercel canonical | `vercel.app/templates/flux_schnell` | **404**, 78-byte body |
| 5 `x-served-by` | `/templates/` contains `vercel`; homepage contains `framer` | no header at all (301); homepage is `vercel-website` now |

So this was never a stale flag — the whole file predates the `/templates/` → `/workflows/` migration and the homepage's move off Framer. With this branch's commands, **all five pass.**

One detail worth calling out because it justifies a choice that might otherwise look incidental: simply adding `-L` would *also* have made step 1 pass, but wrongly. `curl -L` on `/templates/flux_schnell` lands on `/workflows/flux_schnell/`, which is a 460-byte `noindex` meta-refresh stub, not the real page — and `curl` does not follow meta-refresh. Moving the fixture to the shareId form `/workflows/flux_schnell-75fb83052003/` is what makes the assertion meaningful. This PR does that.

Only caveat, and not a blocker: `75fb83052003` is a live shareId hardcoded into the check, so this breaks if that workflow is ever removed. It is present in the production hub index and in the i18n artifacts today. Worth a comment in the file at some point, or picking a fixture with a stronger guarantee of permanence.

Worth merging — a check that fails on every commit stops being read, and this one has been failing long enough to have trained everyone to skip it.

PR #985, opened 2026-07-01
feat: design use cases and model seo pages

MaanilVerma:
## Summary

Adds a new SEO landing surface to the workflow hub — per-model and per-use-case pages that turn the existing template catalog into ready-to-run, indexable collection pages — and folds in the hub redesign work (hero, creator cards, category overflow, FAQ) that shipped alongside it. Model and use-case pages are generated from the catalog at build time, gated so only substantive pages get indexed, and wired into the sitemap and hreflang graph.

## Changes

**What**

- New model landing pages (`/workflows/model/<name>/`) that group the catalog into model families and render a hero, workflow grid, capabilities, "why Comfy", FAQ, and a keep-exploring rail. Model-name variants redirect into a single canonical family slug.
- New use-case landing pages (`/workflows/use-cases/<slug>/`) built from a curated registry that resolves each page to a real template cluster by model/tag filters; a page with no matching templates never routes.
- Both surfaces get index pages that share one shell, plus editorial copy loaded per slug and a brand-safety denylist that fails the build if a denied term reaches a page's slug, keyword, or title.
- Indexation is gated: a page is `noindex` unless it's backed by real templates and its copy passed the quality bar, and the sitemap advertises exactly the indexable set — so a thin or denied page can never enter the sitemap. Model-page hreflang is reciprocal across locales, and every JSON-LD graph (CollectionPage, Breadcrumb, FAQ) is escaped through one shared serializer.
- Provider/model logo resolution is consolidated into a single module and rendered as static badges across landing pages and workflow cards; badges dedupe by resolved logo so aliases like Google/Gemini show once.
- Hub redesign carried in the same branch: restyled hero, creator cards that always show avatar + name with a broken-image fallback, an inline "+N" category overflow pill, a plus/minus FAQ toggle, and shared Button/ButtonPill variants rendered as links.

**Breaking**

None. Existing template detail, creator, and download routes are unchanged; the old `provider-logos` module is replaced by the consolidated one with its behavior preserved and re-tested.

## Review Focus

- **The indexation contract is the crux** — the route's `noindex` decision and the sitemap's indexable set are derived from the same conditions (real cluster + quality-passed + not brand-denied). Worth confirming they can't drift: a page that renders indexable must be in the sitemap, and a noindex page must not be. Same question for model-page hreflang reciprocity between the default and localized routes.
- **Content-quality engine was intentionally dropped, not shipped half-wired** — an earlier draft included a validator that nothing called; it was removed rather than left as dead surface. Indexation currently gates on the `qualityFailed` flag from editorial content, not a live validator. If a build-time validator is wanted, that's a deliberate follow-up.
- **Scope boundaries:** editorial JSON content (model/use-case copy) is data, not reviewed as code here. Localized model routes are non-indexable thin mirrors of the English page by design.
- Large diff because it also carries the redesign commits and a round of self-review fixes (JSON-LD escaping, in-page CTA no longer opening a new tab, dead-code removal, a11y on FAQ/badges).

## Testing

Unit tests cover the pure logic that decides what ships and how it's labeled — logo resolution and precedence, route/URL helpers, keyword title-casing, and thumbnail selection — since those feed both the rendered pages and the sitemap. Rendered pages and hreflang/sitemap output are verified by the existing build + SEO audit rather than duplicated here.

- `tests/unit/model-logos.test.ts` — logo resolution, alias dedupe by resolved logo, and logos-over-models precedence (replaces `provider-logos.test.ts`).
- `tests/unit/schema.test.ts` — `titleCaseKeyword` including the AI/acronym cases.
- `tests/unit/routes.test.ts`, `tests/unit/urls.test.ts` — the new model/use-case route helpers and pricing-CTA UTM construction.
- `tests/unit/video-thumbnail.test.ts` — still-thumbnail selection after centralizing it in `media-utils`.

Gate: `pnpm test` → 236 passed (20 files); `astro check` and `eslint` clean on the changed scope. Manual: verified a model page, a use-case page, both index pages, and a redirecting model variant render, and that a zero-cluster use-case does not route.

## ScreenRecord
 

https://github.com/user-attachments/assets/08d7ad14-9fb3-402f-a470-63882320d011


MaanilVerma:
@coderabbitai review

site/src/components/workflow-pages/LandingPageBody.astro
```diff
@@ -0,0 +1,214 @@
...
+  logos: card.logos ?? [],
+  cta: card.cta,
+}));
+const capabilityCards = enrichedCards.slice(2);
+const hasCapabilities = capabilityCards.length > 0;
+const capabilitiesEyebrow = t('useCase.capabilities', locale);
+const capabilitiesHeading = t('useCase.whatYouCanCreate', locale);
+// Use-case pages take the authored intro; model pages keep the modelSpec summary.
+const capabilitiesSummary = hasStyles ? capabilitiesIntro : modelSpec?.summary;
+
```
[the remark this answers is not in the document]
MaanilVerma: Ignore we don't need it. Product decision.

site/astro.config.mjs
```diff
@@ -8,53 +8,85 @@ import path from 'node:path';
...
+const contentPasses = (dir, slug) => {
+  const contentPath = path.join(dir, `${slug}.json`);
+  if (!fs.existsSync(contentPath)) return false;
+  try {
+    return JSON.parse(fs.readFileSync(contentPath, 'utf-8')).qualityFailed !== true;
+  } catch {
+    return false;
+  }
+};
+const indexableModelSlugs = new Set(
```
dante01yoon, line 53: issue: The sitemap's indexable set and the routes' noindex decision — the PR's stated indexation contract — are derived from different datasets with hand-mirrored predicates, so they can drift silently. In a production build the routes compute noindex from the live hub API index (`loadSerializedTemplates`; hub-api.ts throws rather than falling back), while `indexableModelSlugs`/`indexableUseCaseSlugs` here derive from the repo-synced `src/content/templates/*.json` snapshot with a raw `qualityFailed !== true` check. A model family near `MIN_CLUSTER_SIZE`/`MIN_CLUSTER_USAGE` that qualifies in one dataset but not the other yields either an indexable page missing from the sitemap or a noindex page listed in it, with no build-time failure (`validate-sitemap` checks XML structure and status codes, not noindex agreement). And if the prebuild sync didn't run, the `fs.existsSync` guard silently empties `contentTemplates` and drops every model/use-case URL from the sitemap. Suggest deriving both sides from one shared predicate over one dataset, or a postbuild assertion that sitemap membership equals the rendered-indexable set.

site/src/lib/workflow-pages/seo-page.ts
```diff
@@ -0,0 +1,113 @@
...
+import { t } from '../../i18n/ui';
+import type { Locale } from '../../i18n/config';
+import type { ModelGroup } from './model-groups';
+import type { SeoPageDef } from './use-cases';
+import type { GeneratedSeoContent } from './schema';
+import { isIndexable } from './governance';
+
+/** Indexable only when rich and backed by real templates. Shared by both kinds. */
+function pageIndexable(isRich: boolean, templateCount: number): boolean {
+  return isIndexable({ clusterSize: templateCount, humanEdited: true, qualityPassed: isRich });
```
dante01yoon: issue: `pageIndexable` hardcodes `humanEdited: true` at `isIndexable`'s only call site, so the human-review gate can never fail. Setting `"humanEdited": false` in any landing JSON (the zod schema in content/config.ts explicitly parses the field) still renders the page indexable and keeps it in the sitemap — and `GeneratedSeoContent` doesn't expose the field, so no caller can wire it in. Either thread `content.humanEdited` through (add it to `GeneratedSeoContent`, read it here), or delete the field from the schema/JSONs and drop the parameter from `isIndexable`.

site/src/components/workflow-pages/LandingHero.astro
```diff
@@ -0,0 +1,112 @@
...
+        class="mt-6 max-w-[46ch] font-display text-[1.0625rem] font-light leading-[1.6] text-content"
+      >
+        {description}
+      </p>
+      {meta && <p class="mt-3 font-display text-sm font-light text-hub-muted">{meta}</p>}
+
+      <div class="mt-9 flex flex-wrap items-center gap-2">
+        {
+          ctaButtons.map((cta) => (
+            <Button
```
dante01yoon, line 77: issue: The above-the-fold "Try on Cloud" hero CTA on every model/use-case landing page carries none of the `run-cloud-btn` / `data-location` / `data-template` hooks that PostHogAnalytics.astro's delegated click handler matches (and autocapture is disabled), so `hub:signup_cta_clicked` / `hub:run_button_clicked` never fire from it — the primary conversion action on these pages is invisible except via destination-side UTM. The pricing CTA in CtaBanner.astro has the same gap (no class hooks, no inline `va` handler). This repeats the untracked-CTA measurement gap from the #950 navbar redesign. Fix: add `run-cloud-btn` + `data-location` (+ `data-template` when a lead shareId exists) — `Button` passes `class`/`data-*` through, as ClosingCta.astro demonstrates.

site/src/components/workflow-pages/ClosingCta.astro
```diff
@@ -0,0 +1,62 @@
...
+      {t('template.startNow', locale)}
+    </p>
+    <Button
+      as="a"
+      href={href}
+      target="_blank"
+      rel="noopener noreferrer"
+      variant="brand-solid"
+      size="nav"
+      class:list={[analytics && 'usecase-cta', 'mt-8']}
```
dante01yoon: issue: The closing CTA omits the `run-cloud-btn` class, so the PostHog delegate never emits `hub:signup_cta_clicked` from it; and on model pages ([name].astro passes no `analytics` prop) the `usecase-cta` class, data attributes, and inline `va` script are all conditional on that prop, so the button carries no tracking at all. Clicks on the closing band of e.g. /workflows/model/wan/ record zero events in both PostHog and Vercel Analytics. Fix: add `run-cloud-btn` + a `data-location` (as TryOnCloudButton.astro and HeaderMain.vue do), independent of the `analytics` prop.

site/src/lib/routes.ts
```diff
@@ -11,11 +10,18 @@ const localize = (path: string, locale?: string) =>
...
+export const modelsIndexPath = (locale?: string) => localize(MODEL_BASE, locale);
+export const modelPath = (name: string, locale?: string) =>
+  localize(`${MODEL_BASE}${name}/`, locale);
 export const tagPath = (tag: string, locale?: string) =>
   localize(`/workflows/tag/${tagSlug(tag)}/`, locale);
+const USE_CASES_BASE = '/workflows/use-cases/';
+export const useCasesIndexPath = (locale?: string) => localize(USE_CASES_BASE, locale);
+export const useCasePath = (slug: string, locale?: string) =>
+  localize(`${USE_CASES_BASE}${slug}/`, locale);
 export const creatorPath = (username: string, locale?: string) =>
```
dante01yoon, line 24: issue: `creatorPath` changed from `slugify` to `encodeURIComponent`, but `getCreatorUrl` in SearchPopover.vue (line ~293) — a file this PR edits — and `creatorUrl` in TemplateDetailPage.astro (line ~52) still slugify. For any username with uppercase, underscore, dot, or space (e.g. `enigmatic_e`), hub cards link `/workflows/enigmatic_e/` while the search popover links `/workflows/enigmatic-e/`; the raw-segment lookup in [username].astro finds nothing for the slugified form and redirects to /workflows/, so those creator links silently dump users on the index page.

site/src/lib/workflow-pages/use-case-resolver.ts
```diff
@@ -0,0 +1,111 @@
...
+}
+
+// Qualifying model groups, memoized per catalog: relatedness only links rich,
+// indexable model pages, never a bare noindex grid.
+const qualifyingGroupsCache = new WeakMap<SerializedTemplate[], ModelGroup[]>();
+
+export function qualifyingGroups(catalog: SerializedTemplate[]): ModelGroup[] {
+  let groups = qualifyingGroupsCache.get(catalog);
+  if (!groups) {
+    groups = deriveModelGroups(catalog).filter((group) => group.qualifies);
```
dante01yoon: issue: `qualifyingGroups`'s comment promises links go only to "rich, indexable model pages, never a bare noindex grid", but the filter checks only `group.qualifies` — weaker than the page's actual indexability predicate (`qualifies && content && !qualityFailed`). Priority families in the catalog with no editorial JSON (Veo, GPT-Image, Seedream, Hunyuan3D, Nano Banana, Stable Audio…) pass it, so /workflows/model/ index cards and use-case related rails funnel users and crawl equity to noindex bare-grid pages, contradicting the inline contract. Fix: filter through the `isModelPageRich` predicate — which needs a Node-importable content-existence check, since content-loaders.ts is astro:content-only.

site/src/pages/workflows/[username].astro
```diff
@@ -233,7 +237,6 @@ const detailBreadcrumb = detailData
         locale={locale}
```
Comment (dante01yoon) [inline site/src/pages/workflows/[username].astro:237]: question: The `hideAuthor` suppression was removed from HubWorkflowCard/WorkflowGrid and its four call sites, so creator-profile and creators-directory grids now render each creator's own avatar + name on every card, duplicating the page's own header/sidebar. Intentional in the redesign, or a dropped invariant?

site/src/i18n/ui.ts
```diff
@@ -79,14 +79,26 @@ export const UI_STRINGS: Record<Locale, Record<string, string>> = {
     'labels.createdBy': 'CREATED BY',
     'labels.lastUpdate': 'LAST UPDATE',
     'labels.categories': 'CATEGORIES',
-    'cta.tryCloud': 'Try on Comfy Cloud',
+    'cta.tryCloud': 'Try Comfy Cloud',
+    'cta.viewAll': 'View All',
+    'model.aboutEyebrow': 'About This AI Model',
+    'model.heroEyebrow': 'AI Model',
+    'model.metaH1': '{label} ComfyUI Workflows',
+    'model.metaTitle': '{label} ComfyUI Workflows — Free Templates',
```
dante01yoon, line 87: question: The new SEO meta keys (`model.metaH1/metaTitle/metaDescription/subheading`, `useCase.metaDescription/subheading/banner*/workflowsCount*`) exist only in the `en` block, while `model.heroEyebrow` added in the same PR is translated into all other locales — so localized model/use-case pages ship an English <title>, meta description, and h1 inside translated chrome via the `en` fallback. Since the PR describes localized routes as non-indexable thin mirrors by design: is English meta there accepted, or should these keys be translated like `heroEyebrow`?

site/src/pages/workflows/model/[name].astro
```diff
@@ -4,80 +4,160 @@ import BaseLayout from '../../../layouts/BaseLayout.astro';
...
+  redirectTo?: string;
 };
-
 const locale = getLocaleFromPath(Astro.url.pathname);
-const basePath = modelPath(name!);
 
-const title = `${modelName} ComfyUI Workflows - Free Templates`;
-const description = `Browse ${templates.length} free ComfyUI workflow templates using ${modelName}. Ready-to-use AI generation workflows for image, video, and audio creation.`;
+if (redirectTo) {
+  return Astro.redirect(localizeUrl(modelPath(redirectTo), locale), 301);
```
Comment (dante01yoon) [inline site/src/pages/workflows/model/[name].astro]: question: This route is prerendered, so `Astro.redirect(..., 301)` for variant slugs is emitted as a static meta-refresh page served with HTTP 200 (it does carry noindex + canonical), while the [locale] route (prerender=false) returns a genuine 301 for the same variant — inconsistent status codes across locales and weaker link-equity consolidation on the surface this PR exists to optimize. If real 301s are wanted, platform-level redirects (vercel.json or astro.config `redirects`) would cover the static route.

site/src/lib/workflow-pages/card-images.ts
```diff
@@ -0,0 +1,143 @@
...
+  for (const template of pool) {
+    if (used.has(template.name)) continue;
+    const score = relevance(cardTokens, template);
+    if (score > bestScore) {
+      best = template;
+      bestScore = score;
+    }
+  }
+  // No semantic match → highest-usage unused template in the pool.
+  return best ?? pool.find((t) => !used.has(t.name)) ?? null;
```
dante01yoon: suggestion: `best ?? pool.find((t) => !used.has(t.name))` returns an arbitrary highest-usage unused grid template on zero token overlap, so the documented resolution order (step 2: "content-matched, unused, from the whole-dataset fallback") is unreachable whenever the grid has any unused template left — a card with no token overlap shows an unrelated grid image, marks it used, and pushes a later genuinely-matching card onto a worse one. Move the zero-relevance grid fallback after the fallback-pool content match, or amend the doc comment to match the actual order.

site/src/components/workflow-pages/LandingHero.astro
```diff
@@ -0,0 +1,112 @@
...
+].map((cta) => {
+  const external = /^https?:/i.test(cta.href);
+  return {
+    ...cta,
+    target: external ? '_blank' : undefined,
+    rel: external ? 'noopener noreferrer' : undefined,
+  };
+});
+
+const heroTemplate = templates.find((template) => hasStillThumbnail(template.thumbnails));
```
dante01yoon: suggestion: `templates.find((t) => hasStillThumbnail(t.thumbnails))` is re-implemented here, in workflows/model/[name].astro (~141), and in workflows/use-cases/[slug].astro (~97), and the pages' `reservedTemplates` exclusion assumes their local copy matches the hero's internal pick. Extract a shared `firstTemplateWithStill()` next to `firstStillAcross` in media-utils.ts — or pass the resolved hero template in as a prop (use-cases/[slug].astro already computes `heroTemplate`) — so a future change to hero selection can't silently desynchronize the reservation.

site/src/pages/workflows/model/[name].astro
```diff
@@ -4,80 +4,160 @@ import BaseLayout from '../../../layouts/BaseLayout.astro';
...
+const heroCtaUrl = modelCtaUrl('hero');
+const footerCtaUrl = modelCtaUrl('footer');
+
+const modelsUrl = absoluteUrl(localizeUrl('/workflows/model/', locale));
+const breadcrumbTrail = [
+  { label: t('labels.models', locale), href: localizeUrl('/workflows/model/', locale) },
+];
+
+const structuredData = buildCollectionPageJsonLd({ name: h1, description, url: canonicalUrl });
+const breadcrumbStructuredData = buildBreadcrumbJsonLd([
```
Comment (dante01yoon) [inline site/src/pages/workflows/model/[name].astro]: suggestion: The 4-crumb BreadcrumbList is built inline with `buildBreadcrumbJsonLd` (near-identically duplicated in use-cases/[slug].astro) instead of extending `buildWorkflowBreadcrumb` in structured-data.ts, whose documented purpose is keeping default and localized routes' JSON-LD identical. The drift it exists to prevent is already visible at PR head: the [locale] model route emits 3 crumbs while this route emits 4, so the two hreflang-linked variants of the same page carry different BreadcrumbList JSON-LD.

site/src/pages/workflows/use-cases/[slug].astro
```diff
@@ -0,0 +1,276 @@
...
+    name: t('breadcrumb.workflows', locale),
+    item: absoluteUrl(localizeUrl('/workflows/', locale)),
+  },
+  { name: t('template.useCases', locale), item: useCasesUrl },
+  { name: h1 },
+]);
+// Gate on rich content like the model page, so a noindex/thin page never ships FAQ rich-results.
+const faqStructuredData = richContent ? buildFaqJsonLd(richContent.faqItems) : null;
+
+const fallbackImages = [...templates, ...fallbackPool]
```
Comment (dante01yoon) [inline site/src/pages/workflows/use-cases/[slug].astro]: suggestion: The related-rail image fallback (flatten remaining thumbnails, exclude hero, mutating `takeFallbackImage` queue) re-implements the used-set + fallback-pool assignment that card-images.ts `assignCardTemplates` already provides, and the model route (~[name].astro:145-152) hand-rolls a third, positionally-indexed variant. One shared helper (or `assignCardTemplates` with `used` seeded with the hero) keeps the two rails from drifting and gets duplicate-image prevention for free.

dante01yoon (commented):
Review of the SEO landing surface + hub redesign follow-ups. The core architecture is sound and the PR's own review-focus notes were a good map — but the two named cruxes both have real gaps: the sitemap and the routes derive the indexation contract from **different datasets** with hand-mirrored predicates, and the `humanEdited` governance gate is hardcoded on. Also, given the untracked-CTA measurement gap the #950 navbar redesign caused, note that the new landing pages' hero and banner CTAs ship with the same gap (only ClosingCta is partially instrumented).

6 blocking issues, 3 questions, 12 non-blocking suggestions/nits. One observation folded here rather than inline: every new SEO page renders `HubNavbar` with `showSearch=true`, inlining the full serialized catalog into the SearchPopover island props — it follows the existing hub pattern, but it's worth weighing on pages whose whole point is fast indexable HTML.

site/src/lib/workflow-pages/use-cases.ts
```diff
@@ -0,0 +1,235 @@
...
+
+export interface SeoPageDef {
+  /** Canonical kebab-case URL segment, used verbatim (not re-slugified). */
+  slug: string;
+  title: string;
+  h1: string;
+  /** Target keywords; `keywords.primary` drives copy + meta fallback. */
+  keywords: KeywordModel;
+  /** Word-count band the quality validator enforces, by keyword difficulty/value. */
+  wordCountTier: WordCountTier;
```
dante01yoon: suggestion: `wordCountTier` is dead config on all 10 SEO_PAGES entries — the doc comment says "the quality validator enforces" it, but nothing reads the field (generate-ai.ts enforces a flat 100-word floor and doesn't process SEO_PAGES), so an author choosing `long` for a high-value page gets no enforcement. Since the content-quality validator was deliberately dropped per the PR description, delete `wordCountTier` (and the `WordCountTier` export in schema.ts) until the follow-up validator actually consumes it.

site/src/components/workflow-pages/cards/MediaCard.astro
```diff
@@ -0,0 +1,89 @@
...
+
+const { eyebrow, title, text, image, logos = [], cta } = Astro.props;
+---
+
+<article
+  class="group/pill-trigger group flex h-full flex-col rounded-4xl bg-hub-surface p-2 transition-colors duration-300 hover:bg-hub-surface-hover"
+>
+  {
+    image &&
+      (cta?.href ? (
```
dante01yoon: suggestion: MediaCard and SplitCard (~line 63) each duplicate an entire `cta?.href ? <a>…</a> : <div>…</div>` block whose branches contain identical children (~15 copy-pasted lines per component). AppCard.astro already has the pattern that collapses this: `const ImageTag = cta?.href ? 'a' : 'div'` plus a conditional attribute spread, writing the children once.

site/src/components/workflow-pages/ProductShowcase.vue
```diff
@@ -0,0 +1,198 @@
...
+        <div
+          :class="
+            cn(
+              'rounded-5xl relative flex w-full items-center justify-center overflow-hidden p-0.5',
+              isVisible && 'animate-border-spin'
+            )
+          "
+        >
+          <div class="bg-page relative size-full overflow-hidden rounded-[calc(2.5rem-2px)]">
+            <video
```
dante01yoon, line 87: suggestion: The first feature's autoplay .webm is fetched, decoded, and played eagerly on every landing page even below the fold (`client:visible` defers only hydration, and muted autoplay runs under `display:none`), and phones pay for it twice — once in the hidden desktop copy, once in the visible mobile copy. Cheaper: `preload="none"` without SSR autoplay, `play()` from the already-present `useIntersectionObserver`, and one shared video element instead of the desktop/mobile duplicate.

site/src/pages/[locale]/workflows/model/[name].astro
```diff
@@ -34,37 +39,54 @@ const typedLocale = locale as Locale;
...
 const allTemplates = await loadSerializedTemplates(() => getCollection('templates'));
 
-// Find templates matching this model slug
-const matchingTemplates = allTemplates.filter((tmpl) =>
-  tmpl.models.some((m: string) => slugify(m) === name)
-);
+// Resolve the model family via the same grouping the English page uses, so both
+// routes agree on canonical slugs, variant redirects, and display label — the
+// hreflang cluster between them is only reciprocal if the slugs resolve identically.
+const groups = deriveModelGroups(allTemplates);
```
Comment (dante01yoon) [inline site/src/pages/[locale]/workflows/model/[name].astro]: suggestion: This prerender=false route re-serializes the full catalog (`loadSerializedTemplates` maps the entire index into a fresh array) and re-runs `deriveModelGroups` over it on every request, for inputs that are constant for the life of `indexCache`. Memoize the serialized catalog (and optionally the derived groups) at module level, matching the existing `indexCache`/`profileCache` pattern in hub-api.ts.

site/src/pages/workflows/model/[name].astro
```diff
@@ -4,80 +4,160 @@ import BaseLayout from '../../../layouts/BaseLayout.astro';
...
 
-  for (const template of allSerialized) {
-    for (const model of template.models) {
-      const normalized = slugify(model);
-      if (!modelMap.has(normalized)) {
-        modelMap.set(normalized, []);
-      }
-      modelMap.get(normalized)!.push(template);
+  // Astro rejects duplicate params.name, so every canonical + variant slug must be unique.
+  const claimed = new Set<string>();
```
Comment (dante01yoon) [inline site/src/pages/workflows/model/[name].astro:56]: suggestion: Adding one model page currently takes coordinated edits in three files keyed by a slug computed elsewhere (PRIORITY_MODELS/FAMILY_RULES, a landing JSON whose filename must equal `slugify(family label)`, MODEL_TO_LOGO), and nothing fails when they drift — e.g. add "hunyuan video" to PRIORITY_MODELS plus `hunyuan-video.json`, but the catalog tag "HunyuanVideo" groups to slug `hunyuanvideo`, so the authored copy is silently dead and the page ships as a thin noindex grid while the build stays green. A build-time orphan assertion (every landing JSON slug must match a canonical generated group slug, in both this route and the [locale] one) would catch the whole class.

site/src/components/workflow-pages/CtaBanner.astro
```diff
@@ -0,0 +1,29 @@
...
+    <div class="flex flex-col gap-8 font-display text-[1.0625rem] leading-[1.6]">
+      <p class="font-light">{body}</p>
+      {kicker && <p class="font-semibold">{kicker}</p>}
+    </div>
+  </div>
+  <a
+    href={cta.href}
+    target="_blank"
+    rel="noopener noreferrer"
+    class="inline-flex h-12 shrink-0 items-center justify-center rounded-2xl bg-page px-5 font-display text-sm font-extrabold uppercase tracking-[0.05em] text-brand transition-opacity hover:opacity-90"
```
dante01yoon: nit(non-blocking): The banner CTA hand-rolls a button-styled anchor with a long inline class string instead of the shared `Button as="a"` that ClosingCta and LandingHero use, and it omits the shared `buttonVariants` base — so on keyboard focus it shows the browser-default outline instead of the 3px ring the sibling CTAs get. Needs a small inverse-on-brand variant rather than being a drop-in swap, but keeps future Button fixes flowing here.

site/src/components/workflow-pages/sections/CapabilitiesSection.astro
```diff
@@ -0,0 +1,54 @@
...
+export interface CapabilityCard {
+  title?: string;
+  text: string;
+  image?: CardImage | null;
+  logos?: ModelBadge[];
+  cta?: { label: string; href: string } | null;
+}
+
+interface Props {
+  locale: Locale;
```
dante01yoon: nit(non-blocking): `locale: Locale` is declared in Props (and passed from LandingPageBody) but never used — the labels arrive pre-translated. Drop the prop, the `Locale` import, and the pass-site.

site/src/lib/workflow-pages/card-images.ts
```diff
@@ -0,0 +1,143 @@
...
+  let score = 0;
+  for (const token of a) if (b.has(token)) score += 1;
+  return score;
+}
+
+/**
+ * Relevance of a template to a card's words. Tags and models are the strongest
+ * topical signal, so they weigh more than incidental title/description words.
+ */
+function relevance(cardTokens: Set<string>, template: MatcherTemplate): number {
```
dante01yoon, line 89: nit(non-blocking): `relevance()` re-tokenizes every template's tags/models/title/description for every (card, template) pair, and `withStill()` re-filters (a no-op after `buildFallbackPool`) and re-sorts the whole-catalog pool twice per page. Memoizing per-template token sets and sorting once in `buildFallbackPool` keeps this linear; build-time cost only, but it runs for every landing page.

site/src/lib/workflow-pages/seo-page.ts
MaanilVerma: Let me revisit this , original idea was that we will create a programmatic seo pipeline, and idea was to ignore already generated articles

site/src/pages/workflows/[username].astro
Comment (MaanilVerma) [reply site/src/pages/workflows/[username].astro:237]: Yes, this is intentional on the design side.

site/src/i18n/ui.ts
MaanilVerma: Yes, English only for now! Follow up PR will have content for all the locales

dante01yoon (approved):

PR #992, opened 2026-07-06
 feat(site): add HowTo, SoftwareApplication, and ItemList schema to SEO pages

MaanilVerma:
### Summary
The model and use-case landing pages now emit additional schema.org structured data or crawlers and LLM consumers: step-by-step HowTo metadata, a SoftwareApplication entity, and ItemList entries for visible workflow grids.

### Changes
**What:**
- Rich model and use-case detail pages now describe their "how to use" steps and frame themselves as a runnable tool in structured data, so search engines can surface a how-to snippet and a software listing for them.
- The software listing deliberately makes **no price or rating claim** — Comfy Cloud isn't free and we have no rating data, so asserting either would be false structured data that can get a page flagged.
- Every detail page now enumerates its workflow grid as a machine-readable list, and the model/use-case index pages enumerate their cards the same way — so the structured data mirrors exactly what's visible on the page, each entry linking to its own detail page.
- All of this is gated to the rich, indexable pages only. Thin pages that are already `noindex` stay minimal (breadcrumb + collection), so we never advertise rich results for a page we're keeping out of the index.
- New markup is emitted only when the backing data actually exists — no how-to without steps, no empty lists — mirroring the existing FAQ builder's "skip when empty" behavior.

**Breaking:** None. The shared collection-page builder stays byte-identical for its existing callers when no list is passed; breadcrumb, FAQ, canonical, hreflang, and the sitemap's noindex exclusions are all untouched.

### Review Focus
- The one decision worth a look: the software-listing schema omits `offers`/`aggregateRating` on purpose. If a reviewer expects a free-tier Offer, that's the deliberate call — see the rationale comment in `structured-data.ts`.
- All JSON-LD goes through the existing hardened serializer (`serializeJsonLdForScript`); no raw `JSON.stringify` reaches a `<script>` sink. Worth confirming the list URLs/thumbnail URLs (which now appear in indexable markup) carry no signed tokens — traced clean, they're the same public CDN URLs already rendered as ``.
- Scope boundary: VideoObject was considered and deliberately **not** included — an animated grid thumbnail isn't the page's primary content, so marking it up as a video would misrepresent the page. WebSite/Organization are also intentionally out — those are domain-level entities owned by the `comfy.org` (Framer) root, not this `/workflows` sub-app.
- Built output was inspected directly: rich pages carry Breadcrumb + Collection(+list) + FAQ + HowTo + SoftwareApplication; thin pages carry only Breadcrumb + Collection; index pages carry the card list. No VideoObject anywhere.

### Testing
Unit tests pin the builder contracts directly, since that's where the correctness lives (empty-input handling, the no-price/no-rating guarantee, and script-safe escaping) — and the rendered HTML was verified against a real build to confirm the wiring and gating behave end-to-end.

- `structured-data.test.ts` — asserts how-to returns nothing for empty/blank steps and renumbers survivors; the software listing never emits a price or rating and drops an empty feature list; the collection builder stays byte-identical without a list and nests a correctly-positioned list with one; and that `</script>`/`&` inside step text is escaped, not passed through.
- Gate: `astro check` 0 errors · eslint clean · `vitest` 246 passed.
- Manual: built the site and confirmed each of the four page types emits the expected `@type` set, all JSON-LD parses as valid JSON, the list URLs are absolute and single-locale, and thin `noindex` pages stay minimal.


site/src/lib/structured-data.ts
```diff
@@ -93,18 +93,92 @@ export function buildFaqJsonLd(faqItems: FaqItem[] | undefined) {
...
     ...(params.inLanguage ? { inLanguage: params.inLanguage } : {}),
+    ...itemList,
+  };
+}
+
+/**
+ * schema.org `HowTo` for a page's ordered "how to use" steps, or `null` when
+ * there are no non-blank steps (so the caller can skip an empty graph).
+ */
+export function buildHowToJsonLd(params: {
```
dante01yoon, line 144: question (non-blocking): The PR frames these builders as surfacing a Google "how-to snippet" and a "software listing" rich result, but Google deprecated HowTo rich results in Aug–Sep 2023 (removed the search appearance and the docs), and `SoftwareApplication` isn't eligible for a software-app rich result without `offers.price` + `aggregateRating`/`review` (deliberately omitted here). The markup is still valid schema.org and harmless — useful for LLMs/other crawlers — so no need to remove it, but the stated Google rich-result benefit won't materialize. Worth updating the description to reflect the real value (machine-readable metadata for non-Google consumers).

dante01yoon (commented):
Solid, well-tested SEO structured-data addition. Verified independently: the builders correctly skip empty/blank inputs, the new `ItemList` mirrors exactly the templates fed to the visible `<WorkflowGrid>` (same `templates` set on both the model and use-case pages), `numberOfItems` matches the emitted list, and all JSON-LD flows through the hardened `serializeJsonLdForScript`. No correctness issues. One non-blocking note on the search-visibility rationale. (0 blocking, 1 non-blocking.)

dante01yoon:
please resolve conflicts

site/src/lib/structured-data.ts
MaanilVerma: Good catch, and confirmed against Google's own docs:

HowTo rich results were [deprecated in Sep 2023](https://developers.google.com/search/blog/2023/08/howto-faq-changes) — the desktop appearance and the Search Console report were both removed. Google explicitly notes there's no need to proactively remove the markup.

SoftwareApplication isn't eligible for a software-app rich result without offers.price + aggregateRating/review, which we deliberately omit (asserting ratings/prices we don't have would violate their guidelines).

So the markup stays — it's valid schema.org and useful as machine-readable metadata for LLMs and other non-Google crawlers — but you're right that the framing over-claimed a Google rich-result benefit that won't materialize. Fixed the builder + call-site comments to say that plainly in cf7e8ab1 (comment-only, no markup change).

site/src/pages/workflows/model/[name].astro
```diff
@@ -110,13 +113,33 @@ const footerCtaUrl = modelCtaUrl('footer');
 const modelsUrl = absoluteUrl(modelsIndexPath(locale));
 const breadcrumbTrail = [{ label: t('labels.models', locale), href: modelsIndexPath(locale) }];
 
-const structuredData = buildCollectionPageJsonLd({ name: h1, description, url: canonicalUrl });
+const structuredData = buildCollectionPageJsonLd({
+  name: h1,
+  description,
+  url: canonicalUrl,
+  items: buildTemplateItemListEntries(templates, locale),
```
Comment (dante01yoon) [inline site/src/pages/workflows/model/[name].astro]: suggestion: The localized sibling route was left behind — `src/pages/[locale]/workflows/model/[name].astro:83` still emits a bare CollectionPage although it renders the same template grid, can be indexable via the same `isModelPageIndexable` gate, and is hreflang-clustered with this page. After this PR, `/workflows/model/flux/` enumerates every template while `/ja/workflows/model/flux/` has no `mainEntity`. One-line fix there: `items: buildTemplateItemListEntries(matchingTemplates, typedLocale)` — both values already in scope, and the new unit test "locale-prefixes the detail URL for a non-default locale" isn't exercised by any route today. (HowTo/SoftwareApplication should *not* be mirrored — those correctly stay gated to pages that render the editorial copy.)

site/src/pages/workflows/use-cases/[slug].astro
```diff
@@ -127,6 +132,21 @@ const breadcrumbStructuredData = buildWorkflowBreadcrumb(locale, { name: h1 }, [
...
+const howToStructuredData = qualityContent
+  ? buildHowToJsonLd({
+      name: `How to create ${titleCaseKeyword(keyword)} with ComfyUI`,
+      steps: qualityContent.howToUse,
+      description: subheading,
+    })
+  : null;
+const softwareAppStructuredData = qualityContent
+  ? buildSoftwareApplicationJsonLd({
+      name: def.title,
```
Comment (dante01yoon) [inline site/src/pages/workflows/use-cases/[slug].astro]: suggestion: `SoftwareApplication.name` here gets the raw title-tag string (e.g. `AI Headshot Generator — ComfyUI Workflows` from use-cases.ts), while the model page passes the clean `h1` — the two pages name the application inconsistently. `h1` is already in scope (line 92) and already used for the breadcrumb name at line 128; passing it here matches the model page and keeps the brand suffix out of the entity name.

dante01yoon (commented):
Adds HowTo/SoftwareApplication/ItemList JSON-LD builders and wires them into the model, use-case, and index SEO pages. No blocking correctness regressions found — the no-items collection path is byte-identical as claimed, all sinks go through the hardened serializer, and the empty-input gating holds up. Leaving 5 suggestions (locale-route parity, ItemList vs the 30-card SSR grid, a pre-existing contradictory SoftwareApplication on the detail page, naming/placement), 1 question about ItemList gating on noindex pages, and 4 non-blocking nits (including a Prettier failure that will surface in site-ci once the stack targets main).

site/src/lib/workflow-pages/index-cards.ts
```diff
@@ -3,13 +3,37 @@
...
+/**
+ * Map a page's workflow grid to `ItemList` entries for its CollectionPage —
+ * each template linking to its own detail page, mirroring the visible grid.
+ * Templates with no derivable detail URL are skipped.
+ */
+export function buildTemplateItemListEntries(
+  templates: SerializedTemplate[],
+  locale?: string
+): ItemListEntry[] {
+  return templates.flatMap((tpl) => {
```
dante01yoon: suggestion: This enumerates the full templates array, but the grid these pages render (WorkflowGrid.vue) only ships the first 30 cards in the initial HTML (`displayCount = ref(30)` + slice) and sorts client-side — so on large families (Wan: 37+20+19 templates, Nano Banana Pro: 43) the ItemList claims more items, in a different order, than the served page contains, and the doc comment's "mirroring the visible grid" isn't accurate. Either cap entries to the initial render count via a constant shared with WorkflowGrid.vue, or reword the comment to state the deliberate full-catalog enumeration.

site/src/lib/structured-data.ts
```diff
@@ -95,18 +95,94 @@ export function buildFaqJsonLd(faqItems: FaqItem[] | undefined) {
...
+  };
+}
+
+/**
+ * schema.org `SoftwareApplication` framing a landing page as a runnable tool —
+ * machine-readable metadata, not a Google rich result (that needs `offers` and
+ * `aggregateRating`, which we never assert: Comfy Cloud isn't free and no rating
+ * data exists).
+ */
+export function buildSoftwareApplicationJsonLd(params: {
```
dante01yoon, line 171: suggestion: This is now the second source of truth for the SoftwareApplication shape — `src/pages/workflows/[slug].astro:148-156` hand-rolls one with the same `applicationCategory`/`operatingSystem` strings plus the `offers: { price: '0' }` claim this builder's comment explicitly disclaims ("Comfy Cloud isn't free"). After merge the site contradicts itself on whether the product is free, and edits to the constants here silently diverge from the detail page. Consider refactoring the detail page to call this builder (dropping the $0 offer) — fine as a follow-up in the stack.

site/src/pages/workflows/use-cases/[slug].astro
```diff
@@ -119,6 +123,7 @@ const structuredData = buildCollectionPageJsonLd({
   name: title,
   description,
   url: canonicalUrl,
+  items: buildTemplateItemListEntries(templates, locale),
```
Comment (dante01yoon) [inline site/src/pages/workflows/use-cases/[slug].astro]: question: Was "rich markup gated to the indexable pages only" meant to cover the ItemList too? Pages failing `isUseCasePageIndexable` (noindex, line 102) still ship the full ItemList here (same on the model page), unlike FAQ/HowTo/SoftwareApplication. If the gate was intended: `items: noindex ? undefined : buildTemplateItemListEntries(templates, locale)` — the builder provably falls back to the pre-PR bare CollectionPage. If the ItemList is deliberately ungated because it mirrors the grid that noindex pages do render, ignore this.

site/src/components/workflow-pages/SeoIndexPage.astro
```diff
@@ -19,9 +20,7 @@ import {
 
 interface Props {
   locale: Locale;
-  /** Root-relative index path (no locale prefix), e.g. "/workflows/model/". */
   basePath: string;
```
dante01yoon, line 25: nit(non-blocking): The deleted doc comments ("Root-relative index path (no locale prefix)…" and the `heading` one) documented a contract that's still load-bearing — `localizeUrl` (i18n/utils.ts:19) blindly prefixes, so a locale-prefixed `basePath` (e.g. `modelsIndexPath(locale)`; the helper accepts an optional locale) would double-prefix the canonical and hreflang URLs into `/ja/ja/…`. Restore the comments, or fold the invariant into the file header.

site/src/pages/workflows/model/[name].astro
```diff
@@ -234,6 +257,24 @@ const relatedCards = resolveRailImages(relatedCardData, group.templates, {
       />
     )
   }
+  {
+    howToStructuredData && (
```
Comment (dante01yoon) [inline site/src/pages/workflows/model/[name].astro]: nit(non-blocking): This file and use-cases/[slug].astro now each carry four copies of the identical `{x && <script is:inline type="application/ld+json" set:html={serializeJsonLdForScript(x)} />}` block (this PR adds the 3rd and 4th in each). A single `{[breadcrumbStructuredData, faqStructuredData, howToStructuredData, softwareAppStructuredData].filter(Boolean).map((data) => …)}` per page renders the same output and makes the next schema type a one-array-entry change. The per-block pattern predates this PR elsewhere, so also fine to leave for a site-wide cleanup.

site/src/lib/workflow-pages/index-cards.ts
```diff
@@ -3,13 +3,37 @@
...
 import type { CardBadge, SeoPageCard } from './schema';
 import { SEO_PAGES } from './use-cases';
 import { qualifyingGroups, resolveUseCasePageTemplates } from './use-case-resolver';
 
+/**
+ * Map a page's workflow grid to `ItemList` entries for its CollectionPage —
+ * each template linking to its own detail page, mirroring the visible grid.
+ * Templates with no derivable detail URL are skipped.
+ */
+export function buildTemplateItemListEntries(
```
dante01yoon: suggestion: This helper doesn't fit index-cards.ts — the module header says "Card resolvers for the model + use-case index pages", but the only consumers are the *detail* routes, the index pages build their entries inline in SeoIndexPage.astro, and the function drags in imports (`workflowDetailPath`, `absoluteUrl`, `ItemListEntry`) unused by the card resolvers. A small `workflow-pages/item-list.ts` would be a truer home (and could also host SeoIndexPage's inline mapping); at minimum, update the module header.

site/src/lib/structured-data.ts
```diff
@@ -95,18 +95,94 @@ export function buildFaqJsonLd(faqItems: FaqItem[] | undefined) {
...
+
 export function buildCollectionPageJsonLd(params: {
   name: string;
   description: string;
   url: string;
   inLanguage?: string;
+  /** Child pages to enumerate as a nested `ItemList`; omit for a bare CollectionPage. */
+  items?: ItemListEntry[];
 }) {
+  const itemList = params.items?.length
```
dante01yoon: nit(non-blocking): `itemList` actually holds a `{ mainEntity: … }` envelope, not an ItemList, and the wrapper diverges from this file's `...(x ? { x } : {})` idiom (lines 33, 125, 136, 186). `const mainEntity = params.items?.length ? { '@type': 'ItemList', … } : undefined` then `...(mainEntity ? { mainEntity } : {})` — same bytes out, and the name matches the content.

site/src/lib/routes.ts
```diff
@@ -27,6 +28,17 @@ export const thumbnailPath = (asset: string) =>
     ? asset
     : `/workflows/thumbnails/${asset}`;
 
+/**
+ * Resolve a thumbnail asset to an absolute URL for structured-data image fields.
+ * Hub-CDN URLs are already absolute; local assets are rooted then absolutized.
+ * Returns `undefined` for a missing thumbnail so callers can omit the field.
+ */
+export const resolveAbsoluteThumbnail = (thumbnail: string | null | undefined): string | undefined => {
```
dante01yoon: nit(non-blocking): This line is 103 chars — over the site's Prettier printWidth 100 — and `prettier --check` also fails on the two new test files (structured-data.test.ts lines 27/81/97/124, index-cards.test.ts line 34); verified with prettier 3.2.5 + site/.prettierrc. site-ci's `format:check` only runs on PRs targeting main, so it won't flag this stacked PR but will fail once the stack lands. `pnpm run format` in site/ fixes all three files.

dante01yoon (approved):

PR #995, opened 2026-07-07
feat(site): refresh hub OG/Twitter share card image and caption

MaanilVerma:
## Summary

Refreshes the social share card (OpenGraph + Twitter) for the ComfyUI Workflows hub: a new share image and a punchier tagline now render wherever a hub link is pasted — Slack, X, Discord, iMessage, LinkedIn. The new image uses a new filename so the major platforms fetch it fresh instead of serving a stale cached card.

## Changes

**What:**

- The hub's default share card now shows updated artwork and reads "Ready-to-run workflows. Pick one, hit run, see results in seconds. No building from scratch." across every entry point that shares a hub link.
- The share-card tagline is a dedicated field, separate from the page's meta description — so the marketing copy shows on social while the SEO/search snippet and structured data keep the text Google indexes.
- The tagline is translated across all 11 supported locales, so a shared link renders the native caption for each language.
- Category, model, and tag listing pages pick up the new share image while keeping their own page-specific descriptions (a "Flux workflows" card still describes Flux, not the generic hub tagline).
- Share cards now carry image alt text and an explicit image MIME type, which the Twitter/X validator flags when missing.
- The new image lives in a version-controlled assets folder rather than the synced-templates directory, so it ships as a normal committed asset with no ignore-rule exceptions.

**Breaking:** None. Per-page dynamic cards (individual workflow and creator previews) are untouched, and the search-result snippet + JSON-LD description are unchanged.

## Review Focus

- **The key decision** is separating the social caption from the page description. `meta.ogDescription` is a new i18n key consumed only by the share tags; `meta.description` still drives `<meta name="description">` and the structured-data description. If you'd rather the punchy caption also become the Google snippet, that's a one-line change — flag it.
- **Cache-busting via filename.** The image moved to a new path, so platforms refetch automatically without anyone re-scraping through debugger tools. The old default image and its now-dead ignore-rule exception are removed in the same change.
- **Scope boundary — hub only.** The Comfy homepage (comfy.org root) lives in a separate repo and is deliberately not touched here; its share card is a follow-up in that codebase.
- The one bit of logic worth a look is the MIME derivation in `SEOHead.astro`: it reads the extension off the image path *after* stripping any query string, so the dynamic `og.png?…thumbnail=x.jpg` endpoint isn't mislabeled as JPEG.

## Testing

Verification here is about what actually renders in the shared card, so the proof is the built HTML plus confirming each locale resolves its own caption — unit tests add little for static metadata.

- `astro check` (typecheck): 0 errors, 0 warnings.
- Production build: succeeded; the new image ships to the build output.
- Rendered the built hub landing `<head>` and confirmed `og:description` / `twitter:description` carry the new caption, `og:image` / `twitter:image` point at the new asset, and the alt/type/dimension tags are present and correct.
- Confirmed `t('meta.ogDescription', locale)` returns native text (not an English fallback) for a spread of locales including RTL Arabic.
- Manual check to run before merge: paste a hub URL into the X Card Validator / opengraph.xyz after deploy to confirm the fresh card renders.


dante01yoon (commented):
Clean OG/Twitter share-card refresh. Verified independently: no lingering references to the removed `/workflows/og-default.png` anywhere in `site/`, the new `brand/og-default-v2.png` is exactly 1200×630 (matches the hardcoded `og:image:width/height`) and lands in a non-ignored, committed path, the MIME derivation strips the query string before reading the extension (so a dynamic `og.png?…` endpoint isn't mislabeled), and `ogDescription` is intentionally scoped to the hub index pages while category/model/tag keep their page-specific descriptions. No issues found. (0 blocking, 0 non-blocking.)

dante01yoon (approved):

PR #1029, opened 2026-07-22
refactor(sync): refresh template usage on template PRs

MaanilVerma:
## Summary

The usage refresh now runs on template PRs and commits the updated numbers to the PR branch, instead of trying to write to main on a schedule.

## Changes

**What:**
- The scheduled version couldn't land its changes: main requires a pull request (repository ruleset), and Actions isn't permitted to open one, so both the direct push and the PR attempt failed in CI. It now triggers on PRs that touch template JSON and commits the refreshed usage onto that branch — the same shape as `sync-custom-nodes` and the version bump.

**Breaking:** None. The fetch and fold steps are unchanged.

## Review Focus

Uses the built-in token, so no new secrets. Worth knowing: usage now refreshes as a side effect of template PRs rather than on a clock — a standalone scheduled refresh would need the Cloud Code Bot credentials (`APP_ID` + `CLOUD_CODE_BOT_PRIVATE_KEY`), which this repo doesn't have.

## Testing

Ran the workflow's steps locally against the live index: fetch pulled 439 run_clicks, the fold reported 413/571 coverage, staging picked up only the 12 index files, and a second run produced no further changes. Two earlier CI runs confirmed the failure modes this replaces.

### DIFF
--- a/.github/workflows/refresh-run-click-usage.yml
+++ b/.github/workflows/refresh-run-click-usage.yml
@@ -2,37 +2,37 @@ name: Refresh Run-Click Usage
 
 # Fetches fresh per-template run_clicks from the Algolia templates_index and folds
 # them into templates/index*.json (the `usage` field the cloud/desktop pickers rank
-# on). Opens a PR rather than committing to main, so the existing version-check +
-# publish flow ships it with a human merge. Runs biweekly (1st & 15th) and on demand.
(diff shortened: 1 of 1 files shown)


MaanilVerma:
Superseded by a cleaner single-purpose PR.

PR #1034, opened 2026-07-22
feat(site): sync Hub navbar with upstream comfy.org

MaanilVerma:
## Summary

Brings the Hub navbar back in line with comfy.org, which had drifted since the fork was taken in June. The Products dropdown had been reduced to a single Docs link, the new-release badges were missing, and the Download/Launch buttons silently lost their hover reveal in Chinese.

## Changes

**What**

- The Products dropdown now lists Comfy MCP, Launches and Supported Models alongside Docs, and its featured card points at Comfy MCP instead of the old Seedance workflow.
- Products and Community carry a NEW badge, matching the marketing site.
- The desktop CTA buttons reveal their full label on wide screens again in Chinese. The old code found the split point by looking for a space, and Chinese has none, so the prefix came back empty and the reveal did nothing. Labels now come from separate translation keys.
- The badge pill is the right shape. This repo had overridden Tailwind's small radius to 6px, which read too round on the skewed pill, and no other step in the scale landed on the 4px the design wants. Dropping the override gets it back.
- Badge text sits on its optical centre. PP Formula rides high in its box, so without a nudge the word "NEW" floated toward the top of the pill.
- On mobile the badge sits next to its label instead of drifting into the middle of the row, and the chevron stays flush right.
- The mobile menu's screen reader description no longer just repeats the title.

https://github.com/user-attachments/assets/8ba365f8-c5b8-4b16-9dc3-6191eb376f34


site/src/components/hub/nav/HeaderMain.vue
```diff
@@ -19,22 +19,20 @@ const {
...
-}
+/** Chinese sets solid, so a Latin gap reads as a typo. Korean and Japanese keep theirs. */
+const ctaGap = locale === 'zh' || locale === 'zh-TW' ? '' : '\u00a0';
 
 const ctaButtons = [
   {
-    ...splitLabel(t('nav.downloadLocal', locale)),
+    prefix: t('nav.ctaDesktopPrefix', locale),
+    core: t('nav.ctaDesktopCore', locale),
     ariaLabel: t('nav.downloadLocal', locale),
```
dante01yoon, line 26: issue: The visible CTA now uses the newly localized `ctaDesktopPrefix`/`ctaDesktopCore`, but `ariaLabel` still comes from `nav.downloadLocal`, which remains English in Japanese, Korean, Spanish, French, Russian, Turkish, Arabic, and Portuguese. Since `aria-label` overrides the descendant text, screen-reader users hear `DOWNLOAD DESKTOP` while sighted users see the localized label (and the cloud CTA has the same problem on line 36). Could this accessible label be constructed from the localized split values, or could the legacy full-label keys be translated consistently?

site/src/styles/global.css
```diff
@@ -14,7 +14,6 @@
   --font-display: 'PP Formula', 'Inter', system-ui, sans-serif;
   --font-body: 'Inter', system-ui, -apple-system, sans-serif;
 
-  --radius-sm: calc(var(--radius) - 4px);
```
dante01yoon, line 17: suggestion: Removing the global `--radius-sm` override changes every `rounded-sm` consumer, not only the new badge pill—for example `ExpandableText.vue` also moves from 6px to Tailwind's 4px default. Unless that broader visual change is intentional, could the 4px radius be scoped to the accent badge instead?

dante01yoon (changes requested):
The navbar sync is structurally sound and the focused checks pass, but the localized CTA accessible names need to match the newly localized visible labels before merge. One non-blocking styling scope concern is noted inline as well.

site/src/components/hub/nav/HeaderMain.vue
MaanilVerma: Fixed.

site/src/styles/global.css
MaanilVerma: Yes, this is intentional.

dante01yoon (approved):
The follow-up commit resolves the localized CTA accessible-name mismatch: each affected locale now supplies localized full labels matching the visible prefix/core text. The broader radius change was confirmed intentional.

Validation: 303 unit tests pass; `astro check` reports 0 errors; focused TypeScript ESLint checks pass. The preview job failure is unrelated to this diff (`Hub API returned 525`); the site build, lint, and SEO checks are green.

PR #940, opened 2026-06-15
fix: implement hard 404 responses for invalid .json/ URL paths and empty taxonomy pages

balpreetgrowthnatives:
## Summary
This PR addresses soft 404 flags by ensuring that the Astro site's server-side rendered (SSR) dynamic pages return a hard 404 HTTP status code (instead of rendering a blank UI shell with `200 OK` or triggering a `302` redirect) when a request is made with invalid data, an empty taxonomy, or an appended trailing slash.

## Changes Made
- **Early Truncation Check**: Added a check for `Astro.url.pathname.endsWith('.json/')` at the top of the dynamic pages' frontmatter to return a hard 404 `Response` immediately, blocking bad trailing slash Vercel redirects from matching dynamic layouts.
- **Workflow Detail Pages**: Updated dynamic pages to return a hard 404 response on missing workflow detail/data or database fetch errors.
- **Taxonomy Pages**: Changed empty/invalid checks for categories, models, and tags on localized dynamic routes to return a hard 404 response instead of redirecting or returning a 200 OK blank layout.
  - Modified files:
    - [\[username\].astro](file:///home/nishant/growthnativesworkspace/comfy/workflow_templates/site/src/pages/workflows/[username].astro)
    - [\[slug\].astro](file:///home/nishant/growthnativesworkspace/comfy/workflow_templates/site/src/pages/[locale]/workflows/[slug].astro)
    - [\[type\].astro](file:///home/nishant/growthnativesworkspace/comfy/workflow_templates/site/src/pages/[locale]/workflows/category/[type].astro)
    - [\[name\].astro](file:///home/nishant/growthnativesworkspace/comfy/workflow_templates/site/src/pages/[locale]/workflows/model/[name].astro)
    - [\[tag\].astro](file:///home/nishant/growthnativesworkspace/comfy/workflow_templates/site/src/pages/[locale]/workflows/tag/[tag].astro)

## Verification
- Verified by running `SKIP_AI_GENERATION=true npm run build` successfully on the local environment.

PR #1044, opened 2026-07-27
feat(site): hub localization pipeline + page-payload fix (GTM-291)

mobeenabdullah:
Part 1 of GTM-291 (hub localization), per the approved design doc. This PR is **code + tooling only** — it renders nothing new to users and flips no locale to indexable. The translations themselves are generated + committed by the CI workflow it adds (push English, CI translates).

## What's here

**Page-payload fix** (was going to be a separate PR; folded in):
- The search popover and "view all workflows" grid each embedded the entire ~580-workflow catalog into every page's HTML (~1.3 MB, 83% of a detail page). Both now use a lazily-fetched `/workflows/grid.json`; the popover gets small precomputed discovery slices. **Sample detail page 1,571 KB → 334 KB (-79%); whole-site HTML ~1.03 GB → ~214 MB.** Behavior verified headless (discovery, badge filtering, grid pagination, self-exclusion).

**Translation pipeline** (`src/lib/i18n/` + `scripts/i18n/`):
- Portable, fs-based, `astro:*`-free modules (like `landing-content.ts`) so routes, sitemap, hreflang, and CI share one decision — and the future hub migration relocates files, not logic.
- `resolveLocalizedWorkflow(shareId, locale) → {data, indexable, reason}`: precedence reviewer-override > human-seed > machine > English; never a silent English blend.
- `isLocalePageIndexable`: the one per-page predicate (translated ∧ current-source-hash ∧ reviewed-at-this-hash ∧ supported ∧ flipped) that will drive sitemap/hreflang/canonical/build-assert in Part 2.
- Three locale sets (`AVAILABLE_APP ⊇ SUPPORTED_HUB ⊇ INDEXABLE`), `INDEXABLE` empty.
- Builder: Hub index → English content-of-record + per-field source-hash manifest + seeds the existing human translations (4,100 across 10 locales). Staleness: a changed English hash drops the entry so lobe re-translates and any sign-off auto-goes-stale.
- Glossary: proper-noun preserve list (kept English) + app-terminology mirror (rebuilt in CI from ComfyUI_frontend locales; gitignored).
- Deterministic validators (structure, language/leakage, glossary, URLs, brand-voice) — CI-blocking.
- Working `.i18nrc.cjs` (replacing the abandoned one whose entry never existed) + `i18n:check` guard.

**UI strings → JSON** (`ui.ts` → `locales/*.json`): byte-accurate; `t()` unchanged; unblocks the 39 UI strings that are English-only in every locale today. `.i18nrc.ui.cjs` fills them.

**CI** (`.github/workflows/i18n-update-hub.yml`): regenerates English, seeds human translations, machine-translates with the org `OPENAI_API_KEY`, validates, and opens a translation PR. Never flips a locale.

## Notes for review
- Tests: 33 new i18n unit tests; typecheck 0 errors.
- `pnpm lint` fails only on the pre-existing `BaseLayout.astro` `prefer-rest-params` error (present on main, fixed by #940) — not introduced here.
- Part 2 (localized rendering + prerender + the predicate wired to sitemap/hreflang/canonical, all gated off) is a separate PR that rebases on #940 and preserves its 404/502 semantics.

Refs GTM-291.


mobeenabdullah:
Thanks @coderabbitai — went through each finding:

- **grid.json.ts (200 [] on index failure):** fixed in 22187fe — it now throws (fail-closed) so a missing Hub index blocks the build instead of shipping/caching an empty browse grid.
- **SearchPopover — re-run on catalog load:** fixed — `catalogLoaded` is now a dependency of the debounced search, so a badge-scoped query typed before the lazy catalog lands re-runs with the correct `allowedNames` once it loads (badge-only results already recompute reactively).
- **SearchPopover — retry on load failure:** fixed — a failed catalog load now retries when the popover is reopened with active filters (in addition to the existing retry-on-next-badge), so a transient fetch failure recovers on interaction.
- **.i18nrc.ui.cjs lint:** resolved earlier — added it to the eslint `ignores` list, matching its sibling `.i18nrc.cjs` (both are intentional CommonJS configs).
- **i18n-predicate.test.ts FieldSource:** resolved earlier in 8f3275f — the fixture now uses `'machine'` (a valid `FieldSource`), which also cleared the Astro Check ts2352.
- **Locale JSON untranslated strings:** intentional — those are the known UI-string gap that the CI `locale:ui` step fills each run; they're committed as English fallbacks, not a defect.

mobeenabdullah:
@coderabbitai resolved the remaining threads in 059b030:

- **Untranslated UI strings (10 locale files):** you're right — the CI `locale:ui` step only fills *missing* keys, so present-but-English values were being skipped permanently. I removed the English-identical values from each locale JSON. They still render identically (`t()` falls back to `en`), but they're now missing keys, so the translation step will actually translate them going forward.
- **.i18nrc*.cjs lint:** switched from ignoring them to a proper CommonJS ESLint override (node globals + `no-require-imports` off), so the configs are linted for real issues instead of skipped.

.github/workflows/i18n-update-hub.yml
```diff
@@ -0,0 +1,119 @@
...
+          git config user.name 'github-actions[bot]'
+          git config user.email 'github-actions[bot]@users.noreply.github.com'
+          # Stage first — the generated content/ files are new/untracked on the
+          # initial run, which `git diff` alone would miss.
+          git add site/src/i18n
+          if git diff --cached --quiet; then
+            echo "No translation changes."
+            exit 0
+          fi
+          BRANCH="i18n/hub-translations-$(date -u +%Y%m%d-%H%M%S)"
```
dante01yoon: **The partial-progress run is not actually resumable.** Every invocation checks out the default branch, then saves progress to a new timestamped branch. If a rate-limited run's PR has not merged before the next cron/manual run, that next run cannot see the saved locale files, retranslates the same keys, and opens another PR. Concurrency only serializes runs; it does not carry state between branches. Please reuse one stable automation branch/open PR (rebased or merged with `main`) so each run starts from the last committed partial output.

.github/workflows/i18n-update-hub.yml
```diff
@@ -0,0 +1,119 @@
...
+      - name: Format
+        if: always()
+        working-directory: site
+        run: pnpm exec prettier --write src/i18n
+
+      # Open a PR only when the translation artifacts actually changed. The
+      # committed en.json/manifest anchor the per-page freshness + sign-off gate,
+      # so this is the single place translations enter the repo.
+      - name: Open translation PR
+        if: always()
```
dante01yoon: **A failed quality gate still publishes the invalid artifacts.** `if: always()` lets this step commit/push/open a PR even when `pnpm i18n:validate` or Prettier failed. The normal Site CI workflow does not run `i18n:validate`, so that generated PR can then show green checks despite failing this workflow's deterministic gate. Please gate publication on successful validation/formatting (while still allowing the translation step itself to fail after saving partial, valid output), and/or add `i18n:validate` to PR CI for these artifacts.

site/src/lib/i18n/predicate.ts
```diff
@@ -0,0 +1,57 @@
...
+  );
+  if (untranslated.length > 0) {
+    return { indexable: false, reason: `untranslated fields: ${untranslated.join(', ')}` };
+  }
+
+  // 3. A native reviewer must have signed off on this exact English version.
+  if (!input.review) {
+    return { indexable: false, reason: 'no review sign-off' };
+  }
+  if (input.review.reviewedContentHash !== input.currentContentHash) {
```
dante01yoon, line 56: **Native approval is only bound to the English source, not to the localized artifact that was reviewed.** `ReviewRecord` carries `reviewedArtifactChecksum`, but the resolver/predicate never computes or compares the current resolved locale checksum. After a locale is flipped, changing machine/human/override text without changing English leaves this predicate indexable under the old sign-off. Please deterministically hash the final resolved localized data, pass that checksum into the predicate, and require it to match `reviewedArtifactChecksum` (with a regression test).

site/scripts/i18n/build-content-source.ts
```diff
@@ -0,0 +1,284 @@
...
+ * (older manifest) every field reads as changed, i.e. the whole entry re-translates.
+ */
+export function staleFields(
+  prev: TranslationManifest,
+  next: TranslationManifest
+): Record<string, TranslatableField[]> {
+  const result: Record<string, TranslatableField[]> = {};
+  for (const [shareId, nextHashes] of Object.entries(next)) {
+    const prevHashes = prev[shareId];
+    if (prevHashes == null) continue;
```
dante01yoon, line 103: **A removed then re-approved workflow can silently reuse an obsolete machine translation.** IDs absent from `next` are never marked stale, and `pruneStaleFields` preserves every machine entry absent from the stale map. If an approved workflow disappears, then later returns with changed English, its prior manifest entry is gone, this branch treats it as brand-new, and lobe sees the old locale key as already translated. Please prune machine IDs not present in the current manifest on every build, and cover remove → re-add with a test.

site/scripts/i18n/validate-translations.ts
```diff
@@ -0,0 +1,222 @@
...
+    return fallback;
+  }
+}
+
+function main(): void {
+  const english = readJson<Record<string, WorkflowContent>>(path.join(CONTENT_DIR, 'en.json'), {});
+  const preserveTerms = readJson<string[]>(path.join(GLOSSARY_DIR, 'preserve-terms.json'), []);
+
+  const all: Violation[] = [];
+  for (const locale of SUPPORTED_HUB_LOCALES) {
```
dante01yoon, line 318: **The UI translations generated by `locale:ui` bypass deterministic validation.** This loop only checks `src/i18n/content`; it never checks `src/i18n/locales/*.json`. Several generated UI strings contain required placeholders such as `{label}`, `{count}`, and `{keyword}`, so a model can drop/rename one and still be committed, breaking runtime interpolation. Please validate UI locale value types and exact placeholder multisets against `locales/en.json` before publication (partial files can still allow missing keys).

site/.i18nrc.cjs
```diff
@@ -1,24 +1,39 @@
...
+// (titles, descriptions, meta, how-to, FAQs) from the English content-of-record
+// into each supported locale's machine layer. Run: `pnpm locale`.
+//
+// Kept in CommonJS to match the tooling. The reference prompt is composed from
+// the committed do-not-translate list so glossary + prompt never drift.
 const { defineConfig } = require('@lobehub/i18n-cli');
 
+let preserveTerms = [];
+try {
+  preserveTerms = require('./i18n/glossary/preserve-terms.json');
```
dante01yoon, line 31: **The generated app-terminology mirror is never consumed.** The workflow checks out `ComfyUI_frontend` and `sync-glossary` writes `mirror/{locale}.json`, but this config loads only `preserve-terms.json`; repository-wide there is no reader for the mirror or the locale overrides. Therefore the advertised product-terminology consistency mechanism currently has no effect. Please inject the relevant locale mirror/overrides into each translation prompt (which may require per-locale invocation), or enforce them in validation; otherwise the extra checkout/sync artifact should not be presented as active behavior.

dante01yoon (changes requested):
The payload reduction looks sound, and the latest commits address the earlier catalog-loading and locale-file feedback. I found several remaining correctness gaps in the translation pipeline that need resolving before it can safely generate resumable, review-gated translation PRs. The current preview failure is the same external Hub API 525 seen on #1048, not one of these findings.

.github/workflows/i18n-update-hub.yml
mobeenabdullah: Fixed in 4521201. The workflow now uses one stable branch `i18n/hub-translations`: each run fetches and restores its last committed `site/src/i18n` before translating (so lobe skips already-done keys), then commits back to the same branch with `--force-with-lease` and reuses the open PR. A rate-limited run's partial output is carried into the next run instead of being orphaned on a per-run timestamped branch.

.github/workflows/i18n-update-hub.yml
mobeenabdullah: Fixed in 4521201. `Validate` stays unconditional (`if: always()`), but `Format` and the PR-open step now run only on `success()`, so a validation failure fails the job and skips publication — no invalid artifacts committed, no green-looking PR. A tolerated translate rate-limit keeps the job successful (`continue-on-error`), so valid partial output still publishes. Also added `i18n-validate.yml` so the deterministic validator runs on any PR touching `site/src/i18n/**` or the pipeline scripts/config, closing the gap where normal Site CI skips it.

site/src/lib/i18n/predicate.ts
mobeenabdullah: Fixed in a241ec4. Added `hashResolvedArtifact(data)` (canonical TRANSLATABLE_FIELDS order) in the resolver; the predicate now requires `review.reviewedArtifactChecksum === currentArtifactChecksum` on top of the content-hash term. So changing machine/human/override text after a flip, without touching English, de-indexes the page until re-review. Added a predicate unit test and a resolver regression test (English + content hash unchanged, artifact bytes differ → non-indexable).

site/scripts/i18n/build-content-source.ts
mobeenabdullah: Fixed in 94ebefc. Added `pruneOrphanIds(machine, currentShareIds)`, applied on every build (not gated on stale fields): machine entries whose shareId is absent from the current English source are dropped. So a removed workflow's machine translation no longer lingers, and a later re-add with changed English re-translates fresh instead of reusing the obsolete key. Added a remove → re-add regression test.

site/scripts/i18n/validate-translations.ts
mobeenabdullah: Fixed in 686129c. Added `collectUiViolations` and wired it into `main()`: every present key in `locales/{locale}.json` is checked for value-type parity and an exact `{placeholder}` multiset match against `locales/en.json` (missing keys stay allowed via the English fallback). A dropped or renamed `{label}`/`{count}`/`{keyword}` now fails validation. Added tests.

site/.i18nrc.cjs
mobeenabdullah: Fixed in cef0200 (config) + 4521201 (CI) + 686129c (validator). The config now has a single-locale mode (`HUB_I18N_LOCALE`) that injects that locale's `mirror/{locale}.json` (capped, longest terms first) plus `overrides/{locale}.json` (which win) into the reference prompt — impossible in the all-locales run since the reference is shared across every output locale. CI now translates one locale at a time so the injection applies (and it paces under the OpenAI tier). Separately, the validator now enforces overrides as binding: a curated term the translation fails to render is a violation. So the mirror feeds the prompt and overrides both guide and gate.

mobeenabdullah:
Thanks for the thorough pass @dante01yoon — all six are addressed in `059b030..4521201`, with replies on each thread:

- **Artifact-bound sign-off:** the predicate now hashes the resolved localized artifact and requires it to equal `reviewedArtifactChecksum`, so post-flip edits to machine/human/override text de-index until re-review (predicate + resolver tests added).
- **Orphan pruning:** machine entries absent from the current English source are pruned every build, so a removed → re-added workflow re-translates fresh (remove/re-add test added).
- **UI validation:** `i18n:validate` now checks UI-string types and exact placeholder multisets against `en.json`, and a new `i18n-validate.yml` runs it on any PR touching the artifacts.
- **Publication gating:** `Format`/PR-open run only on `success()`, so a failed validate never publishes; a tolerated translate rate-limit still commits valid partial output.
- **Resumable runs:** one stable automation branch is restored-from and committed-back-to each run, reusing its open PR.
- **Glossary consumption:** translation is now per-locale (`HUB_I18N_LOCALE`), injecting each locale's mirror + overrides into the prompt, and the validator enforces overrides as binding.

Full unit suite + `astro check` green locally; the `preview` red is the same external Hub API 525. Re-requesting your review.

site/scripts/i18n/validate-translations.ts
```diff
@@ -0,0 +1,333 @@
...
+
+  for (const field of TRANSLATABLE_FIELDS) {
+    if (!(field in localized) || localized[field] == null) continue;
+    const enValue = english[field];
+    const locValue = localized[field];
+    const enText = toText(enValue);
+    const locText = toText(locValue);
+
+    // 1. Structure: arrays must match English length; FAQ items well-formed.
+    if (Array.isArray(enValue) && Array.isArray(locValue)) {
```
dante01yoon: **Wrong localized container types still bypass the deterministic gate.** This block only validates structure when both values are arrays. If the English `howToUse`, `suggestedUseCases`, or `faqItems` array is emitted as a scalar/object, the block is skipped; `collectViolations(..., { howToUse: '错误标量' }, ...)` returns no violations, and the resolver can later treat that malformed value as translated content. Please reject array/scalar parity mismatches before length/item checks, validate element shapes/types, and add a scalar-for-array regression test.

.github/workflows/i18n-update-hub.yml
```diff
@@ -0,0 +1,161 @@
...
+          GH_TOKEN: ${{ secrets.PAT_TOKEN }}
+        run: |
+          set -euo pipefail
+          STABLE_BRANCH="i18n/hub-translations"
+          git config user.name 'github-actions[bot]'
+          git config user.email 'github-actions[bot]@users.noreply.github.com'
+          # Stage first — the generated content/ files are new/untracked on the
+          # initial run, which `git diff` alone would miss.
+          git add site/src/i18n
+          if git diff --cached --quiet; then
```
dante01yoon, line 142: **A no-op retry still rewrites the open translation PR head.** Restoring the stable branch makes the staged tree differ from `main` for as long as its PR is open, so this check is non-empty even when the translation run produced byte-for-byte identical artifacts. The workflow then creates an equivalent new commit and force-pushes it every day, rerunning CI and obsoleting/dismissing review state without progress. Compare generated `site/src/i18n` against the fetched stable branch before committing (while separately handling a real base update if desired), so unchanged retries exit without changing the PR SHA.

dante01yoon (changes requested):
The six issues from my prior review are addressed, and the focused i18n tests pass. Two correctness/operational gaps remain: malformed localized container types still bypass the deterministic gate, and no-op scheduled runs rewrite the stable translation PR head. The preview failure remains the external Hub API 525, not a regression in this PR.

site/scripts/i18n/validate-translations.ts
```diff
@@ -0,0 +1,364 @@
...
```
dante01yoon: This still only enters structure validation when both values are arrays. An English array field localized as a scalar/object therefore bypasses the gate. Please enforce array/container parity first and add scalar/object-for-array regression coverage.

.github/workflows/i18n-update-hub.yml
```diff
@@ -0,0 +1,161 @@
...
+        env:
+          GH_TOKEN: ${{ secrets.PAT_TOKEN }}
+        run: |
+          set -euo pipefail
+          STABLE_BRANCH="i18n/hub-translations"
+          git config user.name 'github-actions[bot]'
+          git config user.email 'github-actions[bot]@users.noreply.github.com'
+          # Stage first — the generated content/ files are new/untracked on the
+          # initial run, which `git diff` alone would miss.
+          git add site/src/i18n
```
dante01yoon, line 141: The stable branch is restored into a checkout of `main`, then this stages and compares only against `main`. While the automation PR is open, a no-op retry still sees the existing translation diff, creates an equivalent commit, and force-pushes a new head. Compare generated artifacts with the fetched stable branch state before rebuilding/pushing so no-progress runs leave the PR SHA unchanged.

dante01yoon (changes requested):
Re-reviewed the new head. The latest commit correctly catches non-empty locale files with zero IDs in common with English, but it does not address either blocker from the previous review: wrong container types still bypass deterministic validation, and scheduled no-op runs still rewrite the stable translation PR head. This is not ready to approve yet.

site/scripts/i18n/validate-translations.ts
mobeenabdullah: Fixed in 7678412. The structure check now enforces container parity first: an array field emitted as a scalar/object (or a string field as an array) is a `structure` violation before any length/element check. Array elements are type-checked too — FAQ items must be `{question, answer}` non-empty strings, and how-to/use-case elements must be strings. Added scalar-for-array, array-for-string, and non-string-element regression tests.

.github/workflows/i18n-update-hub.yml
mobeenabdullah: Fixed in 7678412. Before committing, the run now compares the generated `site/src/i18n` against the fetched stable branch (`refs/remotes/origin/i18n/hub-translations`) rather than `main`, and exits without pushing when they match. So a byte-identical daily rerun leaves the PR head and its review state untouched; only a real content change force-pushes. (The restore step fetches into that remote-tracking ref so the comparison — and the later force-with-lease — has a real base.)

mobeenabdullah:
@dante01yoon both remaining gaps are fixed in 7678412 (threads resolved):

- **Container parity**: `collectViolations` now rejects an array field emitted as a scalar/object (and a string field as an array) before any length/element check, and type-checks array elements (FAQ item shapes, string elements) — with scalar-for-array / array-for-string / non-string-element regression tests.
- **No-op force-push**: before committing, the run compares the generated `site/src/i18n` against the fetched stable branch (not `main`) and exits without pushing when unchanged, so a byte-identical daily rerun leaves the PR head and its review state intact; only a real change force-pushes.

Full i18n unit suite + `astro check` green; the `preview` red is the external Hub API 525. Re-requesting your review.

mobeenabdullah:
@dante01yoon kindly review this PR when you get a chance

dante01yoon (approved):

PR #1048, opened 2026-07-28
fix: hard 404 for invalid .json/ paths and empty taxonomy pages (re-host of #940)

mobeenabdullah:
Re-host of #940 by @balpreetgrowthnatives so CI can run with full permissions.

**Why this exists:** #940 is from a fork (read-only access), so its CI gets a read-only token and no repo secrets. That makes the "Comment on PR" step 403 (Lint & Format, SEO Audit) and the Vercel preview fail on missing secrets — even though ESLint/Prettier pass. Those are GitHub fork-PR limitations, not code issues, and can't be fixed from the fork side. This PR carries the **identical commits** (Balpreet retained as author) on an org branch so the same checks run with the permissions they need.

- Original **#940 stays open** (not superseded by content, only by CI-runnability).
- Credit: **@balpreetgrowthnatives** (Balpreet Brar).
- Follow-up (separate): fix the CI so fork PRs don't fail on the comment step, or grant write access, so this workaround isn't needed next time.


site/src/layouts/BaseLayout.astro
```diff
@@ -63,7 +63,7 @@ const bodyClasses = isDark
         <script async src={`https://www.googletagmanager.com/gtag/js?id=${GA_ID}`}></script>
         <script is:inline define:vars={{ GA_ID }}>
           window.dataLayer = window.dataLayer || [];
-          function gtag(){dataLayer.push(arguments);}
+          function gtag(...args){dataLayer.push(args);}
```
dante01yoon: issue: This changes Google’s supported gtag queue message from an `Arguments` object to an `Array`. gtag.js distinguishes those shapes, so the initial `js` / `config` commands (and later events) can be ignored or misinterpreted. Google’s documented shim specifically uses `dataLayer.push(arguments)`: https://developers.google.com/tag-platform/gtagjs. Could we restore the official implementation and narrowly suppress the lint rule instead of changing the queued value’s type?

site/pnpm-lock.yaml
```diff
@@ -10,34 +10,34 @@ importers:
     dependencies:
       '@astrojs/sitemap':
         specifier: ^3.3.1
-        version: 3.7.0
+        version: 3.7.3
       '@astrojs/vercel':
```
dante01yoon, line 14: issue: This lockfile refresh is unrelated to the hard-404 fix and silently changes the installed production/tooling graph without any `package.json` change (for example Astro 5.17.1 → 5.18.2, Tailwind 4.1.18 → 4.3.3, PostHog 1.360.0 → 1.407.2, and a Vite 6 → 8 resolution). That makes this route fix also ship a broad, unreviewed dependency upgrade. Could we restore `site/pnpm-lock.yaml` from `main` and handle dependency updates separately if they are intentional?

dante01yoon (changes requested):
The hard-404 and 502/no-store route changes look correct, including both CodeRabbit follow-ups. I found two unrelated blocking changes in the re-hosted commit that should be removed or corrected before merge. The failing `preview` check is not a code finding here: its log confirms `build-search-index` failed because the Hub API returned 525.

mobeenabdullah:
Thanks @dante01yoon, both addressed in 238da36:

1. **BaseLayout gtag** — restored Google's official `dataLayer.push(arguments)` (kept as the `Arguments` object, not an array, so the `js`/`config` commands aren't misinterpreted) and narrowly suppressed `prefer-rest-params` on just that line instead of changing the queued value's type. ESLint passes.

2. **pnpm-lock.yaml** — restored from `main`. Confirmed `site/package.json` is unchanged vs `main`, so the lockfile now matches it exactly and the unrelated Astro/Tailwind/PostHog/Vite bumps are gone. Any intentional dependency updates can go in a separate, reviewed PR.

Re-requesting your review when you have a moment.

dante01yoon (approved):
Re-reviewed the follow-up commit. Both blocking findings are resolved: the official `Arguments`-based gtag shim is restored with a narrow lint suppression, and `site/pnpm-lock.yaml` now matches `main`. The route handling remains correct and the code checks pass. The remaining `preview` failure is still the same external Hub API 525 during `build-search-index`, not a regression from this PR.

PR #1057, opened 2026-07-29
fix(site): link restore-old-photos to its workflow and keep use-case pins on top

mobeenabdullah:
## What

Fixes the remaining wrong-linking issues on the /workflows/use-cases pages (GTM-305 / follow-up to #1016 and #1052) and repairs the pin mechanism those PRs rely on.

- **restore-old-photos**: the hero and closing CTAs now open the published **Restore Old Photos** workflow (`69850664cf89`, sheet: Keywords v3, App Mode verified on production cloud). The grid is fully curated via pins to the seven restoration-relevant workflows (Restore Old Photos, Seedream 5.0 Pro Image Edit, Topaz Image Enhance, Image Relight, Topaz Illustration Upscale, Magnific Creative Image Upscale, WaveSpeed Image Upscale) instead of ~90 generic Image Edit matches. Previously the hero opened Seedream 5.0 Pro Image Edit, one of the pages Robin flagged as linking to the wrong workflow.
- **WorkflowGrid**: new optional `pinnedShareIds` prop keeps a page's curated pins above the usage/date sort on every tab. Without this, pins with no click history sort to the bottom and fall below the 30-card render cutoff. This is visible on production today: ai-image-to-video's two free-anchor pins from #1016 are absent from its rendered grid.
- **New published workflows pinned**: Image to Video (`3515c5083027`) on ai-image-to-video, Anime Generator (`9f0b568bf8a1`) on ai-anime-generator, Video Upscale (`0740bf78b7b6`, untagged so unreachable by filters) on ai-image-upscaler.
- **ai-tattoo-generator**: pin flagged `isApp` (App Mode verified) so its Comfy Apps tab lists the Tattoo Generator instead of being empty, and the Canny-tagged non-tattoo "AI on the Lot" video workflow is excluded.

## Verification

- 370/370 unit tests, eslint, prettier, `astro check` clean.
- Playwright walkthrough against a local dev build on the production Hub catalog, 23/23 checks: hero CTA share ids per page, pin order (including tab switches to Node Graphs / Comfy Apps), exclusions, and no regressions on headshot/caricature/upscaler/anime/character-replacement.

## Notes for review

- The three new pins are deliberately **not** flagged `isApp` yet; App Mode verification for them is pending (same convention as #1052).
- Avatar Generator and Character Replacement are untouched in this PR; their cleanup is tracked separately in GTM-305.

## Preview environment (resolved)

Previously the Vercel preview built against the test Hub (`HUB_API_URL_PREVIEW` / testcloud), which does not carry these production shareIds, so it silently dropped the new pins. That is now fixed in #1068, which points the preview build at the production Hub (approved-only), so the preview reflects production.

Working preview (all 15 use-case pages, including the fully curated image-to-3d): https://workflow-templates-ngwbeag3a-comfyui.vercel.app/workflows/use-cases

Infra note: because the preview now builds the full production catalog, the Vercel upload is large and can occasionally hit the 24h upload quota; re-running the preview job clears it.

## Second commit: interior-design x tattoo grid split

The shared ControlNet filter made the two pages render near-identical grids (the duplication nav flagged in Slack). Complementary `excludeShareIds` now split the 20 ControlNet/Canny catalog workflows into two disjoint clusters: **interior** keeps depth / inpainting / image-edit (10 items, geometry-preserving room restyling), **tattoo** keeps Canny / line-art / union-control (10 items behind the pinned Tattoo Generator). Zero shared items remain, verified via Playwright clicking through both pages (25/25 checks on the full suite).

## Commits 3-5: team-review curation pass + video visuals + sitemap fix

The whole 10-page set was re-reviewed against the final App Mode sheet (30 July) and manually QA'd page by page:

- **Grid curation** (`544e2631`): caricature, interior, and avatar become fully curated pages (the broad tags matched glitch tools, generic ControlNet demos, and brand-safety-gated face-swap/voice-clone items respectively). Avatar's dedicated workflow + CTA is now **Kling: Avatar 2.0** (`e81f8eb0ee5f`). Anime gains Anima Base v1 + Illustration LoRA pins. Tattoo/upscaler/image-to-video/character-replacement each drop a handful of off-topic matches (incl. two mis-tagged text-to-video workflows).
- **Video visuals** (`4dd09da5`): Hub app thumbnails are often video-only, so the hero and index cards fell back to unrelated (sometimes 404ing) stills. The lead pinned app now fronts its own page and index card as a muted autoplay loop; JSON-LD stops declaring videos as images. This is what finally makes the tattoo/caricature/avatar pages *look* fixed, not just link correctly.
- **Sitemap** (`323278f5`): curated (filter-less) pages stay in the sitemap; the on-disk snapshot the sitemap reads carries no shareIds, so pin-only pages previously looked empty to it.

Verified: 374/374 unit tests, astro check clean, and a 42-check Playwright pass over all 10 pages (hero CTAs, exact curated grids, tab contents, exclusions, hero/card videos).


site/src/lib/workflow-pages/use-cases.ts
```diff
@@ -178,9 +214,17 @@ export const SEO_PAGES: SeoPageDef[] = [
...
+    // Fully curated grid: the broad character tags pulled in face-swap and
+    // voice-clone workflows (brand-safety gated) plus generic reference-to-video
+    // items, so the page pins the avatar / talking-head cluster explicitly and
+    // matches nothing else.
+    filters: {},
+    // Kling: Avatar 2.0 — the dedicated avatar workflow.
+    // isApp left unset: App Mode not verified in cloud.comfy.org.
+    appShareId: 'e81f8eb0ee5f',
+    pins: [
+      { shareId: 'e81f8eb0ee5f' }, // Kling: Avatar 2.0
```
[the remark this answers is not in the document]
mobeenabdullah: Verified against cloud.comfy.org: e81f8eb0ee5f (Kling: Avatar 2.0) runs as a canvas workflow through ComfyUI API nodes, not App Mode. So rather than set isApp: true, I removed appShareId in 36452cf. The workflow stays pinned to lead the grid, so the hero and closing CTAs still resolve to it via the grid-pick fallback (leadShareId = def.appShareId ?? first pinned share), and heroHrefOverride now correctly falls back to the workflow rather than advertising an app. Good catch.

site/src/lib/workflow-pages/use-cases.ts
```diff
@@ -271,5 +344,142 @@ export const SEO_PAGES: SeoPageDef[] = [
...
+        'lyrics to song',
+        'ai songwriting',
+        'make a song with ai',
+      ],
+    },
+    // Fully curated: the song cluster (vocals + lyrics) stays disjoint from the
+    // instrumental ai-music-generator page.
+    filters: {},
+    // Song Generator (Ace-Step 1.5XL). App Mode not verified in cloud yet.
+    appShareId: '3ef4de40106b',
```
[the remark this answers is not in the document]
mobeenabdullah: Good catch, and consistent with the avatar fix. Removed appShareId from all five entries (Song Generator, Music Generator, Image Enhancer, Hairstyle Changer, Image to 3D). In each case appShareId equalled the first pin, and pins lead the grid, so leadShareId still resolves to the same workflow via the grid-pick fallback and the CTA target is unchanged; it just no longer advertises an unverified App Mode route. Pins/grids left intact. Fixed in the latest push.

site/src/components/workflow-pages/LandingHero.astro
```diff
@@ -105,20 +124,32 @@ const heroVisualProps = heroHref
...
-            <img
-              src={heroSrc}
-              alt={title}
-              class="size-full object-cover transition-transform duration-500 ease-out group-hover:scale-105"
-              loading="eager"
-              fetchpriority="high"
-              decoding="async"
-            />
+            {heroVideoSrc ? (
+              <video
```
MaanilVerma, line 135: issue: this `<video>` carries no `poster`, no `preload` and no error handling, replacing an `` that had `loading="eager"` and `fetchpriority="high"` — and `<video>` is not LCP-priority-eligible, so the hero regresses on every video-led page. `ThumbnailDisplay.astro:325` already models the maintained pattern, including a 3s `video.paused && readyState < 3` check precisely because autoplay suppression never fires `error`. Note `getVideoFrameUrl` cannot be reused as-is: `video-thumbnail.ts:10` still allowlists `engcomfy.com` while all 584 live hub thumbnails are on `comfy-hub-assets.comfy.org`, so it returns null for every one — which also means the existing fallback is dead in production. Fix: widen that constant first, then reuse the pattern.

site/src/components/workflow-pages/LandingHero.astro
```diff
@@ -49,8 +54,17 @@ const ctaButtons = [
...
+// app thumbnails are often video-only) fronts its own page as a muted loop
+// instead of yielding the hero to a later template's still.
+const leadTemplate = templates[0];
+const leadVideoThumb =
+  leadTemplate && !hasStillThumbnail(leadTemplate.thumbnails)
+    ? leadTemplate.thumbnails?.find((thumb) => isVideoFile(thumb))
+    : undefined;
+const heroTemplate = leadVideoThumb ? leadTemplate : firstTemplateWithStill(templates);
+const heroVideoSrc = leadVideoThumb ? thumbnailPath(leadVideoThumb) : null;
+const heroImage = leadVideoThumb ? null : firstStillThumbnail(heroTemplate?.thumbnails);
```
MaanilVerma: nit (non-blocking): this ternary is dead — `leadVideoThumb` is only set inside the `!hasStillThumbnail(leadTemplate.thumbnails)` guard, and line 65 then assigns `heroTemplate = leadTemplate`, so `firstStillThumbnail` on that same array is necessarily `null`. Both branches yield `null`. Suggest dropping the condition.

site/src/components/workflow-pages/LandingHero.astro
```diff
@@ -49,8 +54,17 @@ const ctaButtons = [
...
-const heroImage = firstStillThumbnail(heroTemplate?.thumbnails);
+// A video-only lead template (curated pages pin the featured app first, and Hub
+// app thumbnails are often video-only) fronts its own page as a muted loop
+// instead of yielding the hero to a later template's still.
+const leadTemplate = templates[0];
+const leadVideoThumb =
+  leadTemplate && !hasStillThumbnail(leadTemplate.thumbnails)
+    ? leadTemplate.thumbnails?.find((thumb) => isVideoFile(thumb))
+    : undefined;
+const heroTemplate = leadVideoThumb ? leadTemplate : firstTemplateWithStill(templates);
```
MaanilVerma: suggestion: `heroTemplate` now resolves to `templates[0]` on a video-led page, but `use-cases/[slug].astro:107` and `model/[name].astro:154` still derive the reserved hero from `firstTemplateWithStill`, so the two picks diverge — 5 use-case pages and 8 qualifying model families against today's hub index. The page then withholds from `resolveRailImages` a still it never renders, while the hero that is rendered goes unreserved. This falsifies the contract stated verbatim at `media-utils.ts:62`. Both selectors already filter media files so nothing double-renders — the defect is silent divergence, not a visible bug. Suggest exporting the lead-aware pick from `media-utils.ts` so all three call sites share it.

site/src/lib/workflow-pages/use-cases.ts
```diff
@@ -271,5 +344,154 @@ export const SEO_PAGES: SeoPageDef[] = [
...
+        'image to 3d mesh',
+      ],
+    },
+    // Fully curated: the 3D Model cluster (Hunyuan plus the multi-angle camera
+    // .app) lives only on the hub, so the on-disk snapshot the sitemap reads
+    // can't resolve a `3D Model` tag against it. A tag filter would render the
+    // page from the live hub yet silently drop it from the sitemap; pinning the
+    // cluster keeps the page both indexable and listed. Maanil's thumbnail-less
+    // test workflow (2030b1e2fb72) is simply left unpinned.
+    filters: {},
```
MaanilVerma: question: the comment above says the `3D Model` cluster lives only on the hub, but that tag matches 13 templates in this branch's own `templates/index.json` — the count is 0 only on `main`, because `04f33569` renamed the vocabulary to `3D` (37) and `Image to 3D` (20). The contract tripped from catalog tag drift, not from a hub-only cluster, so `filters: { tags: ['Image to 3D'] }` would have restored the page and kept the registry self-maintaining instead of hardcoding eight share ids. `ai-tattoo-generator` carries the same rot — `Canny` is 0 on main and survives only via `ControlNet`. Could we retag rather than curate here?

site/src/components/hub/WorkflowGrid.vue
```diff
@@ -129,7 +136,13 @@ const sortedTemplates = computed(() => {
     });
   }
 
-  return result;
+  if (props.pinnedShareIds.length === 0) return result;
+  const pinRank = new Map(props.pinnedShareIds.map((id, index) => [id, index]));
```
MaanilVerma, line 140: suggestion: this `Map` keeps the last index for a repeated key, while `use-case-resolver.ts:42` dedupes pins first-wins, so one pin-ordering rule now has two hand-rolled implementations with opposite duplicate semantics. `[slug].astro:238` feeds the raw `def.pins` rather than the resolved list, so the two derivations drift independently. No page repeats a pin today, so this is latent — and the client re-sort is genuinely required, so the fix is not to delete this block but to export one `orderByPins` used by both. That would also make the ordering unit-testable, which it currently is not.

MaanilVerma (commented):
suggestion (non-blocking): comment density in the PR in several files — several comments like in `use-cases.ts` restate what the line beneath them already says, and the "No appShareId: App Mode not verified in cloud yet" rationale is repeated near-verbatim five times (lines 369, 396, 429, 457, 483). Comments that encode a temporary state also go stale silently once App Mode is verified, and the file header still promises the grid is filter-resolved "with no list to maintain" while 9 of 15 entries are now curated. Could we keep the per-pin `// Topaz: Image Enhance` labels, which carry real provenance, and lift the repeated rationale into one note on `SeoPageDef.appShareId`?

site/src/lib/workflow-pages/use-cases.ts
mobeenabdullah: Nailed it, thank you. Confirmed the counts: `3D Model` is 0 on main / 13 on this branch, and CI builds the PR merged with main, so the tag resolved empty and dropped the page from the sitemap. Retagged to `Image to 3D` (20 on main) and reverted the hardcoded pins, so the grid is self-maintaining again; kept the two curated pins (Hunyuan lead + the 3D camera .app the tag can’t reach) and the exclude. Verified locally: the page renders a 13-workflow grid. On ai-tattoo: `Canny` is indeed dead on main, but `ControlNet` keeps it indexed, so I left it out of this change to avoid disturbing the tuned interior/tattoo split — happy to drop the dead `Canny` in a follow-up if you’d prefer. (771f0e6)

site/src/components/workflow-pages/LandingHero.astro
mobeenabdullah: Fixed. Exported `heroTemplateFor` from media-utils (lead-aware, mirroring `firstVisualAcross`) and used it in LandingHero plus both hero derivations in `use-cases/[slug].astro` and `model/[name].astro`, so the reserved hero can no longer diverge from what renders. Confirmed `heroTemplateName`/`heroImage` feed only the rail reservation (not OG/JSON-LD), so no OG regression, and added a unit test for the pick. (771f0e6)

site/src/components/workflow-pages/LandingHero.astro
mobeenabdullah: Fixed. With the lead-aware pick flowing through `heroTemplateFor`, heroImage is just `firstStillThumbnail(heroTemplate?.thumbnails)` (null for a video-only lead) with no guard. (771f0e6)

site/src/components/workflow-pages/LandingHero.astro
mobeenabdullah: Genuine, and I verified it: the image-to-3d hero video is served from `comfy-hub-assets.comfy.org`, while `video-thumbnail.ts` only allowlists `engcomfy.com`, so `getVideoFrameUrl` returns null and the poster/fallback is dead exactly as you said. I did not fold it into this PR because widening a shared CDN allowlist has site-wide blast radius (every getVideoFrameUrl caller) and I want to confirm comfy-hub-assets supports Cloudflare `cdn-cgi/media` before flipping it, plus the LCP claim needs a Lighthouse check. Tracked in #1070 (widen the constant, then add poster + preload + the ThumbnailDisplay 3s paused/readyState fallback). Shout if you’d rather it block this PR.

site/src/components/hub/WorkflowGrid.vue
mobeenabdullah: Agreed, good catch on the opposite dedup semantics (last-wins Map here vs first-wins in the resolver, fed raw `def.pins`). It is latent today (no page repeats a pin) and the client re-sort is genuinely needed, so I tracked the proper fix in #1071: extract one `orderByPins(templates, pinnedShareIds)` used by both, with a unit test.

mobeenabdullah:
@MaanilVerma thanks for the thorough pass. Addressed in 771f0e6:

- **Comment density:** lifted the repeated `No appShareId: App Mode not verified` rationale into one note on `SeoPageDef.appShareId`, dropped the five near-duplicate per-entry comments, kept the per-pin provenance labels (`// Topaz: Image Enhance`), and corrected the file header that still claimed every grid is filter-resolved (9/15 are curated).
- **image-to-3d tag drift:** retagged to `Image to 3D`, reverted the hardcoded pins (details on the inline thread).
- **Hero pick divergence + dead ternary:** shared `heroTemplateFor` across LandingHero and both page hero derivations, with a unit test.
- **Video hero poster/allowlist** and **pin-ordering unification:** genuine but larger/riskier, tracked as #1070 and #1071 rather than rushed in here.

Verified locally (lint, typecheck, 41 unit tests) and visually via Playwright (index shows 15, image-to-3d renders its video hero + 13-workflow grid, no regressions).

christian-byrne (approved):

PR #1061, opened 2026-07-29
chore(i18n): update hub workflow translations

comfy-legal:
Automated hub translation refresh. English content regenerated from the Hub index, human seeds applied, gaps machine-translated, deterministic validators run. A partial (rate-limited) run still commits its progress to this same branch so the next run resumes. Does not flip any locale to indexable.


deepme987 (approved):
blind approval, because i can't actually review any of the translations

PR #1068, opened 2026-07-31
fix(ci): build previews against the production hub so all use-case pages render

mobeenabdullah:
## Problem

Preview deploys build against the preview hub (`HUB_API_URL_PREVIEW`), which does not contain the curated/shared workflows that the "fully curated" use-case pages pin. Since `getStaticPaths` skips any use-case page whose grid resolves to zero templates, those pages never build and 404 on the preview, even though they render fine on production and locally. On the use-cases index this shows up as only the tag-filtered pages appearing (7 of 15 on the use-case PR, 10 of 10 pages that exist on main).

## Fix

Point the preview build at the production hub, using the existing `HUB_API_URL_PRODUCTION` / `COMFY_CLOUD_URL_PRODUCTION` secrets:

```
PUBLIC_HUB_API_URL:      HUB_API_URL_PREVIEW      -> HUB_API_URL_PRODUCTION
PUBLIC_COMFY_CLOUD_URL:  COMFY_CLOUD_URL_PREVIEW  -> COMFY_CLOUD_URL_PRODUCTION
```

No new secrets; the production values already exist.

## Verified

Ran the same change on a throwaway build of the use-case PR branch. The preview went from 7 to all 15 use-case pages, and every previously-404 pin-only page (restore-old-photos, ai-avatar-generator, and the rest) now returns 200. On a main-based build, all 10 use-case pages render.

## Tradeoff

The preview build gets slower (roughly 12 to 13 minutes vs 4 to 5), because it now builds the full production workflow catalog instead of the smaller preview set. Flagging in case keeping previews isolated from production data was intentional; if so, the alternative is to seed the curated workflows into the preview hub instead.

### DIFF
--- a/.github/workflows/preview-site.yml
+++ b/.github/workflows/preview-site.yml
@@ -42,8 +42,8 @@ jobs:
         run: pnpm run build
         working-directory: site
         env:
-          PUBLIC_HUB_API_URL: ${{ secrets.HUB_API_URL_PREVIEW }}
-          PUBLIC_COMFY_CLOUD_URL: ${{ secrets.COMFY_CLOUD_URL_PREVIEW }}
+          PUBLIC_HUB_API_URL: ${{ secrets.HUB_API_URL_PRODUCTION }}
+          PUBLIC_COMFY_CLOUD_URL: ${{ secrets.COMFY_CLOUD_URL_PRODUCTION }}
 
       - name: Deploy preview to Vercel
         id: deploy
--- a/.github/workflows/preview-site.yml
+++ b/.github/workflows/preview-site.yml
@@ -44,6 +44,7 @@ jobs:
         env:
           PUBLIC_HUB_API_URL: ${{ secrets.HUB_API_URL_PRODUCTION }}
           PUBLIC_COMFY_CLOUD_URL: ${{ secrets.COMFY_CLOUD_URL_PRODUCTION }}
+          PUBLIC_APPROVED_ONLY: 'true'
 
       - name: Deploy preview to Vercel
         id: deploy


.github/workflows/preview-site.yml
```diff
@@ -42,8 +42,8 @@ jobs:
         run: pnpm run build
         working-directory: site
         env:
-          PUBLIC_HUB_API_URL: ${{ secrets.HUB_API_URL_PREVIEW }}
-          PUBLIC_COMFY_CLOUD_URL: ${{ secrets.COMFY_CLOUD_URL_PREVIEW }}
+          PUBLIC_HUB_API_URL: ${{ secrets.HUB_API_URL_PRODUCTION }}
```
dante01yoon, line 45: issue: Switching this job to the production Hub while leaving `PUBLIC_APPROVED_ONLY` unset makes `listWorkflowIndex()` request `pending,approved,rejected,deprecated`. The resulting Vercel URL is posted on a public PR, so this can publish production workflows that the production deploy intentionally excludes with `PUBLIC_APPROVED_ONLY: 'true'`. Could this build set the approved-only flag as well? The curated workflows needed by these use-case pages should still be present, without exposing unapproved/rejected production content.

dante01yoon (changes requested):
One blocking moderation-boundary issue was found. The production endpoint should retain the production approved-only filter in publicly posted PR previews.

.github/workflows/preview-site.yml
mobeenabdullah: Good catch, thank you. You are right: with the production Hub and `PUBLIC_APPROVED_ONLY` unset, `listWorkflowIndex()` requests `pending,approved,rejected,deprecated`, and the preview URL is public, so that would expose production workflows the production deploy hides. Fixed by matching production and setting `PUBLIC_APPROVED_ONLY: 'true'` in this build.

Verified on a throwaway build of the use-case PR branch with approved-only: the Astro build succeeded and all 15 use-case pages built, and `assertCuratedSharesResolve` passed (it fails the build if any pinned/curated share is missing), so the curated workflows are all `approved` and still present, with no unapproved/rejected content exposed. (The test deploy itself then hit Vercel's 24h upload rate limit from repeated test builds, unrelated to this change.) Re-requesting your review.

dante01yoon (approved):
The blocking moderation-boundary issue is resolved: PR previews now query only approved production workflows, and the curated-share assertion confirms the intended use-case pages still build. Current checks pass.

suggestion: As a non-blocking follow-up, please update `README.md` and `site/README.md`; they still describe `preview-site.yml` as using the preview/configurable API with no status filter.

PR #1086, opened 2026-08-05
fix(site): read app mode from the hub field, drop the filename guess

mobeenabdullah:
## Scope reduced, please re-read rather than re-reviewing the old diff

This started as a fix for workflows not appearing under the Comfy Apps tab, and shipped a committed snapshot as a stand-in for data the hub did not yet expose. **That data now exists**, so most of the PR is obsolete and I have removed it. It is now a two-line change.

## The problem

App-ness was inferred from a `.app` filename suffix:

```ts
isApp: entry.isApp ?? entry.name.endsWith('.app'),
```

Wrong in both directions. It misses apps not named that way, and `templates_all_in_one_image_edit_models.app` carries the suffix while its graph sets `linearMode: false`.

## What changed underneath

`Comfy-Org/cloud#6399` shipped `is_app` and its backfill has run, so the hub answers directly. Measured against the live catalogue:

```
entries:               609
carry the isApp field: 609
isApp = true:          56
name ends in .app:     14
```

Since the field is present on every entry, the `??` fallback never fires. The guess is already dead code.

## Why the snapshot is gone

I checked what it would change before defending it:

```
main shows today: 56
this PR shows:    56
  added:  0
  removed: 0
```

No user-visible difference. And the committed snapshot had **already drifted**, missing "Interior Design - Add furniture into a Render" within days of being generated. A file that has to be refreshed by hand, to guard against a failure mode that no longer exists, costs more than it protects. Removed, along with its refresh script and the shared status list that only existed to serve it.

## What is left, and why it is still worth merging

Deleting the guess changes how this fails. If the hub field ever regressed, the old code would quietly show 14 workflows, some wrong, and nobody would notice for weeks. Now the tab shows nothing, which is obvious and gets reported.

## Note for @MaanilVerma

Your status-mismatch finding was real and is dropped with the script it applied to: the refresh script hardcoded `status=approved` while `listWorkflowIndex` asks for all four on non-production builds, so a pending app rendered in preview but was missing from the snapshot. With no snapshot there is no mismatch. Worth recording in case a snapshot is ever reintroduced.

## Verification

- 553 unit tests, eslint clean, `astro check` 0 errors
- Diff against main is exactly the two `isApp` lines, no unrelated churn. `hub-api.ts` fails `prettier --check` on a long line that is **already unformatted on main** and untouched here; reformatting it would be noise in this PR


site/scripts/refresh-app-mode-snapshot.ts
```diff
@@ -0,0 +1,108 @@
...
+
+interface AppModeWorkflow {
+  shareId: string;
+  title: string;
+}
+
+async function hubGet<T>(path: string): Promise<T> {
+  const res = await fetch(`${HUB_API_BASE}${path}`, { headers: { Accept: 'application/json' } });
+  if (!res.ok) throw new Error(`${res.status} ${res.statusText} — ${path}`);
+  return (await res.json()) as T;
```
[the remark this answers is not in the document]
mobeenabdullah: Good catch, fixed. Every request now carries a 30s deadline via `AbortSignal.timeout(REQUEST_TIMEOUT_MS)`. This mattered more than it looks: the script deliberately aborts rather than writing a shorter snapshot, so a single stalled connection would have hung the whole refresh instead of failing it. 30s is generous because workflow graphs are large payloads. eslint clean, unit tests 24/24, typecheck 0 errors.

mobeenabdullah:
Follow-up needed here once #1104 merges, raised by @christian-byrne while reviewing that PR.

`resolveIsApp` in `site/src/lib/hub-app-mode.ts` falls back to the `.app` suffix when `isApp` is absent:

```ts
entry.isApp ?? (APP_MODE_SHARE_IDS.has(entry.shareId ?? '') || Boolean(entry.name?.endsWith('.app')))
```

Absent is doing two different jobs there, and after #1104 they stop being the same thing.

**Today** `templates/index.json` has no `isApp` at all, so absent means "unknown" and the fallback is the only signal. Correct as written.

**After #1104** absent means "the sync looked and this is not an App". The fallback then contradicts the data it is standing in for. Verified against the actual files: of the 17 templates whose name ends `.app`, `templates_all_in_one_image_edit_models.app.json` has `extra.linearMode: false`, and the other 16 are `true`. So that one is a genuine false positive, and it survives precisely because the correct answer is expressed as an absent key.

The index path (`hub-api.ts:366`) should become `entry.isApp === true` at that point.

The hub API path (`hub-api.ts:467`) is different and should keep its fallback for now: that data comes from the cloud hub, which does not return `is_app` until Comfy-Org/cloud#6399 ships. It also passes no `name`, so only the share-id snapshot applies there.

Not changing it in this PR because the change is only correct after #1104 is merged. Doing it before would read every workflow as a node graph, since no entry carries the field yet.

Ordering: #1104, then this, then cloud#6399 retires the snapshot entirely.

site/scripts/refresh-app-mode-snapshot.ts
```diff
@@ -0,0 +1,117 @@
...
+  );
+  return results;
+}
+
+const entries = await hubGet<HubWorkflowTemplateEntry[]>(
+  '/api/hub/workflows/index?status=approved'
+);
+const withShareId = entries.filter((entry) => Boolean(entry.shareId));
+console.log(`Checking ${withShareId.length} workflows for App Mode…`);
+
```
MaanilVerma: issue: This hardcodes `status=approved`, while `listWorkflowIndex` in `hub-api.ts` only narrows to approved when `PUBLIC_APPROVED_ONLY` is set and otherwise asks for pending, approved, rejected and deprecated. Preview builds take that second path, so a pending workflow published as an app renders in preview but is absent from the snapshot the same build consults, and it shows under Node Graphs. That's the surface people use to check a workflow before approving it. Could the script request the same status set as a non-production build, so the snapshot covers everything a preview can render?

site/src/lib/hub-app-mode.ts
```diff
@@ -0,0 +1,32 @@
...
+/**
+ * Whether a catalog entry belongs under the "Comfy Apps" tab.
+ *
+ * An `isApp` from the hub takes precedence, so the snapshot stops being
+ * consulted once the API returns the field. The `.app` name suffix is a legacy
+ * naming convention retained for templates uploaded before the snapshot.
+ */
+export function resolveIsApp(entry: { isApp?: boolean; shareId?: string; name?: string }): boolean {
+  return (
+    entry.isApp ??
```
MaanilVerma: issue: `entry.isApp ?? ...` means an explicit `false` from the hub wins over both the snapshot and the `.app` suffix, and the PR frames that as the migration path: once the index returns the field, the snapshot stops being consulted. But the index endpoint returning `isApp` at all is what flips every entry to a hub value, and if the backend computes it from the same `.app` suffix this PR just showed is unreliable, the 41 workflows added here silently drop back out. The compatibility note says nothing that was listed before is dropped, which holds today but not after that change. Could we treat a hub `false` as "unknown" and OR the three signals instead, so the snapshot stays a floor until it's deliberately removed?

site/tests/unit/hub-app-mode.test.ts
```diff
@@ -0,0 +1,75 @@
...
+  ['a09d65985659', 'Image Enhancer'],
+  ['3515c5083027', 'Image to Video'],
+  ['c1959fdc5642', 'Image Upscale'],
+  ['a335e0968d76', 'Music Generator'],
+  ['d5ce59e59ff3', 'Photo to Cartoon Style Caricature'],
+  ['69850664cf89', 'Restore Old Photos'],
+  ['3ef4de40106b', 'Song Generator'],
+  ['90d086fef9e3', 'Tattoo Generator'],
+  ['bed989744195', 'Video Face Swap'],
+  ['0740bf78b7b6', 'Video Upscale'],
```
MaanilVerma: suggestion (non-blocking): These 16 ids are a hand-copied subset of the snapshot in the same commit, so the test asserts that a JSON file still contains rows someone pasted out of it. It can only fail when a refresh legitimately removes a workflow, and then it fails as a puzzle: whoever hits it has to work out whether the app was unpublished or the refresh broke. The script already aborts rather than writing a short list, which is the real guard against a partial refresh. Could we drop this list and the legacy one below it, and keep a count floor plus the well-formed-id check instead?

site/tests/unit/hub-app-mode.test.ts
```diff
@@ -0,0 +1,75 @@
...
+  it('files a workflow published with default view "app" under Comfy Apps', () => {
+    expect(resolveIsApp({ name: '69850664cf89', shareId: '69850664cf89' })).toBe(true);
+  });
+
+  it('honours the legacy .app name suffix', () => {
+    expect(resolveIsApp({ name: 'utility_z_image_turbo_2k_upscaler.app' })).toBe(true);
+  });
+
+  it('leaves a workflow published as a node graph out', () => {
+    expect(resolveIsApp({ name: 'e81f8eb0ee5f', shareId: 'e81f8eb0ee5f' })).toBe(false);
```
MaanilVerma: suggestion (non-blocking): These two read a real share id out of the committed snapshot, so they're testing the data file as much as the function. If a refresh ever unpublishes `69850664cf89` the first one fails even though `resolveIsApp` is fine. Could we cover the rule rather than the row, by picking any member of `APP_MODE_SHARE_IDS` for the true case and an id known absent from it for the false case?

site/scripts/refresh-app-mode-snapshot.ts
mobeenabdullah: Confirmed and fixed in `b6976eb`. Verified the mismatch rather than taking it on trust: `listWorkflowIndex` sends `approved` only when `PUBLIC_APPROVED_ONLY` is set and otherwise sends all four statuses, while the refresh script had `?status=approved` hardcoded. So a pending app was in the preview index and absent from the snapshot that same build reads, exactly as you said, on the surface used to review it before approving.

The script now requests the same set. Rather than duplicating the list, I moved it to `src/lib/hub-status.ts` and both import it, since `hub-api.ts` reads `import.meta.env` at load and a plain Node script cannot import from it. That was why the copy existed in the first place, so the fix seemed better than a comment asking the next person to keep them in step.

site/src/lib/hub-app-mode.ts
mobeenabdullah: Taken in `b6976eb`, though the reason turned out to be different from the one in the concern, so worth stating.

The premise does not hold: cloud#6399 derives `is_app` from `workflow_json.extra.linearMode`, never from the filename. I checked, and ComfyUI_frontend#14914 does the same, with a test asserting the suffix alone is not enough. So the 41 would not have dropped out through `.app` inference.

The exposure is narrower and certain rather than speculative. That column is written **at publish time**, so every workflow published before it existed reads `false` until the backfill runs. Under `??` that false wins, so the tab would empty for the entire window between the field shipping and the backfill completing. Your fix is right; the failure mode is a migration window, not a bad inference.

`resolveIsApp` is now `entry.isApp === true || APP_MODE_SHARE_IDS.has(...)`. Both count, neither vetoes.

I also dropped the `.app` suffix rather than OR-ing three signals. It was the original guess, it is wrong in both directions, and `templates_all_in_one_image_edit_models.app` carries the suffix with `linearMode: false`. The snapshot reads `linearMode` from all 55 real apps, including the four the suffix used to catch, which the old `LEGACY_APP_TEMPLATES` test already proved. So keeping it only preserved the one false positive.

site/tests/unit/hub-app-mode.test.ts
mobeenabdullah: Taken in `b6976eb`. Both lists are gone, replaced by a count floor and the well-formed-id check.

Your reasoning is the part I had wrong: the refresh script already aborts rather than writing a short file, so the id list was not guarding against a partial refresh, and it failed as a puzzle when a workflow was legitimately unpublished. The floor catches a truncated snapshot committed by hand without turning an ordinary unpublish into a mystery.

site/tests/unit/hub-app-mode.test.ts
mobeenabdullah: Taken in `b6976eb`. The tests now draw an id out of `APP_MODE_SHARE_IDS` for the true case and use one asserted absent from it for the false case, so they cover the rule rather than a row of the data file. Unpublishing any particular workflow no longer fails a test about resolution.

MaanilVerma:
LGTM.

Final approval rests with @dante01yoon.

mobeenabdullah:
@dante01yoon this one is ready for your review when you have a moment.

Conflicts with main are resolved as of `b286d15` and it is mergeable again. Both were from work that landed on main after this branch was cut, and both resolved by keeping both sides:

- `site/package.json`: #1100 added `partner-nodes:refresh-snapshot` on the same line this branch added `app-mode:refresh-snapshot`. Unrelated scripts, both kept.
- `site/src/lib/hub-api.ts`: main widened the ranking import to include `readEnv` while this branch added the `hub-app-mode` import above it. Both symbols are genuinely used here, `readEnv` for the API base and the approved-only flag, `resolveIsApp` in both entry mappers, so both imports kept.

I checked the merged file differs from main only by this PR's own changes, with no formatting or unrelated churn carried in.

Green after the merge: 535 unit tests, `astro check` 0 errors, eslint clean on the touched file.

**What the PR does:** workflows published as apps were not appearing under the Comfy Apps tab, because app-ness was inferred from a `.app` filename suffix. That guess is wrong in both directions: it misses apps not named that way, and `templates_all_in_one_image_edit_models.app` carries the suffix while its graph sets `linearMode: false`. This reads the author's actual setting instead.

All four of Maanil's review threads are answered. One of them turned out to be a real bug beyond the one reported: the refresh script hardcoded `status=approved` while `listWorkflowIndex` requests all four statuses on non-production builds, so a pending app rendered in preview but was missing from the snapshot the build reads. The status list now lives in `hub-status.ts` and both sides import it.

No rush from my side, and happy to walk through any of it.

deepme987 (approved):

PR #1089, opened 2026-08-06
feat(site): enforce the brand-safety gate on pins, and record the gated workflows

mobeenabdullah:
## What

Makes the keyword sheet's brand-safety gate a real thing in code, and records three gated workflows against the pages they belong on without letting them render.

`SeoPagePin` gains an optional `gate`. `resolveUseCasePageTemplates` drops any pin that declares one, so a gated pin can sit in the registry documenting the intended placement while staying out of the grid. Removing the field is the one-line change that publishes it, once the gate's conditions ship.

| Workflow | Page | Gate |
| --- | --- | --- |
| Video Face Swap (`bed989744195`) | `ai-character-replacement` | `GATED` |
| Face Swap (`c2aae816fe63`) | `ai-character-replacement` | `GATED` |
| Clothes Changer (`8ce4aa90e8af`) | `ai-hairstyle-changer` | `GATED-LITE` |

## Why the resolver, and not just leaving them out

`assertBrandSafe` only inspects a page's slug, keywords and title. Pins are never checked, so a pin has always been the one route by which gated material could reach a public grid, and a green build was never evidence that a pinned workflow was cleared. This closes that gap for every future pin, not just these three.

That was also CodeRabbit's read on the first revision of this PR: either drop the pins, or enforce the gate in the resolver. This does the second, which keeps the intended placement recorded instead of losing it in a closed PR.

## Why these three are worth recording

`ai-character-replacement` resolves three workflows and **zero Apps**. Face swap is the only App in the catalog that fits the page, so while the gate holds, that page cannot meet the "every use-case page lands in a runnable app" goal at all. Two of the fifteen live pages are in the same position for the same reason, so it is worth having the blocker visible in source rather than only in a spreadsheet.

## What the gate says

- **face swap / face swap video (`GATED`)**: "Per Nav 07-14 conditions: consent flow, no celebrity presets, filters, C2PA". The sheet also flags the SERP adjacency being avoided.
- **clothes changer (`GATED-LITE`)**: "Outfit/try-on framing only". It sits on the hairstyle page, which frames it as trying on a new look rather than removing clothing.

I could not find confirmation that the face-swap conditions shipped, so nothing here renders.

## Ask

@nav, when the conditions are met, deleting the `gate` field on a pin publishes it; no other change is needed. Two questions when you have a moment:

1. Do those conditions apply to a workflow appearing in a page's grid, or only to a page targeting the keywords?
2. All three are **already live and publicly browsable on the hub today**, the gate has only ever kept them off the curated pages. If the intent was that they not be publicly reachable at all, that is a separate and more urgent fix than this PR.

## Verification

- New tests: a `GATED` pin never renders, and a `GATED-LITE` pin is dropped while ungated pins beside it survive.
- use-case resolver suite 21/21, eslint clean, `astro check` 0 errors.
- Screenshot of how the pages look with the pins active are attached in the thread below, taken from a local build.

gated-grid-face-swaps


site/src/lib/workflow-pages/use-cases.ts
```diff
@@ -346,6 +346,14 @@ export const SEO_PAGES: SeoPageDef[] = [
...
     // No single clean tag covers them, so filter by the model.
     filters: { models: ['wan2.2 Animate'] },
+    // PROPOSAL, pending sign-off on the sheet's face-swap gate. Face swapping is
+    // the narrow form of character replacement, and without these the page has no
+    // App at all. Pins only: the keywords above must never gain face-swap terms,
+    // because the governance denylist fails the build on them.
+    pins: [
+      { shareId: 'bed989744195' }, // Video Face Swap
+      { shareId: 'c2aae816fe63' }, // Face Swap (image)
+    ],
```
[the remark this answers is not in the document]
mobeenabdullah: Took the second option: the gate is now enforced in the resolver rather than the pins being dropped.

`SeoPagePin` carries an optional `gate` (`GATED` | `GATED-LITE`), and `resolveUseCasePageTemplates` drops any pin declaring one, so these three sit in the registry recording their intended placement without reaching the grid. Deleting the field is the one-line change that publishes a pin once its conditions ship.

You identified a real structural gap, and it was wider than these three pins: `assertBrandSafe` only inspects a page's slug, keywords and title, so a pin has always been the one route by which gated material could reach a public grid — a green build was never evidence a pinned workflow was cleared. That is now closed for every future pin.

Added runtime coverage as you noted it was missing: a `GATED` pin never renders, and a `GATED-LITE` pin is dropped while ungated pins beside it survive. Resolver suite 21/21, eslint clean, astro check 0 errors.

site/tests/unit/fallback-pool.test.ts
```diff
@@ -0,0 +1,46 @@
...
+}
+
+describe('GATED_SHARE_IDS', () => {
+  it('collects every gated pin across all pages', () => {
+    const expected = SEO_PAGES.flatMap((def) =>
+      (def.pins ?? []).filter((pin) => pin.gate).map((pin) => pin.shareId)
+    );
+    expect([...GATED_SHARE_IDS].sort()).toEqual([...new Set(expected)].sort());
+  });
+});
```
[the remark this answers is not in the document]
mobeenabdullah: Fair point, that assertion was tautological. Deleting every `gate` field would have left it green.

The three ids are now named literally in `EXPECTED_GATED_SHARE_IDS`, with the derived check kept alongside so the two have to agree. Publishing a gated workflow is meant to be a deliberate, reviewable change, so updating this list is part of that change rather than a maintenance cost.

The only-candidate test no longer skips itself either; it asserts each expected id is dropped from the pool.

deepme987 (approved):

mobeenabdullah:
Resolved the conflict with main and pushed `17fc513`. @deepme987 flagging you since you approved just before this landed, though nothing in the resolution changes what you reviewed.

The conflict came from #1079 merging (`34827f9`), which touches the same `use-cases.ts`. Both hunks were the same shape: main carries the pre-gate wording, where the face-swap and Clothes Changer workflows are described in a comment and simply left unpinned, and this branch replaces that with real `pins` entries carrying an inert `gate` field. Since converting exactly that omission into an explicit, enforced gate is the point of this PR, the resolution keeps this branch's version in both places.

Checked that the merge kept main's side too: the new `ai-image-extender` use case from #1079 is present, and all three gated pins survived (`bed989744195`, `c2aae816fe63` as `GATED`, `8ce4aa90e8af` as `GATED-LITE`).

Nothing is un-gated by this. All three stay inert.

Verified after resolving: 406 unit tests pass across the 36 site suites, including `use-case-resolver` and `fallback-pool`, plus eslint and prettier clean on the touched file.

A plain merge rather than a rebase, so your approval was not dismissed.

site/src/lib/workflow-pages/use-case-resolver.ts
```diff
@@ -37,9 +37,11 @@ export function resolveUseCasePageTemplates<T extends FilterableTemplate>(
     .filter((template) => !template.shareId || !excluded.has(template.shareId))
     .sort(byUsageDesc);
 
-  // Pins bypass excludes; first id wins.
+  // Pins bypass excludes; first id wins. A pin carrying a brand-safety gate is
+  // dropped here, because pins are the one route that skips `assertBrandSafe`.
   const seen = new Set<string>();
   const pinned = (def.pins ?? []).flatMap((pin) => {
+    if (pin.gate) return [];
```
MaanilVerma, line 44: issue: `useCasePageHasGrid` a few lines below still counts gated pins. It falls back to `def.pins?.length` for a page with no filters, and that count doesn't know about `gate`, so the sitemap sees a grid the resolver refuses to render. `ai-hairstyle-changer` has `filters: {}` and two pins, one of them the GATED-LITE Clothes Changer, so today it survives on the ungated Hairstyle Changer alone. But the moment a fully-curated page's only pins are gated, it goes into the sitemap as indexable while rendering an empty grid, which is exactly what `verify-sitemap-indexability` exists to catch. Could we filter on `!pin.gate` in that count too, and add a case for a curated page whose only pin is gated?

site/src/lib/workflow-pages/use-cases.ts
```diff
@@ -511,3 +527,16 @@ export const SEO_PAGES: SeoPageDef[] = [
...
+/**
+ * Every share id carrying a brand-safety gate, across all pages.
+ *
+ * A gate declared on one page's pin has to hold everywhere, not just on that
+ * page's grid. The landing-page image matcher draws from a whole-catalog
+ * fallback pool, so without a global set a gated workflow can still surface as a
+ * capability card (with a live CTA to it) on a page that never pinned it. That
+ * is how `8ce4aa90e8af` reached `ai-hairstyle-changer`.
+ */
+export const GATED_SHARE_IDS: ReadonlySet<string> = new Set(
```
MaanilVerma: issue: Deriving the global blocklist from pins means a workflow is only gated if some page happens to pin it. The PR frames this as a catalog-wide safety net, but a gated workflow nobody pinned isn't in `GATED_SHARE_IDS`, so it stays in the fallback pool and can still surface as a capability card. It also couples the two meanings of `gate`: pin it and you're declaring both "don't render here" and "withhold everywhere". If someone later removes the hairstyle pin as editorial cleanup, Clothes Changer silently rejoins the pool on every landing page. Could the gated set be its own top-level constant that pins reference, so withholding a workflow doesn't depend on a page choosing to pin it?

site/tests/unit/fallback-pool.test.ts
```diff
@@ -0,0 +1,55 @@
...
+
+describe('GATED_SHARE_IDS', () => {
+  it('holds exactly the share ids gated today', () => {
+    expect([...GATED_SHARE_IDS].sort()).toEqual(EXPECTED_GATED_SHARE_IDS);
+  });
+
+  it('stays in step with the gate declarations in SEO_PAGES', () => {
+    const declared = SEO_PAGES.flatMap((def) =>
+      (def.pins ?? []).filter((pin) => pin.gate).map((pin) => pin.shareId)
+    );
```
MaanilVerma: suggestion (non-blocking): This second test recomputes `GATED_SHARE_IDS` with the same flatMap the export uses, so it compares the implementation against a copy of itself and passes for any input. The test above it already pins the real contract by naming the three ids literally. Could we drop this one and keep the literal list as the single tripwire?

site/tests/unit/fallback-pool.test.ts
```diff
@@ -0,0 +1,55 @@
...
+  // catalog. Without this a gated workflow surfaces as a capability card, with a
+  // live CTA to it, on a page that never pinned it.
+  it('drops every gated share id, so no section can surface one', () => {
+    const gated = [...GATED_SHARE_IDS];
+    const catalog = [tpl('ungated'), ...gated.map((id) => tpl(id))];
+    const pool = buildFallbackPool(catalog);
+    expect(pool.map((t) => t.shareId)).toEqual(['ungated']);
+  });
+
+  it('drops a gated id even when it is the only candidate', () => {
```
MaanilVerma: suggestion (non-blocking): This loop asserts the same property as the test right above it, just one id at a time instead of all together. Both prove a gated id never reaches the pool. Could we drop this one, since the "only candidate" framing doesn't exercise a different code path, the filter runs per template either way?

site/src/lib/workflow-pages/use-cases.ts
```diff
@@ -346,10 +352,16 @@ export const SEO_PAGES: SeoPageDef[] = [
...
-    // would fit this page's intent, but the keyword sheet gates them pending the
-    // consent flow, celebrity-preset block, filters and C2PA. They stay off the
-    // site until that ships; the governance denylist already blocks the terms.
+    // Face swapping is the narrow form of character replacement, and without
+    // these the page has no App at all. Both stay inert while `gate` is set:
+    // the sheet holds them for a consent flow, a celebrity-preset block,
+    // filters and C2PA. Drop the `gate` field once those ship. The keywords
+    // above must never gain face-swap terms regardless, because the governance
+    // denylist fails the build on them.
+    pins: [
```
MaanilVerma: suggestion (non-blocking): There's a lot of prose per line of code across this diff, and much of it is history rather than behavior. This block replaces a comment that said the same thing, the `gate` field's own docblock already explains why gated pins are recorded instead of deleted, and the resolver, the fallback pool, the `.astro` file and two tests each restate it again. Comments describing why an earlier approach was rejected go stale the fastest, since nobody updates them when the reasoning moves on. Could we keep the rationale in the `gate` docblock as the single place it lives, and cut these to what the code can't say, in this case just which workflow each share id is?

site/src/lib/workflow-pages/use-case-resolver.ts
mobeenabdullah: Confirmed and fixed in `feb76b5`.

Reproduced it first: a curated page (`filters: {}`) whose only pin is gated returns `true` from `useCasePageHasGrid` while `resolveUseCasePageTemplates` returns `[]`. The test I added fails on the old code with exactly that mismatch, so it pins the behaviour rather than restating it.

`useCasePageHasGrid` now counts ungated pins only, matching what the resolver actually renders. Added both cases you implied: a curated page whose only pin is gated, and one that still has an ungated pin so the fix does not over-correct.

And you were right that `ai-hairstyle-changer` only survives by luck today. That is no longer load-bearing.

site/src/lib/workflow-pages/use-cases.ts
mobeenabdullah: Taken, restructured in `feb76b5`.

You named the actual defect: the two meanings of `gate` were coupled. A pin saying "do not render here" and the catalog saying "withhold everywhere" are different statements, and deriving the second from the first made withholding depend on an editorial decision.

`GATED_SHARE_IDS` is now declared explicitly, one line per id with the reason it is held, and pins reference those ids rather than defining them. So a workflow nobody pins is still withheld, and your hairstyle-cleanup scenario no longer republishes Clothes Changer.

On keeping them in step, I made the guard deliberately one-directional. `gatedPinsMissingFromWithheldSet()` fails if a page gates a pin that is not withheld catalog-wide, because that is the direction that silently puts a workflow back on the site through the fallback pool. The reverse stays legal, since withholding something no page pins is the whole point of separating them.

Mutation-checked rather than assumed: gating a pin whose id is absent from the set fails the new test, and passes again once restored.

Also took your two non-blocking test notes. The tautological pair is gone, and the fallback-pool assertion now runs against the literal ids instead of `GATED_SHARE_IDS`, so emptying that set fails the test rather than making it vacuously true.

Left alone for now: the prose density you flagged on the third thread. Happy to trim if you still want it, though a chunk of it is the brand-safety reasoning that I would rather not lose from the file.

site/src/lib/workflow-pages/use-cases.ts
mobeenabdullah: Taken in `5855d81`. I said earlier I would rather keep it, and on rereading the diff you are right: the same reasoning was written five times.

Roughly halved, 37 comment lines down to 16. What I kept is the part the code does not say:

- gated pins are dropped because pins are the one route that skips `assertBrandSafe`
- the keyword list must never gain face-swap terms, because the governance denylist fails the build on them
- the withheld-set guard is one-directional deliberately
- two tests are written against literal ids so emptying `GATED_SHARE_IDS` fails them rather than making them vacuous

What went is the history, the restatements of the `gate` docblock, and the same rationale repeated in the resolver, the fallback pool, the page and the tests.

Also merged `main`, which now carries the isApp index from #1104 and the locale listing gate from #1108. 422 tests green, prettier and eslint clean.

Your other two non-blocking notes were already done in `feb76b5`: the self-comparing test is gone, and the per-id loop is gone.

Issue #1091, opened 2026-08-06
manifest.json index sha256 hashes are stale on main; no CI gate catches template manifest drift


christian-byrne:
## Summary

`packages/core/src/comfyui_workflow_templates_core/manifest.json` on `main` records stale `sha256` values for `index.json` and all 12 locale variants. Running the documented sync step on a pristine checkout of `main` — with **no** edits of any kind — produces a 13-line diff.

## Reproduction

```bash
git clone https://github.com/Comfy-Org/workflow_templates
cd workflow_templates          # currently at 5097de61
git status --porcelain         # clean
python scripts/sync/sync_bundles.py
git diff --stat
#  .../comfyui_workflow_templates_core/manifest.json | 26 +++++++++++-----------
```

The 13 entries whose hashes change:

`index.json`, `index.ar.json`, `index.es.json`, `index.fa.json`, `index.fr.json`, `index.ja.json`, `index.ko.json`, `index.mcp.json`, `index.pt-BR.json`, `index.ru.json`, `index.tr.json`, `index.zh.json`, `index.zh-TW.json`

## Cause

https://github.com/Comfy-Org/workflow_templates/pull/1088 ("Archive Getting Started category and Reve templates") moved templates from `templates/` into `archived/` and edited all 13 `templates/index.*.json` files. The manifest hash for each of those is computed over the **pip-filtered** index that `filter_index_for_pip` produces (`scripts/sync/sync_bundles.py:126-141`), so moving templates out changes every one of them. The manifest committed in that PR does not reflect the final state of those files.

`filter_index_for_pip` is a pure deterministic JSON transform with no environment dependence, so this is genuine drift and not a Python-version artifact — reproduced locally on 3.12.

## Why CI did not catch it

Neither manifest-related workflow actually compares committed state against generated state for templates:

- `.github/workflows/validate-manifests.yml` runs `sync_bundles.py` **first** and then validates, so the manifest it checks is always the freshly generated one. It only detects orphaned templates/blueprints, never staleness.
- `.github/workflows/version-check.yml` also re-runs the sync, but its commit step bails out unless the root `pyproject.toml` version changed: *"Root version unchanged; skipping commit (no version bump performed)"*. A PR that edits `templates/index.*.json` without triggering a version bump therefore leaves the manifest stale on `main`.

[...]

PR #1092, opened 2026-08-06
Drop the duplicate first image from the Qwen-Image-Layered templates

christian-byrne:
## The bug

`EmptyQwenImageLayeredLatentImage` allocates `layers + 1` temporal slots:

```python
# comfy_extras/nodes_qwen.py:129
latent = torch.zeros([batch_size, 16, layers + 1, height // 8, width // 8], device=...)
```

Slot `0` is the model's **regenerated copy of the original image**, not a layer.

Both templates then feed the sampler output straight into `LatentCutToBatch(dim="t", slice_size=1)`, which reshapes the *entire* temporal axis into the batch dimension, and decode all of it. Nothing downstream drops anything, so a user who asks for 2 layers is handed 3 images with no indication that the first is a duplicate. The templates ship `layers = 2`.

Confirmed by comfyanonymous:

> yeah our qwen layered template is slightly wrong, the first image is exactly the same as the original and should be discarded

> and all the layers when composited together should recreate the original image

## The fix

Insert a `LatentCut(dim="t", index=1, amount=16384)` between `KSampler` and `LatentCutToBatch` in each affected subgraph:

```
KSampler -> LatentCut("t", 1, 16384) -> LatentCutToBatch("t", 1) -> VAEDecode
```

Before and after node chain

> The figure above is rendered **directly from the template JSON** — node positions, sizes, titles, slot types, widget values and link routing all come from the file (`before` = `origin/main`, `after` = this branch). It is deliberately *not* presented as a screenshot of a running ComfyUI: I do not have the Qwen-Image-Layered weights (~50 GB VRAM) and would not fake a UI capture.

`LatentCut` is existing core (`comfy_extras/nodes_latent.py:175-215`, added in https://github.com/comfyanonymous/ComfyUI/pull/9609) — older than `LatentCutToBatch` (https://github.com/Comfy-Org/ComfyUI/pull/11411), which these templates already require. No core change needed.

Cutting in **latent** space rather than adding an `ImageFromBatch` after `VAEDecode` avoids decoding a full-canvas frame only to throw it away — a whole extra VAE decode and its peak VRAM per run, on a model the template's own note already warns is slow.

`amount` is `nodes.MAX_RESOLUTION` (16384) because `LatentCut` clamps it to what exists (`nodes_latent.py:209`), so it means "everything from index 1 onward" for any `layers` value. Wiring the subgraph's `layers` input into `amount` instead would fail validation at `layers = 0` (`amount` min is 1); with the constant, `layers = 0` degrades correctly to the single original image.

## Effect

| `layers` | before | after |
| --- | --- | --- |
| 0 | 1 image (the original) | 1 image (unchanged) |
| 1 | 2 images | 1 |
| 2 (shipped default) | 3 images | 2 |
| 3 | 4 images | 3 |
| 8 | 9 images | 8 |

Correct under `batch_size > 1` as well: with `layers=3, batch_size=2` the output goes from 8 images (`[0,1,2,3,0,1,2,3]`) to 6 (`[1,2,3,1,2,3]`).

## What changed

- `templates/image_qwen_image_layered.json` — **both** subgraphs (`Image to Layers` and `Text to Layers`)
- `templates/image_qwen_image_layered_control.json` — structurally a different template, but the inner chain is the same and it has the same bug
- `blueprints/image_to_layers_qwen_image_layered.json` — this repo's copy of the blueprint
- `packages/blueprints/src/comfyui_subgraph_blueprints/blueprints/image_to_layers_qwen_image_layered.json` — the generated mirror, via `scripts/sync/sync_blueprints.py`
- manifests regenerated via `scripts/sync/sync_bundles.py` and `scripts/sync/sync_blueprints.py`

Per subgraph the diff is one new node, one new link, one re-targeted link, and the `lastNodeId` / `lastLinkId` counters. No existing node was moved or resized. The new node takes the next free `order` value, so each subgraph keeps its contiguous unique `order` sequence.

`validate_templates.py`, `validate_blueprints.py`, `validate_manifests.py` and `import_blueprints.py` (idempotency) all pass locally.

## Heads-up: this PR carries 13 manifest lines it did not cause

`scripts/sync/sync_bundles.py` also rewrote the `sha256` for `index.json` and its 12 locale variants. Those were **already stale on `main`** — running `sync_bundles.py` on a pristine `origin/main` checkout (currently `5097de61`) produces exactly the same 13 changes with no template edits at all. https://github.com/Comfy-Org/workflow_templates/pull/1088 edited all 13 `templates/index.*.json` files but the manifest it committed does not match them.

I left them in because the README forbids hand-editing the manifest and `version-check.yml` would regenerate them anyway.

**Update — both halves of https://github.com/Comfy-Org/workflow_templates/issues/1091 now have PRs:**

- https://github.com/Comfy-Org/workflow_templates/pull/1094 regenerates the manifest on `main`.
- https://github.com/Comfy-Org/workflow_templates/pull/1095 fixes the guard that let it land — `validate-manifests.yml` ran `sync_bundles.py` *before* `validate_manifests.py`, so it validated a freshly regenerated manifest and could never see drift.

**This PR needs no rebase for either.** `git merge-tree` against 1094 is clean and the merged manifest is byte-identical to the one already on this branch — git resolves the shared 13 lines automatically because both sides make the identical change. Once 1094 lands, this PR's manifest diff reduces to only the lines its own template change causes. `validate_manifests.py` already exits 0 on this branch, so it passes the new gate in 1095 as-is.

## The blueprint is duplicated across two repos, and nothing keeps them in sync

Worth stating plainly because it caught me out while preparing this:

- **Nothing syncs `Comfy-Org/ComfyUI` and `Comfy-Org/workflow_templates`.** No workflow, bot, script or CI check in either repo compares them. `scripts/blueprints/import_blueprints.py --source` is a manual, opt-in copy from an arbitrary local directory; it has no knowledge of the ComfyUI repo. Both copies are hand-committed by humans in ordinary feature PRs, which is the only reason they have stayed identical so far.
- **Neither is authoritative, and the pip package is not the bridge.** ComfyUI serves blueprints from its own git-tracked `blueprints/` directory only — `app/subgraph_manager.py:98` globs `<ComfyUI>/blueprints/*.json` with no `importlib.resources` lookup and no fallback. And although ComfyUI pins `comfyui-workflow-templates==0.11.31`, that meta package does **not** depend on `comfyui-subgraph-blueprints`, so `pip install -r requirements.txt` installs the blueprint *manifest metadata* but not a single blueprint JSON asset.
- **Net effect:** fixing this blueprint here has zero effect on a running ComfyUI, and fixing it in ComfyUI has zero effect on the PyPI/cloud consumers. Both must be patched, and a future divergence will never be caught automatically.

*Within* this repo the mirror is safe: `packages/blueprints/.../blueprints/` is generated by `scripts/sync/sync_blueprints.py` and hard-gated by the `validate` job in `.github/workflows/validate-blueprints.yml:65-73`, which re-runs the script and fails on any diff.

Filed as https://github.com/Comfy-Org/ComfyUI/issues/15363 so the duplication is tracked independently of this bug.

## Still outstanding

**The thumbnails need regenerating.** `templates/image_qwen_image_layered-1.webp` and `templates/image_qwen_image_layered_control-1.webp` still show the current output, duplicate included, so they advertise the bug in the gallery before anyone runs anything. Per this repo's README that is a manual step and needs the weights locally (~50 GB VRAM), so it needs a human with the hardware — deliberately not attempted here.

Node docs also describe the output loosely and should follow separately — `EmptyQwenImageLayeredLatentImage` in `Comfy-Org/embedded-docs` (12 locales) calls `layers` "the number of additional layers ... depth of the latent representation" without ever saying slot 0 is the original.

## Companion PR

`Comfy-Org/ComfyUI` carries an independent copy of the same blueprint at `blueprints/Image to Layers(Qwen-Image-Layered).json`, patched identically in https://github.com/Comfy-Org/ComfyUI/pull/15364. Land both together or the copies drift.


christian-byrne:
<sub>**CI note:** no GitHub Actions workflows have run on this PR yet. GitHub has had an [Actions incident](https://www.githubstatus.com/) open since 2026-08-06 15:22 UTC — *"Webhook triggers remain throttled to aid recovery, so many push and pull request events are not triggering new workflow runs."* Other PRs opened in the same window show the same empty check list. I'll re-trigger and drive the checks to green once Actions recovers.</sub>

christian-byrne:
Pushed an empty commit (`8f7b5a9b`) to re-trigger CI. This PR had **never run any repo workflow** — the original push landed during the 2026-08-06 GitHub Actions outage and only Socket and CodeRabbit had ever reported on it.

Now 11 workflows ran. **10 pass**, including the ones that matter for a templates + manifest change: Validate Templates, Validate Manifests, Validate Blueprints, Check Input Assets, Model Analysis Check, Version Check, Node Compatibility Report, Link Checker, Spellcheck, Sync Custom Nodes.

The one failure, `Build & Test`, is inherited from `main` and unrelated to this diff: `test_http_routes.py::test_static_handler_serves_samples` hard-codes template ids that https://github.com/Comfy-Org/workflow_templates/pull/1088 archived. `main` has been red on it since `5097de61` (https://github.com/Comfy-Org/workflow_templates/actions/runs/31102844125), green at the parent `850a5fad`. https://github.com/Comfy-Org/workflow_templates/pull/1094 fixes it.

Also note the `CodeRabbit` check on this PR reads `pass` with description `Review rate limited` — no review actually happened here.

comfyui-wiki (approved):

PR #1093, opened 2026-08-07
Use canonical Hub workflow view

benceruleanlu:
## What changed

- consume Cloud's canonical `default_view` / `defaultView` fields for Hub workflow classification
- derive the existing presentation-level `isApp` boolean only at serialization boundaries
- retain the `.app` filename check solely as a compatibility fallback for older API responses
- remove curated use-case pin overrides so pinned entries cannot disagree with Cloud
- add regression coverage for enum precedence, legacy fallback, and pinned-entry behavior

## Why

Hub workflows were classified as Apps through filename conventions and hand-maintained overrides. Those paths failed for publishers whose workflow names did not end in `.app`, and could disagree with the workflow's actual saved mode.

This PR makes the Hub's canonical workflow view the source of truth.

## Dependency

Depends on [Comfy-Org/cloud#6350](https://github.com/Comfy-Org/cloud/pull/6350), which adds and repairs the canonical Hub API field.

## Validation

- `pnpm exec vitest run tests/unit/hub-api-status-filter.test.ts tests/unit/use-case-resolver.test.ts` — 27 tests passed
- `git diff --check`

### DIFF
--- a/site/src/lib/hub-api.ts
+++ b/site/src/lib/hub-api.ts
@@ -21,6 +21,7 @@ const HUB_API_BASE = (import.meta.env.PUBLIC_HUB_API_URL || 'https://cloud.comfy
 export type MediaType = 'image' | 'video' | 'audio' | '3d';
 export type ThumbnailVariant = 'compareSlider' | 'hoverDissolve' | 'zoomHover' | 'hoverZoom';
 export type WorkflowStatus = 'pending' | 'approved' | 'rejected' | 'deprecated';
+export type DefaultView = 'workflow' | 'app';
 
 export interface LabelRef {
   name: string;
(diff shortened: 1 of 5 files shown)


benceruleanlu:
Closing: this consumed the `default_view` / `defaultView` field from Comfy-Org/cloud#6350, which is being closed in favor of the already-merged Comfy-Org/cloud#6399. Production serves `is_app` / `isApp` instead — which `hub-api.ts` already prefers over the `.app` filename fallback — so the site self-heals once the `hub-is-app` backfill lands.

Merged as-is, `defaultView` would always be undefined and classification would fall back to the filename heuristic, reintroducing the exact bug (GTM-386). The parts worth keeping — removing the hand-maintained `isApp: true` pin overrides in `use-cases.ts` and the regression tests for app-without-suffix / suffix-without-app — should be redone against the shipped `isApp` field after the backfill has run.

PR #1094, opened 2026-08-07
Regenerate the core manifest and unbreak Build & Test, both left broken by PR 1088

christian-byrne:
Part 1 of 2 for https://github.com/Comfy-Org/workflow_templates/issues/1091 — the stale bytes. The guard that let them land is fixed in https://github.com/Comfy-Org/workflow_templates/pull/1095, which is stacked on this branch and closes the issue.

## What is wrong

Running the documented sync step on a pristine `origin/main` — with no edits of any kind — rewrites 13 `sha256` values:

```bash
git checkout 5097de61      # origin/main
git status --porcelain     # clean
python scripts/sync/sync_bundles.py
git diff --stat
#  .../comfyui_workflow_templates_core/manifest.json | 26 +++++++++++-----------
#  1 file changed, 13 insertions(+), 13 deletions(-)
```

The committed manifest records hashes that do not match the files it packages:

```
$ sha256sum templates/index.json
a9917f8192fcaebdd2c74b86837299c7dd7aae55dc2ef56b59419c2e17bf2234  templates/index.json

$ grep -A1 '"filename": "index.json"' packages/core/src/comfyui_workflow_templates_core/manifest.json
      "sha256": "9e3ec0bd2185c80f88bf8bbc9009d529a6f51270f6f94891030153bb436a4734"
```

Affected: `index.json` and all 12 locale variants — `index.ar`, `index.es`, `index.fa`, `index.fr`, `index.ja`, `index.ko`, `index.mcp`, `index.pt-BR`, `index.ru`, `index.tr`, `index.zh`, `index.zh-TW`.

The published `comfyui-workflow-templates-core` package therefore ships a manifest whose recorded checksums fail against the index files in the same wheel.

## Where it came from

https://github.com/Comfy-Org/workflow_templates/pull/1088 moved templates into `archived/` and rewrote every `templates/index.*.json`. Because the manifest hash is taken over the pip-filtered index (`scripts/sync/sync_bundles.py:120-141`), each of those hashes changed — and the manifest committed in that PR does not reflect the final state of the files.

Bisected to that commit exactly:

| commit | `templates/index.json` actual | manifest records | |
| --- | --- | --- | --- |
| `850a5fad` (parent) | `8093af99…` | `8093af99…` | in sync |
| `5097de61` (PR 1088) | `a9917f81…` | `9e3ec0bd…` | drifted |

## This is real drift, not an environment artifact

The regenerated `manifest.json` is byte-identical under Python 3.9, 3.11 and 3.12 (`7e5d9273…` in all three), so the difference is not a serialization or dict-ordering quirk:

```
$ for v in 3.9 3.11 3.12; do uv run --python $v python scripts/sync/sync_bundles.py >/dev/null; \
    sha256sum packages/core/src/comfyui_workflow_templates_core/manifest.json; git checkout .; done
7e5d92731f4c9fbe4aa2c9629502523f1b5c3939f7722a96acd502f74409a315  …/manifest.json
7e5d92731f4c9fbe4aa2c9629502523f1b5c3939f7722a96acd502f74409a315  …/manifest.json
7e5d92731f4c9fbe4aa2c9629502523f1b5c3939f7722a96acd502f74409a315  …/manifest.json
```

## What this PR does

Nothing but `python scripts/sync/sync_bundles.py`, committing only `packages/core/src/comfyui_workflow_templates_core/manifest.json`. No template, index, bundle or blueprint file is touched. 13 insertions, 13 deletions.

`scripts/validate/validate_manifests.py` goes from exit 1 with 39 errors to exit 0:

```
before:  ❌ Found 39 error(s)                       (exit 1)
after:   ✅ No errors found (82 warning(s))          (exit 0)
```

The 82 warnings are pre-existing and expected — they are the pip-excluded templates (cloud-only and `requiresCustomNodes`), which are deliberately absent from the manifest.

## Also in this PR: `main` has been red on Build & Test since the same commit

Not the manifest, but the same PR and the same failure mode, and it has to be repaired before anything in this stack can be green.

`packages/core/tests/test_http_routes.py` hard-coded eight template ids in a `BUNDLE_SAMPLES` dict. https://github.com/Comfy-Org/workflow_templates/pull/1088 archived six of them:

```
AssertionError: No assets found for template 01_get_started_text_to_image in media-image
FAILED packages/core/tests/test_http_routes.py::test_static_handler_serves_samples
```

`main` was green at `850a5fad` and red from `5097de61` onward — https://github.com/Comfy-Org/workflow_templates/actions/runs/31102844125.

Archiving a template silently invalidated a fixture because nothing tied the fixture to the config it was sampling from. It now samples from `bundles.json` minus the pip-excluded ids, via the same `get_pip_excluded_template_names` that `sync_bundles.py` uses, so the next archive cannot rot it. It also now covers `media-assets-01`, which the hard-coded list omitted entirely. 38 tests pass locally, ruff clean.

## Follow-ups

- **The guard that should have caught this is broken, and that matters more than these 13 lines.** `.github/workflows/validate-manifests.yml` runs `sync_bundles.py` *before* `validate_manifests.py`, so it always validates a freshly regenerated manifest and can never see drift. `.github/workflows/build-test.yml` has the same ordering. Fixed in https://github.com/Comfy-Org/workflow_templates/pull/1095, which is stacked on this branch — that PR carries the red/green proof.
- https://github.com/Comfy-Org/workflow_templates/pull/1092 unavoidably carried these same 13 lines and said so. Once this lands, that PR's manifest diff reduces to only the lines its own template change causes.


christian-byrne:
@coderabbitai review

comfyui-wiki (approved):

PR #1095, opened 2026-08-07
Fail CI on manifest drift instead of regenerating the manifest before validating it

christian-byrne:
Fixes https://github.com/Comfy-Org/workflow_templates/issues/1091 — specifically the CI blind spot that let the stale manifest land. The stale bytes themselves are regenerated in part 1. **Stacked on https://github.com/Comfy-Org/workflow_templates/pull/1094** — review and merge that first; this PR's diff is the three files below.

## The guard was tautological

`.github/workflows/validate-manifests.yml` ran the sync **before** the validation:

```yaml
      - name: Sync manifests first
        run: |
          python scripts/sync/sync_bundles.py      # ← rewrites manifest.json from disk
          python scripts/sync/sync_blueprints.py

      - name: Validate manifest entries
        run: python scripts/validate/validate_manifests.py
```

`sync_bundles.py` regenerates `manifest.json` from whatever is on disk, so `validate_manifests.py` was always handed a freshly generated manifest. It could not observe drift — not because it was too weak, but because it never saw the committed file. `.github/workflows/build-test.yml` had the same ordering ahead of its `pytest` step.

That is how 13 stale `sha256` values shipped on `main`. https://github.com/Comfy-Org/workflow_templates/pull/1088 rewrote every `templates/index.*.json` without a matching manifest, and because it carried no version bump, `version-check.yml`'s commit step also bailed:

```
✅ Root version unchanged; skipping commit (no version bump performed)
```

Three workflows touched the manifest and none of them noticed.

`validate_manifests.py` itself needed no change. It already hashes every asset the manifest references against `templates/`. It only needed to be pointed at the committed state.

## Proof: red on a stale manifest, green when correct

The `main` repair that unbreaks Build & Test now lives in https://github.com/Comfy-Org/workflow_templates/pull/1094, so this PR's diff is exactly three files.

A temporary commit zeroed the `sha256` for `templates/index.json`, then was reverted. Both runs are on this branch:

| | run | result |
| --- | --- | --- |
| 🔴 stale manifest (`d8919c5f`) | https://github.com/Comfy-Org/workflow_templates/actions/runs/31135884834 | **failure** |
| 🟢 correct manifest (`bf4241e5`) | https://github.com/Comfy-Org/workflow_templates/actions/runs/31136027933 | **success** |

The red run:

```
❌ Template 'index': SHA256 mismatch for 'index.json'
   Expected: 0000000000000000000000000000000000000000000000000000000000000000
   Actual:   a9917f8192fcaebdd2c74b86837299c7dd7aae55dc2ef56b59419c2e17bf2234
❌ Found 3 error(s)
##[error]Manifest is out of sync with templates/. Run 'python scripts/sync/sync_bundles.py' and commit the regenerated manifest.
```

Under the old workflow that same commit would have passed — the sync would have rewritten the hash back before the validator ever saw it.

## Changes

**`.github/workflows/validate-manifests.yml`**

- Drop the pre-validation sync. This is the fix.
- Sparse-checkout `templates/` in full. Previously only `templates/*.json` was fetched, which was survivable only because the sync regenerated a manifest with no media entries. Validating the committed manifest requires the `.webp` files to actually be present, or every media asset reports as "not found". Offset by narrowing `packages/` to the two committed manifest files — nothing in this job reads the mirrored media, so the job fetches roughly the same volume as before.
- Trigger on changes to the workflow itself and to `list_pip_excluded.py`. Without the first, a PR that only edits this file would not run it.
- `actions/setup-python@v4` / 3.9 → `@v5` / 3.12, matching `build-test.yml`. Safe now that the job no longer runs the sync; the regenerated manifest is byte-identical under 3.9, 3.11 and 3.12 regardless.

**`.github/workflows/build-test.yml`** — validate the committed manifest immediately before the sync step. This job has no path filter, so it gates every PR, not just template-touching ones.

**`README.md`** — line 170 claimed "CI will fail if the manifest/bundles are out of sync", which was not true for templates. It now is, and the text names the two workflows that enforce it.

## Deliberately not changed

`version-check.yml` still gates its auto-commit on a version bump. The issue offered removing that condition as an alternative, but it is a release-process decision and the maintainers' comment says the gate is there to avoid churn on template-only PRs. With this PR, a template-only PR that leaves the manifest stale fails with an actionable message instead of merging silently — which is the behaviour that was missing.

## Effect on https://github.com/Comfy-Org/workflow_templates/pull/1092

It passes the new gate as-is (`validate_manifests.py` exits 0 on that branch), and it needs no rebase: `git merge-tree` against https://github.com/Comfy-Org/workflow_templates/pull/1094 is clean and the merged manifest is byte-identical to the one already on that branch.


christian-byrne:
@coderabbitai review

comfyui-wiki:
@coderabbitai review

comfyui-wiki (approved):

PR #1098, opened 2026-08-07
feat(i18n): AI translation review in the hub pipeline (GTM-291)

mobeenabdullah:
## Why

Native reviewers found four recurring problems in the machine translations that our deterministic validators structurally cannot catch:

- proper nouns that should stay English were translated (model and node names)
- translations that state something the English does not
- quality too low in some languages
- wording that is wrong for where it sits on the page

Deterministic validation answers "did the model break a rule we wrote down". It cannot answer "is this good Korean" — and that second question is what currently blocks per-language go-live, because it needs a native speaker for every language, for ~600 workflows, on every content change.

## What this adds

**`site/scripts/i18n/review-translations.ts`** reviews the *resolved* text of a whole workflow entry at once — title, description, how-to and FAQ together — so wording that is inconsistent across an entry is visible. It grades against an MQM-style rubric and returns structured JSON: `category` (terminology / accuracy / fluency / context), `severity`, the offending `span`, a `suggestion`, and an English `reason` so a reviewer who does not read the language can still triage.

Critical and major findings become `Violation`s and prune through the path that **already exists**: the field is dropped, the resolver renders English, and the page stays non-indexable. No new gating concept. Minor findings are recorded but never prune — losing a whole field over a nit would be worse than the nit.

Verdicts are keyed by a hash of (English source + translation + rubric version), so a nightly run reviews only what actually changed, and the rubric plus glossary ride in a cached system prompt rather than being re-sent per entry.

## This cannot break the translation pipeline

The review step is `continue-on-error: true`, **and** the script exits cleanly when `ANTHROPIC_API_KEY` is absent. Without the secret, or during an API outage, the pipeline behaves exactly as it does today. That is deliberate: it means this can merge before the secret is provisioned.

## Also in here

- **`enforce`: `I18N_MAX_PRUNE_FRACTION` is now validated.** An empty value became `0`, making every prune count as systemic; a typo became `NaN`, and since every comparison against `NaN` is false, the systemic guard silently stopped guarding. Opposite failures, both silent.
- **`protect`: scan every content file for sentinel collisions, not just English.** `restore` rewrites every file in the directory, so a collision in a locale file corrupts just as badly.
- **Test fix:** the preserve-term case replaced prose with itself (`.replace('在', '在')`), so it never demonstrated that terms survive an actual translation. It now rewrites all non-sentinel text.
- **`.i18nrc.cjs`:** the translation model reads `HUB_I18N_MODEL`, so it can be changed and reverted from a repo variable instead of a code change. An invalid model id fails every translation run, so being able to switch back without a PR matters.

## Verification

442/442 unit tests, `eslint --max-warnings 0` clean, prettier clean, `astro check` 0 errors.

Smoke-tested against real translation data (5 workflows from the current translation branch):

- no API key → skips cleanly, exit 0
- seeded critical finding → the field is genuinely pruned from the real `zh.json`; a minor finding on another field leaves it intact
- systemic guard fires (exit 1) when the prune fraction exceeds the threshold
- an invalid threshold warns and falls back — the pre-fix `NaN` path passed silently
- a sentinel collision planted in `zh.json` (English clean) is now refused, with `en.json` left unwritten

## Needs

- **`ANTHROPIC_API_KEY` repo secret** before review actually runs. Until then the step no-ops.
- Optional: a `HUB_I18N_MODEL` repo variable to move off `gpt-4.1`. The reviewer then measures whether the upgrade helped, rather than us assuming it did.

Does not flip any locale to indexable.


site/i18n/glossary/preserve-terms.json
```diff
@@ -37,5 +37,9 @@
   "GPT-Image",
   "Nano Banana Pro",
   "Nano Banana",
-  "Hunyuan3D"
+  "Hunyuan3D",
+  "ACE-Step",
+  "Dall-E",
+  "Fun",
+  "Qwen-Image-Layered"
```
deepme987, line 44: is this a good idea to maintain this static list?

deepme987: should the agent keep updating these if it finds something missing?

deepme987 (dismissed):

deepme987:
can you check the CI failure and coderabbit comments? Some sound right

site/scripts/i18n/enforce-translations.ts
```diff
@@ -142,7 +187,19 @@ function main(): void {
       {}
     );
 
-    const result = pruneViolatingFields(localeContent, english, locale, preserveTerms, overrides);
+    // Quality findings from the AI reviewer prune through the same path. Absent
+    // review state (reviewer not run, or no findings) contributes nothing, so the
+    // deterministic floor keeps working exactly as before on its own.
+    const reviewFindings = reviewViolations(locale, loadReviewState(locale));
```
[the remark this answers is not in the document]
mobeenabdullah: Fixed. Real bug: enforcement is not gated on the review step, so a verdict could outlive the text it described and prune a translation that had since been re-translated — possibly undoing the very fix it complained about. `reviewViolations` now takes the current English and locale content and skips any verdict whose hash no longer matches; `enforce` passes them. Covered by four tests, including the case where the English source changed underneath a stored verdict.

site/scripts/i18n/review-translations.ts
```diff
@@ -0,0 +1,520 @@
...
+
+/** Reviewer model. Sonnet 5 is the quality/cost sweet spot and is strong across all
+ *  ten target languages, which is where the reported weakness is (ar/tr/ko fluency). */
+const MODEL = process.env.I18N_REVIEW_MODEL ?? 'claude-sonnet-5';
+
+/** Ceiling so a runaway loop fails loudly instead of quietly running up a bill. */
+const MAX_ENTRIES_PER_RUN = Number(process.env.I18N_REVIEW_MAX_ENTRIES ?? '2000');
+
+/** Concurrent in-flight requests. Low enough to stay under standard rate limits. */
+const CONCURRENCY = Number(process.env.I18N_REVIEW_CONCURRENCY ?? '4');
```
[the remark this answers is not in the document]
mobeenabdullah: Fixed, and the concurrency path was the sharp one you called out: `NaN` builds zero runners, so a locale finished with no calls, no failures, and a state file written as if it were reviewed and clean. Both vars now go through `parsePositiveInt`, which warns and falls back rather than silently disabling the guard. Also asserted the concurrency lower bound in the test so it can no longer pass when nothing runs.

site/scripts/i18n/review-translations.ts
```diff
@@ -0,0 +1,520 @@
...
+    promptVersion: PROMPT_VERSION,
+    entries: {},
+  });
+  // A rubric change makes every stored verdict incomparable, so discard them all
+  // rather than mixing verdicts from two different rubrics in one report.
+  if (state.promptVersion !== PROMPT_VERSION) {
+    return { promptVersion: PROMPT_VERSION, entries: {} };
+  }
+  return { promptVersion: PROMPT_VERSION, entries: state.entries ?? {} };
+}
```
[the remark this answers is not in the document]
mobeenabdullah: Fixed. The file is generated but committed, so it can arrive truncated or hand-edited. `loadReviewState` now drops verdicts missing a hash or a findings array and runs the rest through shared validation, so an unknown severity cannot reach `summarize` and report NaN. Added tests for a bad rubric version, malformed verdicts, an unknown severity, and an unparseable file.

site/scripts/i18n/review-translations.ts
```diff
@@ -0,0 +1,533 @@
...
+        // A refusal or a truncated answer is not a verdict — return null so the
+        // entry stays unreviewed rather than being recorded as clean. Log why:
+        // a silent null here is indistinguishable from a transient network blip,
+        // and these two causes need opposite responses (retry vs raise the budget).
+        if (response.stop_reason === 'refusal' || response.stop_reason === 'max_tokens') {
+          console.warn(
+            `[i18n] review: ${shareId} unreviewed (stop_reason=${response.stop_reason})`
+          );
+          return null;
+        }
```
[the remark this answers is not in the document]
mobeenabdullah: Fixed. `model_context_window_exceeded` now joins refusal and max_tokens in the incomplete set, so a truncated body never reaches `JSON.parse` and the entry stays unreviewed for the next run rather than being recorded as clean.

site/scripts/i18n/preserve-protection.ts
```diff
@@ -158,15 +183,25 @@ function main(): void {
...
+    // restore rewrites every file in this directory, not just English, so a
+    // collision anywhere would have genuine content rewritten into a preserve-term.
+    // Checked before writing anything, so a refusal leaves the tree untouched.
+    const contentByFile: Record<string, unknown> = {};
+    for (const name of fs.readdirSync(CONTENT_DIR).filter((f) => f.endsWith('.json'))) {
+      contentByFile[name] =
+        name === 'en.json'
+          ? source
+          : JSON.parse(fs.readFileSync(path.join(CONTENT_DIR, name), 'utf-8'));
+    }
```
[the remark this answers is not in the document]
mobeenabdullah: Fixed. The parse is now per file with the name in the error, since "Unexpected token" is little help when eleven locale files could be the culprit.

site/tests/unit/i18n-review-translations.test.ts
```diff
@@ -0,0 +1,363 @@
...
+  entryHash,
+  parseFindings,
+  pruneOrphanedVerdicts,
+  reviewViolations,
+  reviewLocale,
+  selectEntriesForReview,
+  summarize,
+  type Finding,
+  type ReviewState,
+} from '../../scripts/i18n/review-translations';
```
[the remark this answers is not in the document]
mobeenabdullah: Added. Five tests: absent file, mismatched rubric version, malformed verdicts, an unknown severity reaching `summarize`, and a truncated file.

site/tests/unit/i18n-review-translations.test.ts
```diff
@@ -0,0 +1,363 @@
...
+        inFlight += 1;
+        peak = Math.max(peak, inFlight);
+        await new Promise((r) => setTimeout(r, 5));
+        inFlight -= 1;
+        return [];
+      },
+      3
+    );
+    expect(peak).toBeLessThanOrEqual(3);
+  });
```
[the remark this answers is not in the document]
mobeenabdullah: Good catch — that test passed when the pool ran nothing at all, which is exactly the failure mode of the concurrency bug above. Now asserts `peak > 1` and that every entry was actually visited.

site/scripts/i18n/sync-glossary.ts
```diff
@@ -75,6 +75,22 @@ export const PRESERVE_TERMS: string[] = [
...
+  // "Qwen-Image" protected the head of "Qwen-Image-Layered" and left "Layered"
+  // to be translated.
+  //
+  // "Fun" is deliberately listed bare rather than as each variant
+  // (Fun Inpaint / Fun Inp / Fun Control / Fun Camera Control / Fun Union):
+  // every one of its 38 occurrences in the English content is a model name and
+  // none is the ordinary adjective, and matching is case-sensitive, so lowercase
+  // "fun" in prose is untouched.
+  'Fun',
+  'Qwen-Image-Layered',
```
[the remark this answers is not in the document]
mobeenabdullah: You were right, and this was the most serious one. Verified it: with a bare term, `WanFunInpaintToVideo` became `{{PT}}{{PT}}InpaintToVideo` and `Functionality` became `{{PT}}ctionality` — 17 occurrences in the current content. Terms are now guarded against matching inside a longer run of letters. Deliberately guarded on letters rather than a plain word boundary: `\b` would have stopped `Wan` matching in `Wan2.1` and left the whole family name unprotected. Tests cover both directions.

site/i18n/glossary/preserve-terms.json
mobeenabdullah: Fair question, and I think the honest answer is: a static list is the wrong long-term shape, but the right thing for this PR.

Why it works today: the list is the contract the deterministic validator enforces. Because a term is on it, the validator can prune a field where the term went missing, with no model and no cost. That is what caught 68 fields across the ten languages here.

Why it will not scale: it only covers names someone remembered. This PR exists because "Fun" was not on it. There are ~234 proper-noun-shaped tokens in the English content matching nothing on the list.

The durable version is to derive it from `templates/index.json`, where model and node names already live as structured data, instead of hand-maintaining a parallel copy. I would rather do that as its own PR than widen this one — happy to open it.

mobeenabdullah: That is the natural next step, and I would suggest one guardrail on it.

The reviewer already reports a `terminology` category, so the missing-term signal is there. What I would not do is let it write to `preserve-terms.json` unattended: that file drives pruning, so a wrong entry silently suppresses legitimate translation across all ten languages, and nothing downstream would question it. In this PR I nearly made that mistake myself — my first pass listed `Fun Inpaint`, `Fun Control` and so on, and simulating it beforehand showed it still missed `Fun Inp` and `Fun Camera Control`. A model proposing terms would hit the same trap.

So: have it propose, not commit. Concretely, collect `terminology` findings whose span is an untracked proper noun and open a PR (or comment) listing the candidates with the strings that triggered them, for a human to accept. That keeps the audit trail and keeps a bad term from shipping silently. Happy to add it as a follow-up once this lands.

mobeenabdullah:
@coderabbitai review

All ten inline findings are addressed and pushed. Two were real bugs worth calling out:

- **Bare `Fun` matching inside identifiers** — verified: `WanFunInpaintToVideo` became `{{PT}}{{PT}}InpaintToVideo` and `Functionality` became `{{PT}}ctionality`, 17 occurrences in the current content. Terms are now guarded against matching inside a longer run of letters, guarded on letters rather than `\b` so `Wan` still matches in `Wan2.1`.
- **Stale verdicts pruning fresh translations** — enforcement is not gated on the review step, so a verdict could outlive the text it described. It now skips any verdict whose hash no longer matches the current source and translation.

Also validated the numeric env vars (a NaN concurrency built zero runners and wrote a locale as reviewed-and-clean), validated stored verdicts on load, handled context-window truncation as unreviewed, and named the file on a locale parse failure. 463 tests, lint and typecheck green.

site/tests/unit/i18n-review-translations.test.ts
```diff
@@ -0,0 +1,535 @@
...
+describe('loadReviewState (reads a committed, therefore corruptible, file)', () => {
+  const dir = path.join(process.cwd(), 'src', 'i18n', 'review');
+  const file = path.join(dir, 'zh.json');
+  const write = (data: unknown) => {
+    fs.mkdirSync(dir, { recursive: true });
+    fs.writeFileSync(file, JSON.stringify(data));
+  };
+  afterEach(() =>
+    fs.rmSync(path.join(process.cwd(), 'src', 'i18n', 'review'), { recursive: true, force: true })
+  );
```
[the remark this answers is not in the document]
mobeenabdullah: Fixed, and thank you for catching this one — it was the worst bug in the PR. Unit tests run with `site/` as the working directory, so those fixtures were writing to the same `src/i18n/review` the pipeline commits, and `afterEach` removed the tree. Once review state is committed, running the suite would have deleted it.

The state directory is now injectable and the tests use `fs.mkdtempSync`. Added a case asserting the default still resolves under `src/i18n/review`, so the injection cannot silently redirect production. Verified by leaving a canary file in the real directory and running the suite: it survives.

mobeenabdullah:
@coderabbitai review

site/tests/unit/i18n-review-translations.test.ts
```diff
@@ -0,0 +1,541 @@
...
+
+  it('survives a truncated / unparseable file', () => {
+    fs.writeFileSync(path.join(dir, 'zh.json'), '{"promptVersion": 1, "entries": {');
+    expect(loadReviewState('zh', dir).entries).toEqual({});
+  });
+
+  it('defaults to the real review directory when no dir is given', () => {
+    // Guards the injection itself: the default must still resolve under src/i18n/review.
+    expect(reviewStatePath('zh')).toContain(path.join('src', 'i18n', 'review'));
+    expect(reviewStatePath('zh', dir)).toBe(path.join(dir, 'zh.json'));
```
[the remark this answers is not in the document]
mobeenabdullah: Fixed. Now asserts equality with `path.join(process.cwd(), 'src', 'i18n', 'review', 'zh.json')` — you are right that a substring check would have accepted `review-backup`, which is precisely the redirect this assertion exists to catch.

mobeenabdullah:
@coderabbitai review

site/scripts/i18n/review-translations.ts
```diff
@@ -0,0 +1,606 @@
...
+/**
+ * Bumping this invalidates every stored verdict, forcing a full re-review. It is a
+ * deliberate, costly action: change it only when the rubric itself changes, since
+ * verdicts produced under an older rubric are no longer comparable.
+ */
+export const PROMPT_VERSION = 1;
+
+/** Reviewer model. Sonnet 5 is the quality/cost sweet spot and is strong across all
+ *  ten target languages, which is where the reported weakness is (ar/tr/ko fluency). */
+const MODEL = process.env.I18N_REVIEW_MODEL ?? 'claude-sonnet-5';
```
christian-byrne, line 47: issue: `claude-sonnet-5` is not a valid model ID and will cause every API call to fail with a 4xx at runtime.

Valid IDs as of SDK 0.115.0: `claude-sonnet-4-5`, `claude-sonnet-4-6`, etc. The `Model` type accepts `(string & {})` so TypeScript won't catch this -- but the API will reject every request. Because the step is `continue-on-error: true`, CI stays green while no translation is ever actually reviewed.

Suggest: replace with `claude-sonnet-4-6` (or whatever the intended default is).

site/scripts/i18n/review-translations.ts
```diff
@@ -0,0 +1,606 @@
...
+          },
+          messages: [{ role: 'user', content: buildUserPrompt(shareId, en, target) }],
+        });
+        // A refusal or a truncated answer is not a verdict — return null so the
+        // entry stays unreviewed rather than being recorded as clean. Log why:
+        // a silent null here is indistinguishable from a transient network blip,
+        // and these two causes need opposite responses (retry vs raise the budget).
+        // Any stop reason that means the answer was cut short or withheld: the
+        // body is not a complete verdict, so it must not reach JSON.parse.
+        const incomplete = ['refusal', 'max_tokens', 'model_context_window_exceeded'];
```
christian-byrne, line 564: nitpick (non-blocking): `'model_context_window_exceeded'` is not in the `StopReason` enum (`end_turn | max_tokens | stop_sequence | tool_use | pause_turn | refusal`). Context window overflow arrives as a thrown exception (4xx), not a stop_reason, so it is already handled by the `catch` block in `reviewLocale`. This guard is dead code.

Consider removing it and adding a comment in the catch block noting that context overflow arrives as an API error.

site/scripts/i18n/review-translations.ts
```diff
@@ -0,0 +1,606 @@
...
+  locale: Locale,
+  state: ReviewState,
+  /**
+   * Current content. A verdict describes one exact (source, translation) pair, so
+   * one whose hash no longer matches is about text that has since been replaced —
+   * acting on it would prune a translation that may well have fixed the very
+   * problem the verdict describes. Omitting these is only safe when the caller has
+   * already established freshness; enforcement passes them.
+   */
+  english?: Record<string, WorkflowContent>,
```
christian-byrne: nitpick (non-blocking): `english` and `localeContent` are optional with a JSDoc note that omitting them means "the caller vouches for freshness." No call site currently omits them -- the optionality exists for no concrete use case and is a footgun for a future caller who skips the docs.

Consider making them required, or splitting into `reviewViolationsChecked` / `reviewViolationsUnchecked` if the no-arg form is genuinely needed.

christian-byrne (commented):
Reviewed with parallel subagents covering logic/correctness, SDK API surface, CI/CD pipeline, and test coverage. One blocking issue found (invalid model ID); everything else is non-blocking.

**Conflict heads-up:** PR #1060 ([DO-NOT-MERGE] by deepme987) also modifies `.github/workflows/i18n-update-hub.yml` to swap PAT → bot app token. If this PR merges first, #1060 will have a conflict on the translate step `env:` block. Worth a ping to deepme987 when this lands.

.github/workflows/i18n-update-hub.yml
```diff
@@ -125,6 +125,10 @@ jobs:
         continue-on-error: true
         env:
           OPENAI_API_KEY: ${{ secrets.OPENAI_API_KEY }}
+          # Repo variable, so the translation model can be changed (and reverted)
+          # without a code change. Unset falls back to the model we have actually
+          # run all ten locales through — see site/.i18nrc.cjs.
+          HUB_I18N_MODEL: ${{ vars.HUB_I18N_MODEL }}
```
christian-byrne, line 131: nitpick (non-blocking): HUB_I18N_MODEL is wired here for the translate step, but the AI review script reads process.env.I18N_REVIEW_MODEL -- which is never set in the workflow. Once the model ID default is corrected, the review model cannot be changed via a repo variable without a code change. Not blocking, but worth either wiring I18N_REVIEW_MODEL via vars.HUB_I18N_REVIEW_MODEL in the review step, or adding a comment that the review model is currently hardcoded.

site/tests/unit/i18n-review-translations.test.ts
```diff
@@ -0,0 +1,545 @@
...
+  it('carries findings through to prunable violations — the full path', async () => {
+    const result = await reviewLocale(two, englishTwo, empty, async (shareId) =>
+      shareId === 'wf1' ? [finding] : []
+    );
+    const violations = reviewViolations('zh', result.state);
+    expect(violations).toHaveLength(1);
+    expect(violations[0]).toMatchObject({ shareId: 'wf1', field: 'title' });
+  });
+
+  it('respects the concurrency limit', async () => {
```
christian-byrne, line 344: suggestion (non-blocking): the concurrency test checks peak >1 and <=3 with 10 items at limit=3. Two edge cases are not pinned: limit=1 (serial, peak should equal exactly 1) and 0 items (pool should return without calling the worker). The Math.max/Math.min guard in pooled() handles these today, but a regression would not be caught.

site/tests/unit/i18n-review-translations.test.ts
```diff
@@ -0,0 +1,545 @@
...
+    });
+    expect(result.failures[0]).toContain('429');
+    expect(result.state.entries.wf2).toBeDefined();
+  });
+
+  it('carries findings through to prunable violations — the full path', async () => {
+    const result = await reviewLocale(two, englishTwo, empty, async (shareId) =>
+      shareId === 'wf1' ? [finding] : []
+    );
+    const violations = reviewViolations('zh', result.state);
```
christian-byrne: suggestion (non-blocking): the stale-verdict safety property is tested on each half in isolation -- reviewViolations() drops stale verdicts, and pruneViolatingFields() accepts extraViolations. There is no test that wires both together: stale verdict -> reviewViolations() -> pruneViolatingFields(extraViolations) -> field is NOT pruned. That composed path is what runs in production, and a regression in the wiring would currently go undetected.

christian-byrne (commented):
question: the git add site/src/i18n here stages src/i18n/review/*.json (per-locale AI verdict files) alongside translated content. This is probably intentional -- the incremental-review logic reads them back on the next run via Restore partial progress. But they will appear in the translation PR diff seen by human reviewers. A short comment distinguishing review state from translation output would help.

site/scripts/i18n/review-translations.ts
mobeenabdullah: Checked this before changing it, and `claude-sonnet-5` is valid — it is in the `Model` union of the SDK version this PR pins.

From `node_modules/@anthropic-ai/sdk/resources/messages/messages.d.ts:853` on 0.115.0, the exact version you referenced:

```
export type Model = 'claude-sonnet-5' | 'claude-fable-5' | 'claude-mythos-5' | 'claude-opus-5' | 'claude-opus-4-8' | ... | (string & {});
```

It is the first member, so this is not resolving through the `(string & {})` escape hatch.

The stronger evidence is runtime rather than types: the reviewer was exercised against the live API in CI on a throwaway branch before this PR went up, and it returned real findings. That run is also what surfaced the `max_tokens` starvation bug fixed in this PR (4/5 entries reviewed and 0 pruned before, 5/5 and 3 pruned after). Every call 4xx-ing would have produced no findings at all, so the measured behaviour rules that out.

Leaving the default as is. Switching to `claude-sonnet-4-6` would be a real downgrade in the fluency judgement this reviewer exists to make, and it is overridable per-run anyway — which your workflow comment below made properly configurable.

site/scripts/i18n/review-translations.ts
mobeenabdullah: Same check here — `model_context_window_exceeded` is in the enum, at `messages.d.ts:1046` on 0.115.0:

```
export type StopReason = 'end_turn' | 'max_tokens' | 'stop_sequence' | 'tool_use' | 'pause_turn' | 'refusal' | 'model_context_window_exceeded';
```

The list in your comment is the enum minus its last member, so the guard is reachable rather than dead. Keeping it: an entry whose prompt overflows the window returns 200 with that `stop_reason`, and treating it as reviewed-and-clean is exactly the silent-skip failure mode this PR already had to fix once with `max_tokens`. The worst translations are the ones most likely to trip it, so a miss here is not evenly distributed.

site/scripts/i18n/review-translations.ts
mobeenabdullah: Agreed, and fixed in c049b57 — both are required now.

You are right that no production call site omits them; the risk was the next one. The freshness comparison is the only thing between a stored verdict and a translation that has since been rewritten, so a caller that skipped it would prune on a verdict about text that no longer exists — undoing the very fix the re-translation may have applied. Optional arguments made that a docs-reading problem, and required ones make it unrepresentable.

Dropped the "caller vouches for freshness" test with it, since the behaviour it pinned no longer exists, and added one for the case where a verdict outlives its workflow entirely (archived or renamed).

.github/workflows/i18n-update-hub.yml
mobeenabdullah: Good catch — wired in c049b57 as `I18N_REVIEW_MODEL: ${{ vars.HUB_I18N_REVIEW_MODEL }}`, mirroring `HUB_I18N_MODEL` on the translate step.

Deliberately a second variable rather than reusing the first: the translator and the reviewer should be independently swappable, since changing the model that writes the translations should not silently change the one that judges them. Unset still falls back to the in-code default, so nothing changes until someone sets it.

site/tests/unit/i18n-review-translations.test.ts
mobeenabdullah: Added both in c049b57.

`limit=1` asserts `peak` is exactly 1 over 5 entries — the existing `>1 and <=3` assertion cannot catch a pool that ignores the limit when the limit is 1, since 1 is below its lower bound. The 0-item case asserts the worker is never called and the pool returns rather than waiting on an entry that will never arrive.

site/tests/unit/i18n-review-translations.test.ts
mobeenabdullah: This was the most valuable of the six — added in c049b57, in the enforce suite where both halves are importable.

Two cases, since the wiring can fail in either direction: a verdict matching the current translation must prune the field, and a verdict written before a re-translation must not. Both go through the real `reviewViolations() -> pruneViolatingFields(extraViolations)` path rather than asserting on either half alone.

I checked the tests actually fail rather than trivially passing: disabling the freshness comparison in `reviewViolations` turns 4 tests red across the two suites, including this pair. Before that, as you say, a regression in the wiring would have left both suites green.

mobeenabdullah:
On the question about `git add site/src/i18n` staging the review verdicts alongside the translated content: correct on every point, and the suggestion is a good one. Added the comment in `dbdbdeb`.

**Yes, it is intentional, and yes it is load-bearing.** The restore step at the top of the job does:

```
git checkout "origin/$STABLE_BRANCH" -- site/src/i18n
```

That is the same subtree, so the verdicts come back with the translations. Without them every entry looks unreviewed on the next run and the whole corpus is re-reviewed, which is the cost control this PR is built around.

**Why they are not split into separate paths.** A verdict is keyed to a hash of the exact English and translated text it describes, so the two are only meaningful together. Splitting them gives two ways to get it wrong:

- translations restored without verdicts: every hash misses, the full corpus is re-reviewed and re-paid for
- verdicts restored without translations: findings point at text that no longer exists, and enforcement would prune on them

Keeping them in one subtree makes that restore atomic by construction instead of something the next person editing the workflow has to remember. That is now written down at the `git add` rather than left implicit.

**On the diff noise, which is the real concern.** The verdict files genuinely will show up in the automation PR. Worth knowing what a reviewer is looking at:

- `content/` and `locales/` are the translation output, the thing being judged
- `review/` is machine state, one JSON per locale, and it churns whenever a translation changes because the hash changes with it

So a large `review/` diff is expected on a run that retranslated a lot, and it carries no information a human needs to read line by line. The summary counts in the run log are the useful view of it.

If the churn turns out to bother reviewers in practice, the cheap follow-up is a `.gitattributes` entry marking `site/src/i18n/review/*.json` as `linguist-generated=true`, which collapses it by default in the GitHub diff while keeping the file committed and restorable. Happy to add that now if you would prefer it pre-emptively, though I would rather see one real run first than guess at the volume.

Comment-only change, YAML re-parsed, no behaviour change.

dante01yoon (approved):
The AI review remains a best-effort quality layer while deterministic validation and pruning safeguards stay authoritative. Hash invalidation, retry behavior, prune limits, sentinel protection, and focused tests all look consistent on the current head. I found no blocking issues.

PR #1100, opened 2026-08-07
fix(site): scope hub filter counts to the active tab, and stop free claims on paid workflows

mobeenabdullah:
Two hub bugs found while testing the filter dropdown. Both are user-facing, both are verified against the live catalogue.

## 1. Filter counts ignored the active tab

The Filter popover counted the whole catalogue no matter which tab was selected, so on **Comfy Apps** it advertised totals that tab could never return.

Measured on production before the fix:

| | ALL | Comfy Apps |
| --- | --- | --- |
| Templates in the tab | 597 | 14 |
| Model options offered | 119 | **119, identical** |
| Highest count shown | 40 | **40, identical** |

Selecting **"Wan 37"** on Comfy Apps rendered *"No templates match your filters. Showing 0 of 0"*. It was not one bad option: **99 of 119 model options and 36 of 75 category options had zero apps behind them**, so most of the filter was dead on that tab. Categories were affected the same way (`API`, rendered as "Partner Nodes", showed 209 against 5 actual apps).

**Cause.** `HubBrowse` passes the unfiltered catalogue as `facet-templates`, deliberately, so counts stay stable while badge filters are applied. That intent is fine for badges. The tab got swept into the same exemption: `activeTab` is a separate ref, and the tab filter was applied in `WorkflowGrid` after and separately from the facet source, so the counts never saw it.

**Fix.** A tab is a scope, not a refinement. `templatesInTab()` is now the single predicate that both the grid and the facet source read, so they cannot drift. Counts still ignore badges, unchanged. Values with no templates in the tab drop out of the list entirely, so a dead option can no longer be selected.

After the fix:

| Tab | Options | Max count | Tab total |
| --- | --- | --- | --- |
| ALL | 75 | 209 | 597 (unchanged) |
| Node Graphs | 74 | 206 | 583 |
| Comfy Apps | 18 | 5 | 14 |

209 on ALL = 206 Node Graphs + 3 Comfy Apps.

**Not a bug:** the filter *results* were always correct. All 7 results from a Nano Banana Pro selection genuinely carry that model. Only the advertised count was wrong.

## 2. Workflows that charge credits advertised a free run

The hero CTA chose its label from the `API` tag alone:

```ts
const isApiTemplate = data.tags?.includes('API') ?? false;
```

That tag is applied by hand at publish time, so **106 approved workflows that call OpenAI, Gemini, ByteDance, Topaz and others carried no tag and rendered "Try free on Comfy Cloud"** for a run that costs credits.

Reproduces on production today: [Storyboard To Video](https://www.comfy.org/workflows/f4e29143100c-f4e29143100c/) has no tags at all, calls `OpenAIGPTImageNodeV2`, `GeminiNode` and `ByteDance2ReferenceNode`, and says "Try free on Comfy Cloud". [Topaz Image Enhance](https://www.comfy.org/workflows/1c0a3a9faad3-1c0a3a9faad3/), which is tagged, correctly says "Try on Comfy Cloud".

**Fix.** `isBillableWorkflow()` accepts either signal: the curated tag, or a partner node found in the workflow itself. The node scan needs one request per workflow, so it runs in `refresh-partner-nodes-snapshot.ts` and the result is committed, following the existing `refresh-feature-flags-snapshot.ts` pattern.

Subgraphs are walked as well as the root graph. A subgraph appears in the node list as a bare UUID with its real nodes under `definitions.subgraphs[]`, so reading only the top level missed 7 workflows.

### Known gap, stated plainly

Detection is limited by `scripts/data/mcp/api_node_model_options.json`, which holds **93 node classes and is stale** (its `source` is a path on someone's laptop). `MiniMaxH3ImageToVideo` is absent from it, and also from the cloud repo's 214-entry `nodeclassifier` list, so **MiniMax H3 still reads as free**. Regenerating that file with `scripts/mcp/scan_api_nodes.py` against a current ComfyUI closes most of the gap. The durable fix is the hub deriving the tag at publish time, where the classifier already lives.

A vendor-name prefix heuristic would have caught MiniMax H3, but it also flagged `ReverseImageBatch`, a core utility node, so it is not here. Detection fails open on unknown node types, which is why the tag remains the other half of the signal.

## Verification

- **All 597 workflow pages built and checked: 597 correct CTAs, 0 wrong.**
- Filter counts verified per tab in the browser, including the invariant that no count exceeds its tab total.
- 415 unit tests pass, including 8 new for `templatesInTab` and facet scoping, and 9 for the partner-node lookup.
- `pnpm run lint` clean, `astro check` 0 errors 0 warnings.

## Note on `hub-api.ts`

`scripts/` runs that module under plain Node, where `import.meta.env` is undefined and a direct read throws. It now reads env through `ranking`'s existing `readEnv` helper, which already handles both contexts.


site/src/lib/partner-nodes.ts
```diff
@@ -0,0 +1,43 @@
...
+ * Whether a workflow should be presented as billable rather than free.
+ *
+ * Either signal is enough: the curated `API` tag, or a partner node found in the
+ * workflow itself. They cover different gaps, so both are checked.
+ */
+export function isBillableWorkflow(
+  tags: string[] | undefined | null,
+  shareId: string | undefined | null
+): boolean {
+  return (tags?.includes('API') ?? false) || usesPartnerNodes(shareId);
```
dante01yoon: issue: Absence from this positive-only snapshot is treated as proof that a workflow is free, but it also means the workflow was published after the snapshot. This already fails on the current preview: the snapshot was generated at 15:42, while approved workflow `0309de53eb52` now calls `ByteDanceSeedreamNodeV2`, has no `API` tag, is absent here, and renders **Try free on Comfy Cloud**. I scanned all 598 currently approved workflows against the same 93-node source and this is the one untagged paid workflow missing from the committed snapshot. Could the snapshot also record every scanned share ID and make unscanned/unknown IDs fail closed (no free claim), with a regression for a post-snapshot ID? Regenerating only the positive list will become stale again on the next publication.

site/src/components/hub/WorkflowGrid.vue
```diff
@@ -106,21 +107,23 @@ watch(
...
-
 const store = useHubStore();
 const displayCount = ref(30);
 
+// Counts deliberately ignore the badge selection so they stay stable while a
+// filter is applied, but they must still honour the tab. Counting the whole
+// catalogue here advertised totals the tab could never return: on Comfy Apps,
+// "Wan 37" produced an empty grid because none of those 37 are apps.
+const facetSource = computed(() =>
+  templatesInTab(props.facetTemplates ?? workingSet.value, store.activeTab.value)
```
dante01yoon, line 118: issue: Scoping the available facet values does not reconcile an already-active badge. On the preview, **All → Filter → Wan → Comfy Apps** leaves `Wan` active, hides it from this scoped facet list, and still renders `No templates match your filters / Showing 0 of 0`. `HubBrowse` applies the retained badge before this tab predicate, and `setTab` never clears or prunes it. Could the tab transition clear incompatible badges (or clear all badges as the smallest predictable behavior) and add this exact transition as a regression test?

dante01yoon (changes requested):
I found two blocking paths that still reproduce the user-facing failures this PR intends to remove: the positive-only billing snapshot fails open as soon as a workflow is published, and tab changes retain filters that the newly scoped facet list hides. I verified both against the current preview/live catalogue; focused unit tests pass but do not cover these transitions.

site/src/lib/partner-nodes.ts
mobeenabdullah: Fixed, and you were right that regenerating the positive list alone would just go stale again.

The snapshot now records `scannedShareIds` next to `shareIds`, and a workflow earns the free claim only by having been scanned *and* come back clean. Anything the run never saw is treated as billable, so a publish between refreshes under-promises rather than promising a run it cannot honour.

Your example confirms it. `0309de53eb52` is now scanned and flagged. The catalogue also went 597 to 598 between your review and this push, which is exactly the drift you described.

Regression tests cover an unknown id being billable, a missing id being billable, and a scanned-and-clean id staying free, plus an invariant that every flagged id was actually scanned.

site/src/components/hub/WorkflowGrid.vue
mobeenabdullah: Good catch, that transition was broken and I had made it worse: before this PR the option at least stayed visible so you could untick it.

Fixed by reconciling the badges rather than clearing them all. `badgesAvailableIn` prunes badges the new tab cannot satisfy and keeps the ones that still match, so switching from All to Comfy Apps drops `Wan` but keeps something like `Image Edit` that exists in both. Clearing everything was the other option you offered, but it throws away filters that are still meaningful.

It lives in `lib/hub-tabs.ts` rather than inside the component so it is unit testable, and your exact sequence is now a test: `All, filter Wan, Comfy Apps` leaves no badge behind. Also covered: a badge that survives, tag badges, and an unrecognised badge type being kept rather than silently dropped.

site/src/lib/hub-tabs.ts
```diff
@@ -0,0 +1,61 @@
...
+
+  const available: Record<string, Set<string>> = {
+    model: new Set(scopedTemplates.flatMap((t) => t.models ?? [])),
+    tag: new Set(scopedTemplates.flatMap((t) => t.tags ?? [])),
+  };
+
+  return badges.filter((b) => {
+    const values = available[b.type];
+    return values ? values.has(b.value) : true;
+  });
```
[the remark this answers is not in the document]
mobeenabdullah: Good catch, and it was worse than a style point: `available["constructor"]` returns `Object.prototype.constructor`, a function, and `.has()` on it throws. So an unknown badge type would crash rather than be kept, which is the opposite of what the doc comment promises.

Switched to a `Map`. Regression test covers `constructor`, `toString`, `hasOwnProperty` and `__proto__`, asserting both that it does not throw and that the badge survives.

mobeenabdullah:
@dante01yoon ready for another look. Both findings are fixed and CI is green.

**Fail-open snapshot** (`3c35cdf`): the snapshot now records `scannedShareIds` alongside the positive list, and `isBillableWorkflow` returns true for any share id the scan never inspected. Absence is no longer read as proof a workflow is free, so a workflow published after the snapshot is treated as billable rather than free. Your example `0309de53eb52` now reads as billable.

**Unreconciled badge on tab change** (`3c35cdf`): `badgesAvailableIn` prunes any active badge the new tab cannot satisfy, applied from a watcher in `WorkflowGrid`. All, filter Wan, then Comfy Apps now clears `Wan` instead of leaving it active over an empty grid.

Also took CodeRabbit's follow-up (`36d9f28`) to key the badge-type lookup off a Map rather than an object literal, so a badge type of `constructor` or `toString` resolves to nothing instead of hitting `Object.prototype`.

Reassigning to you: the three threads are resolved and the only thing outstanding is the changes-requested from before these commits landed.

site/scripts/refresh-partner-nodes-snapshot.ts
```diff
@@ -0,0 +1,116 @@
...
+let cursor = 0;
+
+// A read failure aborts the whole run. Skipping a workflow would silently
+// restore the free claim on one that charges credits, which is the bug this
+// snapshot exists to prevent.
+async function worker(): Promise<void> {
+  while (cursor < shareIds.length) {
+    const shareId = shareIds[cursor++];
+    const detail = await getWorkflow(shareId);
+    scanned.push(shareId);
```
dante01yoon: issue: `scannedShareIds` currently means “the workflow was fetched,” not “every node type was authoritatively classified.” Because `api_node_model_options.json` is explicitly known to be stale, a fetched workflow containing a newly shipped paid node is added to `scanned` but not `withPartnerNodes`, so `isBillableWorkflow()` treats it as free. This is visible now for the MiniMax H3 Text-to-Video, Reference-to-Video, and Image-to-Video pages, which show “Try free on Comfy Cloud” despite using paid nodes. Could billing eligibility come from the hub publish/index contract (or otherwise fail closed for unrecognized node types) rather than allowing this frontend snapshot to certify those workflows as free?

dante01yoon (changes requested):
The tab-scoped count fixes and the previous review findings are resolved. One blocking billing-classification issue remains: the snapshot still certifies workflows as free from a known-incomplete node catalogue, and current preview pages demonstrate false free claims.

site/scripts/refresh-partner-nodes-snapshot.ts
mobeenabdullah: Confirmed, and the mechanism is narrower and worse than "the snapshot is stale". Traced all three pages you named.

**All three genuinely call a paid node, and all three are certified free today.**

| template | paid node | where |
| --- | --- | --- |
| `video_minimax_h3_r2v` | `MiniMaxH3ReferenceToVideo` | root graph |
| `video_minimax_h3_t2v` | `MiniMaxH3ImageToVideo` | inside a subgraph |
| `video_minimax_h3_i2v` | `MiniMaxH3ImageToVideo` | inside a subgraph |

None carries the `API` tag, so the tag is not saving them either. Their `api_*` siblings do carry it, which is why only these three show the free claim.

**The root cause is a node rename upstream, not just age.** `api_node_model_options.json` holds the old names:

```
MinimaxTextToVideoNode, MinimaxImageToVideoNode, MinimaxSubjectToVideoNode,
MinimaxHailuo03TextToVideoNode, MinimaxHailuo03FirstLastFrameNode, MinimaxHailuo03ReferenceNode
```

The workflows use `MiniMaxH3ReferenceToVideo` and `MiniMaxH3ImageToVideo`. Same nodes, renamed. So the scan finds nothing it recognises, the workflow lands in `scannedShareIds` and not in `shareIds`, and `isBillableWorkflow` reads that as proof of freeness.

Worth saying explicitly: the subgraph walking is working. Two of the three hide the paid node inside `definitions.subgraphs[]` and the scan does reach it. It reaches it and does not recognise the name.

**So your framing is right.** `scannedShareIds` means "we fetched this and found nothing we know about", and the code treats that as "nothing paid is in it". Those are only the same statement while the node list is complete, and it silently stopped being complete.

**On the fix, I do not think I should pick unilaterally, because the honest version is a product change.** The snapshot currently scans 598 workflows and flags 220. **378 workflows claim a free run on the strength of "we recognised nothing paid".** Failing closed on unrecognised node types, done properly, removes the free claim from all 378 until something can vouch for them, because the list only enumerates paid nodes and there is no corresponding list of known-safe ones to check against.

Three options as I see them:

1. **Hub contract, your first suggestion.** Correct and durable: cloud already decides what to bill, so eligibility should come from the publish/index contract rather than a frontend snapshot inferring it. Cloud PR Comfy-Org/cloud#6399 is already adding a derived field to that contract for App Mode, so the shape exists. This is the one I would build toward.
2. **Regenerate the node list and add a staleness guard.** Fixes these three today and re-runs the scan against current names. Does not fix the class: the next rename reopens it silently unless the refresh also fails loudly when it finds a workflow node type it cannot classify at all.
3. **Fail closed now.** Safest, and drops the free claim from 378 workflows. Under-promising is the right direction for a billing claim, but that is a call for whoever owns the pricing message rather than mine to make in a filter-counts PR.

My recommendation is 2 now and 1 next, with the refresh script recording every unrecognised node type it encounters so staleness surfaces as a diff instead of as a wrong claim. I can do that here if you agree. I did not want to change what 378 pages say about money without you and Mobeen agreeing on the direction first.

One caveat on regenerating: `api_node_model_options.json` is produced by scanning `comfy_api_nodes/` in a ComfyUI checkout, which I do not have here, so option 2 needs that checkout to be done properly rather than by hand-adding the three names I happen to have found.

mobeenabdullah: Went and tested the two fixes that would keep this in the frontend. Both are dead ends, and I think that settles it in favour of your first suggestion.

**1. The workflow JSON does not identify API nodes.** I compared a paid node against core nodes in the same file:

```
MiniMaxH3ReferenceToVideo   properties = ["Node name for S&R"]
KSamplerSelect              properties = ["Node name for S&R"]
VAEDecode                   properties = ["Node name for S&R"]
```

No flag, no `cnr_id`, nothing. A paid node and a core node are indistinguishable in the graph, so the name list is the only frontend signal available.

**2. A vendor-name heuristic cannot separate paid from core.** I built one and ran it over all 1083 node types in the catalog. Requiring a token boundary does keep `ReverseImageBatch` out, which was the objection last time. But it flags **88 types, and the false positives are structural, not tunable**:

```
LTXVAddGuide, LTXVScheduler, LTXVConditioning, LTXVCropGuides ...   (21 core LTX nodes)
Flux2Scheduler                                                      (core)
MiniMaxH3ImageToVideo, KlingAvatarNode, RecraftTextToImageNode ...  (genuinely paid)
```

`Ltxv` and `Flux2` are vendor prefixes shared by core nodes for open models and by paid API nodes. No amount of tuning separates "LTXV the open model's core nodes" from "LTXV the API node", because the names carry no such distinction.

**So the frontend cannot classify this. It has to be told.** Which is your first option, and it is the one I would build.

**One thing worth being explicit about, because it changes what this PR is.** On `main` the free claim is decided by a single line:

```ts
const isApiTemplate = data.tags?.includes('API') ?? false;
```

The three H3 pages carry no `API` tag, so **they claim a free run on `main` today**. This PR does not introduce that, and does not regress it. What it does is take the count of correctly-billable workflows from "whatever is hand-tagged" to 220, and leave a residual that the tag-only baseline also had.

That reframes the decision. Failing closed now would drop the free claim from 378 workflows, most of which are genuinely free, to close a gap that predates this PR. I do not think that trade is mine to make in a filter-counts PR, and I would rather not make it quietly.

**What I would suggest instead**, in order:

1. **Merge this as the improvement it is**, with the residual tracked rather than implied to be fixed. Happy to reword the PR description so it does not read as a complete solution, since right now it arguably does.
2. **Move eligibility to the hub contract.** Cloud already decides what to bill, and `Comfy-Org/cloud#6399` is adding a derived field to that contract right now, so the shape exists.
3. **A structural improvement available in the meantime, if you want one here:** have the snapshot record the node types per workflow, not just a yes/no. That splits the expensive half (which workflow uses which nodes, needs a catalog pass) from the cheap half (which nodes are paid, just a list). Regenerating `api_node_model_options.json` would then correct these three pages immediately without re-fetching 598 workflows, and the next rename becomes a list refresh rather than a silent wrong claim.

Tell me which of those you want and I will do it. If you would rather fail closed despite the 378, say so and I will, but I would want Mobeen to agree the messaging change first.

mobeenabdullah: Fixed in `be0cd59`, and the diagnosis moved twice on the way, so both corrections are worth stating plainly.

**Correction 1: the three MiniMax H3 pages you named are not wrong.** I confirmed them earlier and I should not have. `MiniMaxH3ImageToVideo` and `MiniMaxH3ReferenceToVideo` are **core** nodes in `comfy_extras/nodes_minimax_h3.py`, category `model/conditioning/minimax`, taking `clip` and `vae` and returning conditioning plus a latent. The three `video_minimax_h3_*` templates load local weights (`minimax_h3_video_vae_fp16.safetensors`, `minimax_h3_fl2va_pruned_int8_convrot.safetensors`) and sample locally through `UNETLoader`, `CLIPLoader`, `SamplerCustomAdvanced`. The paid MiniMax family is `MinimaxHailuo03*` in `comfy_api_nodes/`, it is already catalogued, and the `api_minimax_h3_*` templates that use it are correctly billed. Two families, near-identical names, no rename.

**Correction 2: your underlying point was right, and worse than staleness.** `api_node_model_options.json` is a model-dropdown index, not a paid-node registry. `scan_api_node_models.py` ends every node it finds no dropdown for (`if not model_options: continue`), so it held **93 of 234** API nodes by construction. Regenerating it could never have fixed that, which is why the stale-snapshot framing did not lead anywhere.

**What it was costing.** I audited every workflow the site presents as free against the full API node set. **Eight approved workflows advertised a free run while calling a paid node**, none of them carrying the `API` tag:

| Workflow | Paid node |
| --- | --- |
| FLUX Virtual Try-On | `FluxVTONode` |
| Beeble SwitchX: Video Edit | `BeebleSwitchXVideoEdit` |
| Video Aspect Ratio Extension | `KlingOmniProEditVideoNode` |
| Video to Seamless Loop Converter | `OpenAIChatNode` |
| Bria: Remove Video Background | `BriaRemoveVideoBackground` |
| Bria: Remove Image Background | `BriaRemoveImageBackground` |
| Quiver: Image to SVG | `QuiverImageToSVGNode` |
| Quiver: Text to SVG | `QuiverTextToSVGNode` |

So your instinct that this should not happen was correct, just not via the examples.

**The fix.** The scan now collects every node id declared in `comfy_api_nodes`, covering both the `IO.Schema` and `NODE_CLASS_MAPPINGS` styles, into `api_node_ids.json`, and the snapshot reads that. Living in that directory is what makes a node billable, so the list is complete by construction rather than by upkeep. Flagged workflows go 220 to 318; 79 of the newly flagged were already billable through the `API` tag, 8 are the corrections above, and **no workflow loses a correct free claim**.

On your `price_badge` point: this treats any API node as billable, which matches Christian's read that API nodes do not run for free-tier users regardless of price. If a genuinely free API node exists, this over-marks it as billable, which is the safe direction to be wrong in.

Agreed the durable answer is the server deciding this at publish time rather than the site inferring it. This is the stopgap, not a claim to have solved it.

scripts/mcp/lib/scan_api_node_models.py
```diff
@@ -18,6 +18,8 @@
 )
 FOR_DICT_LOOP_RE = re.compile(r"""for\s+\w+\s+in\s+(\w+)\s*:""")
 DYNAMIC_OPTION_RE = re.compile(r"""IO\.DynamicCombo\.Option\(\s*["']([^"']+)["']""")
+NODE_MAPPINGS_RE = re.compile(r"""NODE_CLASS_MAPPINGS\s*[:=][^{]*\{(.*?)\}""", re.S)
+MAPPING_KEY_RE = re.compile(r"""["']([A-Za-z0-9_]+)["']\s*:""")
```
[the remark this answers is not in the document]
mobeenabdullah: Real risk, fixed in `8c7d37e`. The failure mode is worse than an empty file: a wrong `COMFYUI_REPO_PATH` scans cleanly and finds nothing, so it looks like a successful run, and nothing downstream can tell "no API nodes exist" apart from "the checkout was wrong". Every workflow would then read as free.

Guarded both writes rather than only the new one, since the same scan feeds `api_node_model_options.json` and it had no guard either. The scan now exits non-zero before touching either file.

Verified both directions: an empty `comfy_api_nodes` directory exits 1 and leaves both files untouched in `git status`, and a real checkout still writes 234 ids.

I did not take the temporary-file write. A partial write leaves invalid JSON, which fails loudly at parse time, whereas the empty registry was dangerous precisely because it stays valid and plausible. The guard covers the silent case, which was the one worth closing.

site/scripts/refresh-partner-nodes-snapshot.ts
```diff
@@ -0,0 +1,122 @@
...
```
dante01yoon: issue: `scannedShareIds` still assumes a share ID identifies an immutable graph, but the backend preserves `share_id` across re-publishes and repoints it to a new `published_workflow_version` (`cloud/database/schema/published_workflow.go`; `cloud/common/hub/repository.go`'s `CreateHubPublish` upsert). That means a workflow scanned as free can later be re-published with a paid node under the same share ID; it remains in `scannedShareIds` and absent from `shareIds`, so `isBillableWorkflow()` continues to advertise a free run until this committed snapshot is manually refreshed. The full API-node registry fixes classifier coverage, but not this backend state transition. Could billability come from the Hub publish/index contract, or could the snapshot be keyed to an immutable backend version/content hash and fail closed when the current version differs?

dante01yoon (changes requested):
The new registry now covers the full current `comfy_api_nodes` set, and the prior tab/filter findings remain resolved. One blocking billing-classification state transition remains: a Hub share ID is stable across re-publishes, so `scannedShareIds` can certify a changed graph as free.

site/scripts/refresh-partner-nodes-snapshot.ts
mobeenabdullah: Confirmed in the backend code, and fixed in `6be89b9`.

Verified rather than taken on trust: `CreateHubPublish` upserts the bridge row with `OnConflictColumns(publishedworkflow.FieldWorkflowID, publishedworkflow.FieldType)` and then `UpdatePublishedWorkflowVersionID()`. It never reassigns `share_id` on the conflict path, so the id survives and is repointed at the new `published_workflow_version`. Your description of the transition is exactly right, and my snapshot was treating a share id as a stable handle on a graph, which it is not.

**The fix: a scan is only evidence about the version it scanned.** The snapshot now records the publish date each workflow had when it was scanned, and `wasScannedForPartnerNodes` requires that date to still match. `scannedShareIds: string[]` becomes `scannedAt: Record<shareId, date>`. A re-published workflow no longer matches, so it falls through to billable exactly like one the snapshot never saw.

No extra requests: the index already exposes that date as the version's create date (`e.Date = row.PublishedVersion.CreateTime` in `rowsToTemplateEntries`), so a re-publish moves it by construction.

**Residual, stated rather than hidden:** `date` is day-granular, so a re-publish later on the same day as a refresh is still invisible to this. I could not find anything finer in `HubWorkflowTemplateEntry` (no version id, no updated timestamp), so closing that last gap needs the contract, not the site.

Which is your underlying point, and I agree with it. This is a stopgap that now fails safe on the transition instead of ignoring it. The durable version is the one you and Christian landed on in Slack: derive billability at publish time and expose it on the workflow data model including the list response. That is scoped on GTM-411 now, mirroring the `is_app` field in cloud#6399: column, publish-time write, exposure on the summary, detail and template-entry schemas, plus a backfill.

Tests: three new ones cover the transition (still-vouched when the date matches, billable when it has moved, billable when no date arrives). Mutation-checked by removing the date comparison, which fails exactly those three.

dante01yoon (approved):

PR #1115, opened 2026-08-11
fix(i18n): treat a blank review-model variable as unset

mobeenabdullah:
The first real AI review run against `zh` reviewed **0 of 603 entries**. Every one failed with:

```
400 {"type":"invalid_request_error","message":"model: String should have at least 1 character"}
```

## Cause

The workflow passes the reviewer model through from a repository variable:

```yaml
I18N_REVIEW_MODEL: ${{ vars.HUB_I18N_REVIEW_MODEL }}
```

`HUB_I18N_REVIEW_MODEL` is not set, and **GitHub renders an unset `vars.X` as an empty string** rather than omitting the variable. The script read it with `??`:

```ts
const MODEL = process.env.I18N_REVIEW_MODEL ?? "claude-sonnet-5";
```

`??` only falls back on `null`/`undefined`, so `MODEL` was `""`.

From the failing run, note the env line is present but blank:

```
env:
  TRANSLATE_LOCALES: zh
  ANTHROPIC_API_KEY: ***
  I18N_REVIEW_MODEL:
[i18n] review: [zh] reviewing 603 entr(ies)
[i18n] review: [zh] 0 finding(s) across 0 entr(ies).
[i18n] review: [zh] 603 entr(ies) could not be reviewed: ... model: String should have at least 1 character
```

## Why it was quiet

The review step is `continue-on-error: true`, so the job reported **success** and the pipeline published normally. The only symptom was an entire locale coming back unreviewed. The safety design did hold: unreviewable entries are recorded as unreviewed rather than clean, so nothing was wrongly pruned and no bad translation was certified. But a silent zero-work run is exactly the shape of failure worth removing.

This was introduced when the variable was first wired up. Before that the env var was never set at all, so it was `undefined` and the fallback worked.

## Fix

`||` instead of `??`, trimmed, so unset, empty and whitespace all fall back. Same semantics the translator already uses for its own knob (`.i18nrc.cjs` reads `process.env.HUB_I18N_MODEL || "gpt-4.1"`). Extracted to a named function so the blank case is testable rather than resolved inline at module load.

## Checks

- 5 new tests covering unset, empty, whitespace, a real override, and the invariant that it never returns a blank model
- Mutation-checked: restoring `??` turns **4** of them red, so they are not vacuous
- Full suite **488 passing**, `eslint --max-warnings 0` clean, `astro check` **0 errors**

## Audited for the same bug elsewhere

Only this one call site was affected. `TRANSLATE_LOCALES` uses `?? ""` but then splits and filters, so blank correctly means "all locales". `PUBLIC_HUB_API_URL` and `HUB_I18N_MODEL` already use `||`.


mobeenabdullah:
Extended this PR to move both pipelines onto current models, after researching what each one will actually accept. Two findings, and one of them would have broken translation silently.

## Reviewer: `claude-opus-5`

Straightforward. Our request sends `model`, `max_tokens` and `output_config.effort`, and deliberately sends **no** `temperature` (current Claude models reject it). So the model id is the only thing that changes.

One note left in the code for whoever tunes it next: thinking is on by default on Opus 5 and its tokens come out of the same `max_tokens: 16000` budget as the findings. That is precisely the failure that once made the entries with the most to report come back clean, so the two want changing together, with `stop_reason` checked afterwards.

## Translator: `gpt-5.6-terra`, and it needs `temperature: 1` to work at all

This is the part worth reading. Setting the model alone would have failed **every** call.

- `gpt-5.6-terra` is real, from the GPT-5.6 family (Sol / Terra / Luna).
- **GPT-5.x reasoning models accept only the default temperature** and reject an explicit value with a 400.
- `lobe-i18n` defaults `temperature` to `0` and always sends it (`temperature: this.config.temperature`, default `0`). Our `.i18nrc.cjs` never set it, so it inherited `0`.

So the config now pins `temperature: 1` alongside the model. Determinism comes from the reference prompt and the deterministic enforcement pass instead, which is where it should come from anyway.

Checked the rest of the request shape while I was in there: lobe calls `chat.completions.create` and sends `response_format`, `temperature` and `top_p`. It does **not** send `max_tokens`, so the `max_completion_tokens` rename that trips other integrations does not apply here. `top_p` is only sent when `topP` is configured, and ours is not.

Both defaults stay env-overridable, which is the point of the override: a bad id can be reverted from a repo variable without a code change.

## On the red `Validate Translations` check

Not caused by this PR, which touches two script files and a config. It is a merge-ordering artefact on `main`: #1061 was last validated on 08-10 against the 39-term glossary, #1098 then added four preserve terms (`Fun`, `Qwen-Image-Layered`, `ACE-Step`, `Dall-E`) bringing it to 43, and the two merged 21 seconds apart. So `main` now pairs the stricter glossary with content that predates it, and the validator correctly reports 68 fields across ten locales.

The pipeline already has the remedy: the enforce step prunes exactly these fields so they fall back to English on a non-indexable page. It has not yet run against `main` for all ten locales, because the run I dispatched was scoped to `zh`. This will go red on every PR that runs this workflow until that happens, not just this one.

## Checks

488 tests passing, `eslint --max-warnings 0` clean, `astro check` 0 errors, and `.i18nrc.cjs` re-required to confirm it still parses and resolves (`modelName: gpt-5.6-terra | temperature: 1`).

site/scripts/i18n/review-translations.ts
```diff
@@ -42,9 +42,38 @@ const REVIEW_DIR = path.join(process.cwd(), 'src', 'i18n', 'review');
...
+ * Reviewer model. Opus 5 rather than a Sonnet tier because this is a judgement
+ * task in ten languages, which is where the reported weakness is (ar/tr/ko
+ * fluency), and a false verdict here removes a field from a real page.
+ *
+ * Note for anyone tuning `max_tokens` below: thinking is on by default on this
+ * model and its tokens come out of the same budget as the findings. That is the
+ * exact failure that made an earlier run report clean on the entries with the
+ * most to say, so change the two together and check `stop_reason` afterwards.
+ */
+export const DEFAULT_REVIEW_MODEL = 'claude-opus-5';
```
deepme987, line 55: not a concern rn, but something to flag for costs

deepme987 (approved):

site/.i18nrc.cjs
```diff
@@ -73,13 +73,21 @@ Tone: direct, factual, creator-first. The human directs the model; never phrase
...
-  modelName: process.env.HUB_I18N_MODEL || 'gpt-4.1',
+  // Translation model, still overridable by env so a bad id can be reverted from a
+  // repo variable without a code change. A wrong id fails every translation call,
+  // so the ability to switch back instantly is the reason the override stays.
+  //
+  // gpt-4.1 was the previous default because it was the one all ten locales had
+  // actually been run through. It is now safe to raise it on evidence rather than
+  // hope: the AI reviewer scores the output, so a regression shows up as findings
+  // instead of going unnoticed. Revert to 'gpt-4.1' if this model regresses.
+  modelName: process.env.HUB_I18N_MODEL || 'gpt-5.6-terra',
```
dante01yoon: issue: `gpt-5.6-terra` is absent from the context-window map embedded in the pinned `@lobehub/i18n-cli@1.26.0`. Its runtime computes the chunk limit as `Math.floor((modelContext[modelName] - promptTokens) / 3)` and only applies `splitToken` when it is below that value. For this ID the lookup is `undefined`, the limit becomes `NaN`, and `splitToken: 6000` is never applied; the JSON splitter then falls back to effectively one key per request (and an initial empty chunk) instead of the intended larger batches. That materially increases calls, prompt overhead, and exposure to the 30k TPM limit. Please upgrade/patch the CLI so this model has a context window, or otherwise ensure the configured model produces the intended 6000-token grouping, with a focused integration check.

dante01yoon (changes requested):
The blank reviewer-model fix is sound and its tests pass, but the added translator model has a runtime incompatibility with the pinned `@lobehub/i18n-cli@1.26.0` that changes batching behavior. Please make the model known to the CLI's context-window calculation (for example via a compatible dependency update/patch), and add a check that `splitToken: 6000` still produces grouped chunks for the configured model.

christian-byrne (approved):
Looks good but shouldn't you set the variable in github repo settings as well? Do you need me to do it because of access?

mobeenabdullah:
Verified the batching finding against the installed CLI, and it is correct. Pushed a fix.

**Reproducing it.** The bundled `getSplitToken` is:

```js
let n = (Vi[e.modelName || It] - j(t)) / 3;
return e.splitToken && e.splitToken < n && (n = e.splitToken), n = Math.floor(n), n;
```

`Vi` tops out at `gpt-5.2`, so `Vi["gpt-5.6-terra"]` is `undefined` and the limit is `NaN`. `6000 < NaN` is false, so `splitToken` is never applied, and every `<=` inside `splitJSONtoSmallChunks` is false too. Running the real chunker over a 5-key object:

| model | lookup | limit | chunks |
| --- | --- | --- | --- |
| `gpt-5.2` | 400000 | 6000 | 1 |
| `gpt-5.6-terra` | `undefined` | `NaN` | 6, a leading empty one plus one key each |

Exactly the leading empty chunk and one-key-per-request you described.

**Upgrading does not fix it.** I checked 1.27.0: its table is byte-identical to the pinned 1.26.0's and still stops at `gpt-5.2`. There is also no config path around it, since the context window is only ever read from that table.

That left picking a model the CLI knows, or patching the table. The table lives inside the minified `dist/cli.js`, so a patch would be pinned to minified output and would need redoing on every bump, to buy one model generation. I went with `gpt-5.2`: newest id the CLI can size, 400k context, `splitToken: 6000` stays the binding constraint. Still a large step up from the `gpt-4.1` this replaced, and `temperature: 1` is still required.

**Guarding it.** `tests/unit/i18n-translator-model.test.ts` parses the table out of the installed CLI rather than a copy, and asserts the configured model is in it and that `splitToken` still binds. Confirmed it fails on the previous value:

```
The translator is set to "gpt-5.6-terra", which the installed @lobehub/i18n-cli
has no context window for. It will batch one key per request instead of honouring
splitToken. Pick a model the CLI knows, or upgrade it to a version that knows this one.
```

So bumping either the model or the dependency now re-checks them against each other.

Main is merged in, which clears the `Validate Translations` failure this branch inherited.

On the reviewer-model cost note: agreed it is worth watching. The two controls already in place are that verdicts are keyed by a hash of English + translation + rubric version, so a nightly run only pays for entries that actually changed, and `I18N_REVIEW_MAX_ENTRIES` caps a single run. I would rather measure the first full zh run than pre-emptively drop a tier, since a wrong verdict removes a field from a real page. Happy to revisit once there are real numbers.

site/.i18nrc.cjs
mobeenabdullah: Confirmed and fixed in f659a21.

`Vi` in the pinned bundle tops out at `gpt-5.2`, so the lookup is `undefined` and the limit is `NaN`. Running the real chunker over a 5-key object gives 1 chunk on `gpt-5.2` and 6 on `gpt-5.6-terra` (a leading empty one plus one key each), exactly as you described.

1.27.0 ships a byte-identical table, so upgrading does not help, and the context window is only ever read from that table so there is no config path around it. Patching would mean pinning a patch to minified `dist/cli.js` and redoing it every bump for one model generation, so I moved to `gpt-5.2`: newest id the CLI can size, 400k context, `splitToken: 6000` stays binding.

`tests/unit/i18n-translator-model.test.ts` parses the table out of the installed CLI and asserts the configured model is in it. Verified it fails on the old value. Full detail in the PR comment.

site/scripts/i18n/review-translations.ts
mobeenabdullah: Agreed, worth watching. Two controls already limit it: verdicts are keyed by a hash of English + translation + rubric version, so a nightly run only pays for entries that actually changed, and `I18N_REVIEW_MAX_ENTRIES` caps a single run.

I would rather measure the first full zh run than drop a tier pre-emptively, since a wrong verdict removes a field from a real page. I will post the actual token numbers from that run here so the tradeoff is a measured one.

site/scripts/i18n/check-config.ts
```diff
@@ -22,8 +40,42 @@ try {
...
+  const { modelName, splitToken } = require(
+    path.join(process.cwd(), '.i18nrc.cjs')
+  ) as I18nConfig;
+  if (!(modelName in contextWindows)) {
+    errors.push(
+      `translator model "${modelName}" has no context window in the installed @lobehub/i18n-cli. ` +
+        'It would batch one key per request instead of honouring splitToken. Pick a model the CLI ' +
+        'knows, or upgrade it to a version that knows this one.'
+    );
+  } else if (effectiveSplitToken(contextWindows[modelName], PROMPT_TOKEN_ALLOWANCE, splitToken) < 1) {
```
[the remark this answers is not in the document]
mobeenabdullah: Good catch, both halves are real. Fixed in 833fc4a.

`in` matches everything on `Object.prototype`, so `toString` passed the lookup and returned a function as its context window. The capacity check did not stop it either: that yields `NaN`, and `NaN < 1` is false, so the second guard waved it through too. Now an own-key lookup, and the derived limit is tested for finiteness rather than compared, since a non-finite limit is the exact shape of the bug this guards.

Verified: `HUB_I18N_MODEL=toString pnpm i18n:check` now exits 1.

site/tests/unit/i18n-translator-model.test.ts
```diff
@@ -0,0 +1,65 @@
...
+/** The committed config, read with any local env override removed. */
+function committedConfig(): { modelName: string; splitToken: number } {
+  const previous = process.env.HUB_I18N_MODEL;
+  delete process.env.HUB_I18N_MODEL;
+  try {
+    delete require.cache[require.resolve(CONFIG_PATH)];
+    return require(CONFIG_PATH);
+  } finally {
+    if (previous !== undefined) process.env.HUB_I18N_MODEL = previous;
+  }
```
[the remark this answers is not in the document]
mobeenabdullah: Agreed, fixed in 833fc4a. The cache is now cleared in `finally` as well as before the load, so the entry left behind is not one built with the override stripped. No current caller depends on it, but tests share a worker and this is a cheap footgun to remove.

site/scripts/i18n/check-config.ts
```diff
@@ -22,8 +40,52 @@ try {
   errors.push(err instanceof Error ? err.message : String(err));
 }
 
+// 3. The model actually about to be used must be one the CLI can size requests
+// for. The unit test covers the committed default, but the model is also
+// settable from a repo variable that overrides it at run time, and a wrong value
+// there degrades silently: batching collapses to one key per request while every
+// call still succeeds. This is the only place that sees the resolved value, so
+// it is the only place that can catch it. Checked before translating, so a bad
+// override costs one failed step rather than a night of oversized bills.
```
MaanilVerma, line 49: suggestion: This says the check is the only place that sees the resolved model, but in CI it doesn't see it at all. The `Check pipeline config` step in `i18n-update-hub.yml` has no `env:` block, so `HUB_I18N_MODEL` isn't in scope there, while the `Translate content` step two steps later does pass it. Set that variable to a bad id and the check reads the committed `gpt-5.2`, exits 0, and the translator runs on the override anyway. Could we add `HUB_I18N_MODEL: ${{ vars.HUB_I18N_MODEL }}` to the check step? The guard already works once it can see the value, `HUB_I18N_MODEL=gpt-5.6-terra pnpm i18n:check` exits 1.

mobeenabdullah: Good catch, and it is the sharpest kind: the guard could not fail on the one case it was written for. Fixed in `03a11ab`, exactly your suggestion.

```yaml
      - name: Check pipeline config
        working-directory: site
+       env:
+         HUB_I18N_MODEL: ${{ vars.HUB_I18N_MODEL }}
        run: pnpm i18n:check
```

Traced the chain before changing it: `check-config.ts` `require`s `.i18nrc.cjs`, which resolves `process.env.HUB_I18N_MODEL || 'gpt-5.2'`. With no env block that step could only ever see `gpt-5.2`, so pointing the repo variable at a model the pinned CLI cannot size would pass the check and then run the translator on it, one key per request.

Verified both directions locally:

```
pnpm i18n:check                              -> exit 0
HUB_I18N_MODEL=gpt-5.6-terra pnpm i18n:check -> exit 1
  translator model "gpt-5.6-terra" has no context window in the installed
  @lobehub/i18n-cli. It would batch one key per request instead of honouring
  splitToken.
```

**One thing I deliberately did not add:** `I18N_REVIEW_MODEL` alongside it. There are two separate knobs, and `check-config.ts` only validates the translator model through `.i18nrc.cjs`; it never reads the reviewer one. Passing it would imply a check that does not exist. The reviewer model is a different provider and is not subject to this context-window sizing at all, so if it should be guarded that is its own check rather than a free-looking env line here.

Also fixed the comment above the assertion, which claimed this step was the only place that sees the resolved model. That was the belief that produced the gap.

dante01yoon (approved):
The previous batching incompatibility is resolved: the configured translator now uses a CLI-known model, focused coverage verifies the context-window lookup and binding splitToken, and the workflow preflight receives the same HUB_I18N_MODEL override as translation. I also verified the normal config succeeds and an unsupported override fails before translation.

PR #1117, opened 2026-08-11
fix(i18n): prune the fields that violate the widened glossary

mobeenabdullah:
Clears the red `Validate Translations` check, which currently fails on **every** pull request against main, not just the one that happens to run it first.

## What happened

Two PRs merged 21 seconds apart, and the ordering matters:

| | |
| --- | --- |
| #1061 (translations) last validated | 2026-08-10 02:46, against a **39-term** glossary |
| #1098 (reviewer) added 4 preserve terms | `Fun`, `Qwen-Image-Layered`, `ACE-Step`, `Dall-E` → **43** |
| Merged | 16:50:45 and 16:51:06 |

So main pairs a **stricter glossary** with content written before it. The validator is right: 68 fields across ten locales have a brand name translated away. The clearest example is `Fun` in "Wan 2.2 5B Fun Inpaint" rendered as a normal adjective.

## The fix is the pipeline own remedy, run against main

The enforce step already exists for exactly this: a field the model got wrong is **dropped**, so the resolver renders English there and the page stays non-indexable. Nothing is rewritten and no translation is invented. It simply had not run against main for all ten locales, because the run I dispatched yesterday was scoped to `zh`.

```
[i18n] enforce: pruned 68 field(s) total across all locales.
[i18n] validate: no violations.
```

Worst case in any single locale is **0.4%** of fields (`tr`, 15 of 4221), far below the 15% systemic threshold that would mean a broken run rather than a normal tail. Pruned fields per locale: tr 15, zh 12, zh-TW 10, es 9, ar 8, ja 5, pt-BR 4, ko 2, fr 2, ru 1 (68 total).

## Why these fields were wrong in the first place

They are the tail the glossary widening was designed to catch. Listing only a model family protects the head of a name and leaves the rest exposed, so "Wan" was protected inside "Wan 2.2 5B **Fun** Inpaint" while "Fun" was translated as an ordinary word. The four terms were added on measured survival evidence, and this is the existing content those terms now correctly flag.

## After this lands

These fields fall back to English on pages that are already non-indexable, so nothing user-facing regresses. A future translation run can retranslate them correctly now that the terms are protected, and the AI reviewer will score the result.

Deterministic: no API calls, no model involved.


deepme987 (approved):

dante01yoon (approved):
Verified the regression and the proposed remedy against `main`:

- `i18n:validate` fails on `main` with exactly 68 glossary violations.
- Running the existing deterministic `i18n:enforce` path prunes exactly those 68 fields.
- The generated locale-file diff is byte-for-byte identical to this PR, and `i18n:validate` then passes.
- The current head checks are green, including Validate Translations, Build Site, build-and-test, lint, preview, and SEO Audit.

suggestion: The PR description's percentage summary is stale/inaccurate: current output reports `tr` at 15/4221 (0.4%) and `zh` at 12/4221 (0.3%), not `ar` at 0.2% as the worst case. This does not affect the fix or the 15% safety threshold, but updating the description would keep the evidence precise.

mobeenabdullah:
You were right, and thanks for the independent reproduction. I have corrected the description.

The error was mixing two units in one paragraph: the per-locale list was *lines changed* in the diff, while the "worst case" figure was a *field* count, so `ar` looked worst because its 8 fields sat next to everyone else's line counts. Recounting pruned fields consistently:

| locale | pruned fields | of 4221 |
| --- | --- | --- |
| tr | 15 | 0.4% |
| zh | 12 | 0.3% |
| zh-TW | 10 | 0.2% |
| es | 9 | 0.2% |
| ar | 8 | 0.2% |
| ja | 5 | 0.1% |
| pt-BR | 4 | 0.1% |
| ko | 2 | <0.1% |
| fr | 2 | <0.1% |
| ru | 1 | <0.1% |

68 total, matching your numbers exactly. The 15% systemic threshold conclusion is unchanged.

PR #1126, opened 2026-08-13
ci(site): stop the preview deploy uploading the whole repository

mobeenabdullah:
## The failure

The `preview` job dies partway through its upload:

```
Uploading [===============-----] (1.3GB/1.8GB)
Error: fetch failed
##[error]Process completed with exit code 1.
```

It reads on a PR as broken code. It is a transfer large enough to be unreliable, and it succeeds or fails by luck rather than by anything in the diff.

## Why 1.8GB

Both `preview-site.yml` (per PR) and `preview-cron.yml` (nightly) deploy the same way:

```
npx vercel@latest deploy --cwd "$GITHUB_WORKSPACE" --prebuilt --archive=tgz
```

`--archive=tgz` tars the whole working directory, and there was no `.vercelignore`. A prebuilt deploy reads only `.vercel/output`, which the step copies to the repo root a few lines above, so the rest of the tar is dead weight:

| Directory | Tracked size |
| --- | --- |
| `input` | 1.9 GB |
| `output` | 933 MB |
| `templates` | 497 MB |
| `thumbnail` | 362 MB |
| `docs` | 128 MB |
| `site` | 48 MB |

Roughly 3.9GB of tracked content, plus `node_modules`, compressed into the 1.8GB that fails to upload.

## The change

One new file, `.vercelignore`, excluding what a prebuilt deploy never reads.

**Every pattern is anchored with a leading slash, deliberately.** `.vercelignore` uses gitignore matching, where an unanchored pattern matches at any depth. A bare `output` would also exclude `.vercel/output` and deploy an empty site, and a bare `node_modules` would strike a bundled dependency inside the output. `/site` already covers `site/node_modules`, so anchoring costs nothing.

Exclusions are listed explicitly rather than ignoring everything and re-including the output, so no rule here can exclude the artifact being deployed.

## Verification

- All 21 patterns confirmed anchored, and all 20 paths confirmed to exist at the repo root
- Checked that no pattern can match `.vercel/output`

**This PR cannot verify itself, which is worth stating plainly.** `preview-site.yml` triggers on `paths: ['site/**']` and this changes only a root file, so the `preview` check does not run here. I verified it instead by dispatching **Preview Site** manually against this branch via `workflow_dispatch`, which exercises the real deploy under the new ignore list. The upload size in that run's log is the measurement.

Nothing outside CI is touched, and no site code changes.

### DIFF
--- a/.vercelignore
+++ b/.vercelignore
@@ -1,5 +1,6 @@
-# The preview deploy runs `vercel deploy --cwd <repo root> --prebuilt --archive=tgz`
-# (.github/workflows/preview-site.yml). `--archive=tgz` tars the whole working
+# Both preview deploys run `vercel deploy --cwd <repo root> --prebuilt --archive=tgz`:
+# .github/workflows/preview-site.yml (per pull request) and preview-cron.yml
+# (nightly). They share this ignore list. `--archive=tgz` tars the whole working
 # directory, while a prebuilt deploy needs only `.vercel/output`, so without this
 # file every preview uploads gigabytes of repository content. That upload dies
 # intermittently with "Error: fetch failed" partway through, which reads on a PR
@@ -40,6 +41,8 @@
 /README.md
 
 # Sources and dependencies. The build already ran, and its output was copied to
-# the repo-root .vercel/output before this deploy step.
+# the repo-root .vercel/output before this deploy step. `/site` already covers
+# site/node_modules, so this rule is anchored too rather than left to match at
+# any depth, where it could strike a bundled dependency inside .vercel/output.
 /site
-node_modules
+/node_modules
--- a/.vercelignore
+++ b/.vercelignore
@@ -1,0 +1,45 @@
+# The preview deploy runs `vercel deploy --cwd <repo root> --prebuilt --archive=tgz`
+# (.github/workflows/preview-site.yml). `--archive=tgz` tars the whole working
+# directory, while a prebuilt deploy needs only `.vercel/output`, so without this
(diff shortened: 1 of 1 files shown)


mobeenabdullah:
Closing this. The premise was wrong, and I can show it rather than argue it.

I dispatched **Preview Site** against this branch to measure the effect, since the `preview` check does not trigger on a root-only change. The deploy succeeded, and the upload was:

```
Uploading [====================] (1.8GB/1.8GB)
```

Identical to before. The ignore list changed nothing, and this line from the same log says why:

```
Using prebuilt build artifacts from .vercel/output
```

With `--prebuilt`, the CLI only ever packaged `.vercel/output`. It was never uploading `input/`, `output/`, `templates/` or `node_modules`, so excluding them saves nothing. **The 1.8GB is the built site itself**, not repository content riding along, which is what I assumed from the size alone.

So the real question is why a prerendered static build weighs 1.8GB, and that is a different investigation: the build spends 220s prerendering static routes and then copies them into the output. Whatever the answer, it is not a `.vercelignore` change, and an ignore file that provably does nothing is worse than none, because the next person reads it as protection that already exists.

CodeRabbit's two findings were both fair and are worth recording even though the file is going away: the comment named only one of the two workflows that deploy this way, and the unanchored `node_modules` rule would have matched at any depth including inside `.vercel/output`. I had anchored everything else for exactly that reason and missed that one.

The original symptom is real and still open: the upload fails partway through often enough to read as broken code. It is a size problem, so the fix has to reduce what the build emits.

PR #1132, opened 2026-08-14
fix(site): add manual 301s for model-slug 404s with no template data yet

balpreetgrowthnatives:
## Summary

Adds manual 301 redirects for ~35 `/workflows/model/*` URLs that currently 404, sourced from a site-wide 404 crawl. Model-page redirects are normally auto-derived by `deriveModelGroups()` from each template's `models` field (`site/astro.config.mjs`), but these slugs don't appear in any template yet, so nothing generates a redirect for them today.

## Changes

- New `manualModelSlugRedirects` map in `site/astro.config.mjs`, merged into the `redirects` config as `{ ...manualModelSlugRedirects, ...modelSlugRedirects }` — the auto-derived map always wins on overlap, so once a template does get tagged with one of these slugs, the manual entry becomes a no-op rather than a conflicting source of truth.
- Variant → canonical family redirects: all `wan*`, `flux*`, `kling*`, `ltx*`, `seedance*`, `qwen-image*` variants, plus `z-image-turbo → z-image` and `nano-banana-2 → nano-banana-pro`.
- No-live-page families (`sdxl`, `sd1-5`, `sd3-5`, `hunyuan-3d`, `gemini3-pro-image-preview`, `gpt-image-1`, `gpt-image-1-5`, `none`) redirect to the model index (`/workflows/model/`) rather than staying dead ends.
- Both trailing-slash forms are keyed per variant, matching the existing convention documented above `modelSlugRedirects`.

## Review Focus

- **Seedream is deliberately excluded from the Seedance redirect.** The source crawl suggested `seedream-*` → `/workflows/model/seedance/`, but this codebase's own `FAMILY_RULES`/`PRIORITY_MODELS` in `model-groups.ts` already treat Seedream as a distinct family from Seedance (separate regex rule, separate priority entry). Merging it into Seedance's page would contradict that taxonomy, so `seedream-4-0`, `seedream-4-5`, and `seedream-5-0-lite` route to the model index instead until Seedream has its own qualifying page.
- This introduces more instances of an existing route-collision warning (`"<slug>" is defined in both "<slug>" and "<slug>/"`) from keying both slash forms — confirmed via `astro sync` that this exact warning already fires 42 times on `main` before this change; this PR doesn't introduce a new class of issue, just more entries in the pre-existing accepted pattern.
- No changes to `deriveModelGroups()`, `FAMILY_RULES`, or any template content — purely additive, static redirect data.


site/astro.config.mjs
```diff
@@ -81,6 +81,65 @@ const modelSlugRedirects = Object.fromEntries(
...
+      'gemini3-pro-image-preview',
+      'gpt-image-1',
+      'gpt-image-1-5',
+      'seedream-4-0',
+      'seedream-4-5',
+      'seedream-5-0-lite',
+      'none',
+    ].map((variant) => [variant, null]),
+  ]
+    .filter(([variant]) => !canonicalModelSlugs.has(variant))
```
dante01yoon, line 115: issue: This changed expression currently fails the required `astro check`. The live CI log reports `ts(2345)` here because tuple inference widens `variant` to `string | null` (the second tuple element is intentionally nullable), while `canonicalModelSlugs.has()` accepts only `string`. Please give these entries an explicit tuple type so the variant remains a string; the PR's current status rollup is failing on this error.

site/astro.config.mjs
```diff
@@ -81,6 +81,65 @@ const modelSlugRedirects = Object.fromEntries(
...
+    .flatMap(([variant, target]) => {
+      const destination = target ? `/workflows/model/${target}/` : '/workflows/model/';
+      return [
+        [`/workflows/model/${variant}`, destination],
+        [`/workflows/model/${variant}/`, destination],
```
dante01yoon: issue: The deployed preview confirms this slash-form entry is discarded by Astro's route collision: `/workflows/model/wan2-1` returns 301 to `/workflows/model/wan/`, but `/workflows/model/wan2-1/` returns 404. I reproduced the same split for `flux-2-dev`, `sdxl`, `seedream-4-0`, and `none`, and CI logs the corresponding duplicate-route warnings. This leaves the canonicalized trailing-slash URLs—the behavior this PR says it restores—broken. Could we handle both forms in a non-colliding routing layer and validate both against the deployed preview?

dante01yoon (changes requested):
I found two blocking issues: the changed config fails `astro check`, and the deployed preview shows that trailing-slash variants still return 404. Details are inline.

site/astro.config.mjs
balpreetgrowthnatives: Fixed in bc0e9773. The tuple inference issue was exactly as described `.map(v => [v, 'wan'])` doesn't produce a tuple type, so mixing those literals with the `[variant, null]` entries widened `variant` to `string | null` and broke `canonicalModelSlugs.has(variant)`.

Added a typed `pair(variant, target)` helper with an explicit `@returns {[string, string | null]}` JSDoc annotation, so every entry is a proper tuple and `variant` stays `string` throughout. `astro check` is back to the 2 pre-existing errors in `scripts/i18n/review-translations.ts` (unrelated to this file, present on `main`) — verified locally before and after with `git stash` to confirm this diff isn't the source.

site/astro.config.mjs
balpreetgrowthnatives: Confirmed and fixed this wasn't just my new entries, it's a shared bug. I rebuilt locally and inspected the actual `.vercel/output/config.json`: both the `/foo` and `/foo/` keys were collapsing to the byte-identical regex `^/workflows/model/<slug>$` (no trailing-slash handling either way), so the second key was never providing real coverage — matches the known upstream limitation (withastro/astro#13134): Astro's `redirects` config can't express both slash forms for one path.

Root-caused and fixed in ae46e222: added a postbuild step (`site/scripts/fix-redirect-trailing-slash.ts`, wired into `pnpm run build`) that patches the generated Vercel manifest — rewrites each redirect regex to `^/foo/?$` and drops the now-redundant duplicate entries. This fixes it for every model-page redirect, including `flux-2-dev`, `sdxl`, `seedream-4-0`, and `none` — the pre-existing `modelSlugRedirects` cases you tested, not just this PR's additions.

Verified locally against a fixture built from the real manifest shape (redirect entries, header-injection rules, rewrites, and the filesystem-handle marker) — confirmed only redirect routes are touched, duplicates are deduped, and the resulting regex matches both `/foo` and `/foo/` while correctly rejecting `/foobar`.

Could you re-run the preview deploy and re-test `/workflows/model/wan2-1/`, `flux-2-dev/`, `sdxl/`, `seedream-4-0/`, and `none/` against it? Want your confirmation against the live preview before this gets marked resolved, same as your original ask.

PR #1134, opened 2026-08-14
fix(i18n): drop the ru and ar machine layers so the pipeline retranslates them

mobeenabdullah:
Russian and Arabic cannot ship as they are: the AI review prunes over 20% of their fields (ru 20.5%, ar 21.1%), and past 15% the enforce step correctly fails the run rather than English-falling-back a fifth of a locale. Their committed translations predate the translator model upgrade; every locale translated since passes at 4.9% to 10%.

This deletes only the machine-translation layer for the two locales. The nightly run then regenerates both from scratch with the current model and re-reviews them. Human seeds and overrides are separate files and untouched. Nothing here flips any locale indexable.

Deleting from the automation branch does not work, verified today: the pipeline restores branch state with `git checkout <branch> -- site/src/i18n`, which adds and overwrites but never deletes, so main's copies always survive and the translator sees nothing missing. That restore semantics is worth a follow-up fix; this PR stays data-only.

If this merges before tonight's cron, the cron performs both retranslations; otherwise I will dispatch them manually after merge.


mobeenabdullah:
Superseded: #1131 carried the same ru/ar machine-layer deletions (the automation branch already held those commits) and has merged, so the nightly run now regenerates both locales with the current model. Deleting the branch.

PR #1144, opened 2026-08-17
 fix(site): sync navbar links and featured card with comfy.org

MaanilVerma:
## Summary

The Hub navbar had fallen behind comfy.org since the last sync. The Community dropdown was missing two links, the Products featured card still advertised a retired release, and the cloud button used older wording. This brings all three back in line with the live marketing site.

## Changes

**What**

- The Community dropdown now includes Forward Deployed Creatives and Events, both marked new, in the same order comfy.org uses.
- The Products featured card now promotes the LTX 2.5 release instead of the old Comfy MCP card, with the matching image, title, and "EXPLORE NOW" link through to the LTX page.
- The cloud call to action reads "TRY CLOUD FOR FREE" to match the marketing site.
- Fixed a stale translation bug the fallback was hiding: eight non-English locales still had the featured title hardcoded to "Seedance 2.0", a release two syncs old, so they would never have picked up the LTX copy. All eight now read LTX 2.5.

**Breaking**

None. Existing nav links, badges, analytics attributes on the cloud CTA, and the mobile sheet are unchanged. English and Chinese CTA wording already matched and were left alone.

## Testing

The navbar renders from a Vue island, so the dropdown contents only exist after client hydration. Static HTML checks would miss them, so verification was done by driving a real browser plus checking that every new link and asset actually resolves.

- Drove the desktop nav at 1600px: the Community dropdown lists Comfy Workflows, Forward Deployed Creatives (new), Gallery, Events (new), Affiliates (new), Learning in order, with the two new items linking to `/forward-deployed-creatives` and `/events`. The Products card shows the LTX image, "LTX 2.5" title, and "EXPLORE NOW". The cloud button reads "TRY CLOUD FOR FREE".
- Confirmed all four new targets return 200: the LTX card image, `/ltx-2.5`, `/forward-deployed-creatives`, and `/events`.
- Checked string resolution across all 11 locales: every one now renders an LTX title, with no Seedance or MCP text left in the featured keys.

## Screenshots

1-products-ltx-card
2-community-fdct-events
3-cta-try-cloud-for-free
4-zh-products-ltx
5-mobile-community


dante01yoon (approved):
Reviewed route reuse, navigation consumers, locale coverage and fallback behavior, localized copy, asset and destination availability, and the current CI failures. Prior review comments are resolved; the failing jobs completed their actual checks successfully and failed only while posting GitHub comments. No blocking findings.

PR #1145, opened 2026-08-17
feat(i18n): land the zh native review as overrides and sign-off records

mobeenabdullah:
Lands the Chinese native review (Zhixiong Lin, with Tiger's spot-check endorsement in Slack) as two data files:

**overrides/zh.json**: 88 corrected fields across 73 workflows. 85 are Lin's own wording, including the one critical (Qwen-Image-Layered half-translation, his fix is better than the AI's). 3 are AI-suggested fixes that still applied. The other 209 suggested-fix rows were already superseded before review landed: the nightly loop had pruned and retranslated those fields with the current model, and the AI reviewer re-reviewed the fresh text. 7 items are deferred (content churned between snapshots, will confirm wording with Lin) and are safe meanwhile: those fields serve English or fresh reviewed text.

**reviews/zh.json**: checksum-bound sign-off records per workflow, scope stated on each record. Later changes to English or the resolved translation de-index that page until re-review, per the predicate design.

Verified with the site's own resolver: with zh flipped, **519 of 612 detail pages become indexable**; 93 hold themselves back per-page (English fallback in a required field) and rejoin as resolved. i18n:validate clean, 553 unit tests green.

Nothing in this PR flips zh: the INDEXABLE_LOCALES flip is a separate one-line PR after this merges, so the flip diff is reviewable on its own.


imick-io (approved):
lgtm

PR #1152, opened 2026-08-18
feat(i18n): flip zh to indexable

mobeenabdullah:
Turns Chinese on. One line: `INDEXABLE_LOCALES` gains `'zh'`.

> **Stacked on #1151.** Without it this change fails the build: the locale route currently aborts on any page the predicate rejects, and 96 of 615 are legitimately held. Base retargets to `main` once #1151 merges.

## Why now

Chinese passed native review. Zhixiong Lin worked the flagged sheet, and his 85 corrections landed as overrides plus a checksum-bound sign-off record in #1145, after Tiger Fu's spot check endorsed the AI recommendations. The sign-off is what the predicate checks, so this is the gate the design intends.

## What it turns on

Resolved against the live hub index: **519 of 615 pages become indexable.**

- self-canonicals and sitemap entries for those 519
- reciprocal hreflang between them and their English twins
- the footer language switcher, which derives itself from this constant and has rendered nothing until now

The remaining **96 are held** by the per-page predicate (93 with an English-fallback field, 3 published since the last translation run). They keep today's behavior exactly: Chinese for humans, canonical to English, out of the sitemap, out of every hreflang cluster. They join automatically as their fields land, no further PR.

## Verified locally before opening this

Dev server with this change, driven by Playwright, 11/11:

| Check | Result |
|---|---|
| switcher renders on English `/workflows/` | `English  简体中文`, linking `/zh/workflows/` |
| switcher on `/zh/workflows/` | 简体中文 marked `aria-current`, English offered back |
| indexable zh page canonical | self, `/zh/workflows/…` |
| indexable zh page hreflang | emits `zh` alternate |
| held page | 200, still Chinese (`LTX-2.3: 文生视频`), canonical to English, no `zh` alternate |
| English twin of an indexable page | advertises `zh` |
| English twin of a held page | does **not** advertise `zh` |

Full unit suite (560) green with the flip applied, lint and format clean.

Note the switcher label is **简体中文**, not `中文` — that is `LANGUAGES.zh.nativeName` in `src/i18n/config.ts`, and it pairs correctly with `繁體中文` for zh-TW.

## Two things reviewers should know

**Deploying is a separate manual step.** Merging does not deploy this repo — `deploy-site.yml` runs on `workflow_dispatch` or after a successful PyPI publish. Nothing reaches comfy.org until that is dispatched.

**Pre-existing, not from this PR:** English listing pages (e.g. `/workflows/category/image/`) emit every hreflang twice, because `BaseLayout` emits a set and the page passes a second `HreflangTags`. That is live on production today and is unchanged by this change; it belongs with the GTM-349 indexability work.

### DIFF
--- a/site/src/lib/i18n/locales.ts
+++ b/site/src/lib/i18n/locales.ts
@@ -49,6 +49,35 @@ export const SUPPORTED_HUB_LOCALES: readonly Locale[] = LOCALES;
 export const INDEXABLE_LOCALES: readonly Locale[] = [];
 
 /**
+ * Fail-closed guard for a flipped locale, asserted once per build after the
+ * locale routes have resolved every page.
+ *
+ * An individual page the predicate rejects is NOT a build failure: it is held
+ * back (rendered in-locale for humans, canonical to English, absent from the
+ * sitemap and from every hreflang cluster), which is the graceful degradation
+ * the design specifies. Holds are the steady state, not an anomaly — workflows
+ * are published to the hub between translation runs, so a flipped locale always
+ * trails the catalog by however many entries landed since its last run.
+ *
+ * A flipped locale with ZERO indexable pages is a different animal: its
+ * translation artifacts are missing or unreadable, and the language would
+ * silently serve English under its own URLs. That fails the build.
+ */
+export function assertFlippedLocalesIndexable(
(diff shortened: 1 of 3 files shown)


imick-io (approved):
lgtm

PR #1156, opened 2026-08-18
feat(site): API link on workflow hub cards

nav-tej:
Adds a quiet API link next to Try now on every workflow card, deep-linking to the #api section that #1149 adds to detail pages. Surfaces run-from-code before a card is even opened and gives crawlers an internal-link path from the hub to the API-intent sections.

- Stacked on #1149 (base is `nav/workflow-api-payload`); merge that first, then retarget or merge this
- Plain anchor so it works in both the Vue island grid and SSR-rendered cards; `@click.stop` keeps the card click intact
- `#api` targets already have `scroll-mt`; section renders on every detail page, so the link is never dead

Validation: build clean, lint clean, 553/553 unit tests, SSR output on /workflows/ carries the anchors (`href="/workflows/{slug}/#api"`).


mobeenabdullah (approved):
LGTM, verified on the preview build. 30 API links render on desktop and mobile; hrefs are correct and locale-aware (/workflows/{slug}/#api, trailing slash preserved), the #api target exists on detail pages (id="api" + scroll-mt-8 from #1149) so no dead links, and the card root is a div so the inner anchors aren't nested. @click.stop isolates them from the card click. Layout is clean beside the reveal pill.

Only note is merge order: this is stacked on #1149, so #1149 lands first, then retarget this to main.

nav-tej:
On hold along with #1149 (its base), pending Uy's version of this surface.

Two notes for whoever picks this up:

- The preview link in the bot comment above was stale, and the rebuild I triggered failed at the deploy step rather than the build step: the render function had grown past Vercel's 250MB limit. That is fixed on the base branch (`9d97f2e`) and merged in here, so CI should produce a fresh preview on this push.
- This branch was several commits behind its base, which is why the preview showed none of the API-section work even before that. It is caught up now.

PR #1158, opened 2026-08-19
feat(site): highlight the API code blocks, add a TypeScript tab

uytieu:
# Changes on top of #1149
https://github.com/user-attachments/assets/7a19dc48-856d-4df0-860f-203087024169

This branch adds syntax highlighting, code block controls, and a language toggle
to the API section.

## Syntax highlighting
image

Shiki 3.22.0 highlights the workflow payload and the SDK snippets. Astro already
depends on Shiki, so the install added no new packages.

The theme is `kanagawa-wave`. Of the bundled dark themes, its background is the
closest to `--color-primary-comfy-ink`. Its accent sits 15 degrees from
`--color-primary-comfy-yellow` in OKLCH hue.

The build highlights the SDK snippets. The browser highlights the payload after
the panel fetches it. Grammars load one language at a time. The common path
costs 58.6 KB, gzipped.

Highlighting stops above 128 KB of payload. Markup grows about 7 times the
source size. A 1 MB payload produces 7.7 MB of DOM and blocks the main thread
for 284 ms. The limit covers 93% of templates. Larger payloads keep the plain
text.

## Code block headers
image

The payload header shows a file icon and the file name. It replaces the
`GET <url>` string. Two controls sit at the right of the header. The copy button
reads "Copy to clipboard". The download button reads "Download file" and uses
the same URL as the hero button.

image

The SDK block has a new header with two tabs: TypeScript and Python. TypeScript
is the active tab. The copy button copies the active tab.

## TypeScript snippet

`buildSdkSnippet` now takes a `lang` option. The default is `python`, so the 13
tests from #1149 pass without change.

The TypeScript output follows the documented quickstart at docs.comfy.org. The
snippet awaits `workflows.fromFile` and `run`. It does not await
`assets.fromFile`. It indexes outputs. It does not use a loop. Both tabs address the
same node IDs from the graph.

Each tab shows only its own install line. Before this change, one block showed
`pip install` and `npm i` together.

The `sdkIntro` copy now names TypeScript before Python, to match the active
tab.

## Appearance
image

- The scrollbar track is transparent. The thumb uses `--color-border`. A new
  `scrollbar-code` utility holds this color, because `scrollbar-thin` also serves the
  search popover and the browse toolbar.
- Both code blocks wrap long lines. The longest unbroken token in `templates/`
  is 151 characters, so `break-words` is enough.
- The payload block has a bottom fade. It reuses the gradient from
  `ExpandableText.vue`. The fade hides at the end of the scroll.

## Contextual banner
image

The API key sentence moved out of the note paragraph into a full width banner.
The banner holds two text links: "SDK docs" and "Get an API key". The note is
now a normal paragraph at `text-sm`.

The links are text buttons, not buttons with a border. The button size in use
across this repo is `nav`, which is too large for a one line banner.

## Removals

`canonicalJsonUrl` lost its last consumer, so this branch removes it from three
files. This branch also removes the `absoluteUrl` import in
`TemplateDetailPage.astro`.

## New files

- `src/lib/highlight.ts`
- `src/components/template-detail/CodeCopyButton.vue`
- `src/components/template-detail/CodeDownloadButton.vue`
- `src/components/template-detail/code-block-styles.ts`
- `tests/unit/highlight.test.ts`

## Tests

The suite has 576 tests, and all of them pass. This branch adds 10: five for
`highlight.ts` and five for the TypeScript snippet.

`pnpm lint`, `pnpm run check`, `pnpm format:check`, and the production build all
pass.

## Known issues

`pnpm knip` reports `codeBlockControl` as an unused export. This report is
wrong. The knip `project` glob covers `.astro`, `.ts`, `.js`, and `.mjs`, but
not `.vue`, so knip cannot see the two components that import it. An added
`.vue` glob surfaces 26 unused files across the repo, which is work for a
separate change.

---


mobeenabdullah (approved):
LGTM, verified on the preview #api section: Shiki highlighting renders on the payload and SDK snippets, the TypeScript/Python tabs work (TS active), copy + download controls present, clean layout. Checked the risky bits: v-html is only ever fed Shiki output (escaped), oversized/failed highlights fall back to escaped text, so no XSS; the TS snippet matches the documented quickstart (await workflows.fromFile, un-awaited assets.fromFile, await run, real node IDs); canonicalJsonUrl is cleanly removed with no dangling refs; the control buttons have proper aria-labels.

Two non-blocking notes:
- The SDK tabs have role=tablist/tab + aria-selected but no role=tabpanel / aria-controls and no arrow-key navigation, so it is not the full ARIA tabs pattern (buttons still activate via Enter/Space). Minor a11y.
- This removes #1149's visible 'GET <url>' string in favor of the filename header, so the scriptable URL now only lives on the download button. Worth confirming that's intended since #1149's API-intent was to show the canonical GET URL.

Stacking: this and #1156 both sit on #1149. #1149 lands first (or fold this into it); then retarget to main.

PR #1160, opened 2026-08-19
fix(i18n): unfreeze translation runs (JSON mode) and save partial progress on failure

mobeenabdullah:
## What was broken

Every `i18n: Update Hub Translations` run with content to translate has failed since 2026-08-14, so no locale (zh included) has received translations for newly published workflows, and the ru/ar retranslation could never start.

Root cause, from the run logs: gpt-5.2 in plain-text mode returns a complete, correct JSON translation and then appends one stray trailing `}` (the content is full of `{{PTnn}}` placeholder braces, which likely throws off its brace counting). `JSON.parse` and lobe's dirty-json fallback both reject that byte, lobe discards the whole chunk and aborts the locale, enforce sees "0 entries", the run fails, and the success-gated publish step throws the night's spend away. Deterministic, every night. The failing payload from run 32145071626 parses cleanly up to its final character.

## The fix, three commits

1. **Force OpenAI JSON mode** (`site/.i18nrc.cjs`): lobe's `experimental.jsonMode` makes it send `response_format: {type: "json_object"}`, so the API itself guarantees syntactically valid output. The stray brace becomes impossible at the protocol level rather than merely unlikely. (The reference prompt already contains the word "JSON", which json_object requires; a model that rejected the flag would fail loudly with a 400.)

2. **Save partial progress on failure** (workflow): a failed night now commits its translation output to the same stable automation branch the existing restore step already reads, so a genuinely rate-limited backfill night accumulates instead of discarding. Guards: staged files that don't parse as JSON are excluded (a truncated mid-write file must not poison the branch), nothing-to-save exits clean, an unchanged rerun never force-pushes, and the failure path never creates a PR. Merging stays gated exactly as before: the automation PR's own Validate Translations check is red on a partial head.

3. **Retry the partial-progress push**: the first live exercise of the save path lost its commit to a GitHub-side HTTP 500 on push. Three attempts with backoff.

## How it was tested

- Local scratch-repo simulation of the save step, five cases: first save creates the branch; corrupt JSON is excluded while valid progress commits; an early failure saves nothing; an identical rerun no-ops; a second night accumulates a new locale on top of restored progress.
- Live end-to-end on a throwaway branch (stable branch renamed so the real one is untouched, a test-only step forces "progress then failure"): run [32225067559](https://github.com/Comfy-Org/workflow_templates/actions/runs/32225067559) proved the failure path fires and exposed the HTTP 500 (hence commit 3); run [32227947002](https://github.com/Comfy-Org/workflow_templates/actions/runs/32227947002) then saved and pushed the branch successfully, and, with JSON mode active, **gpt-5.2 translated 190k tokens of real zh backlog with zero parse failures**, the first successful gpt-5.2 translation in this pipeline. The throwaway branches will be deleted after review.
- `pnpm run lint`, `pnpm run check`, `npx vitest run` (590/590) all green on the branch.

## After merge

Dispatch `locale: ru`, then `locale: ar`; each run now either finishes or banks its progress, so the backfill converges over a small number of runs. ru/ar remain non-indexable regardless; this only restores the machine layer.

### DIFF
--- a/.github/workflows/i18n-update-hub.yml
+++ b/.github/workflows/i18n-update-hub.yml
@@ -287,3 +287,57 @@ jobs:
               --title "chore(i18n): update hub workflow translations" \
               --body "Automated hub translation refresh. English content regenerated from the Hub index, human seeds applied, gaps machine-translated, deterministic validators run. A partial (rate-limited) run still commits its progress to this same branch so the next run resumes. Does not flip any locale to indexable."
           fi
+
+      # A failed night must still save its translation progress. Without this, a
+      # from-scratch backfill (the ru/ar retranslation) can never converge: the
+      # run rate-limits partway, enforce/validate correctly refuse the partial
+      # result, the success()-gated publish above skips, and the night's OpenAI
+      # spend is discarded — so the next night restarts from the same point
+      # forever. This step commits the partial output to the same stable branch
+      # the "Restore partial progress" step reads, so consecutive nights
+      # accumulate until a run finally completes and publishes normally.
+      #
+      # Merging stays gated exactly as before: this step never creates a PR, and
+      # if the automation PR is already open its own Validate Translations check
+      # goes red on the partial head, so nobody can merge a state main's
+      # validators reject. Only what gets REMEMBERED changes, not what gets
+      # published.
+      - name: Save partial progress on failure
+        if: failure()
+        run: |
+          set -euo pipefail
+          STABLE_BRANCH="i18n/hub-translations"
+          git config user.name 'github-actions[bot]'
+          git config user.email 'github-actions[bot]@users.noreply.github.com'
+          git add site/src/i18n
+          # A truncated write (the process can die mid-file on this path) must
+          # not poison the branch the next run restores from: drop any staged
+          # .json the next run could not parse. Deletions are skipped (nothing
+          # to parse) and stay staged.
+          while IFS= read -r f; do
+            case "$f" in
+              *.json)
+                if ! node -e 'JSON.parse(require("fs").readFileSync(process.argv[1], "utf8"))' "$f" 2>/dev/null; then
+                  echo "::warning::Excluding unparseable $f from the partial save"
+                  git restore --staged -- "$f"
+                fi
+                ;;
+            esac
+          done < <(git diff --cached --name-only --diff-filter=d)
+          if git diff --cached --quiet; then
+            echo "Run failed before producing any translation changes; nothing to save."
+            exit 0
+          fi
+          # Same no-op guard as the publish step: an identical rerun (e.g. a
+          # night that failed before translating) must not force-push a new head
+          # and rerun the automation PR's CI for zero progress.
+          if git rev-parse --verify --quiet "refs/remotes/origin/$STABLE_BRANCH" >/dev/null; then
+            if git diff --quiet "refs/remotes/origin/$STABLE_BRANCH" -- site/src/i18n; then
+              echo "No change vs the saved partial progress; leaving the branch untouched."
+              exit 0
+            fi
+          fi
+          git checkout -B "$STABLE_BRANCH"
+          git commit -m "chore(i18n): save partial translation progress (run failed, do not merge)"
+          git push --force-with-lease origin "$STABLE_BRANCH"
+          echo "Partial progress saved to $STABLE_BRANCH; the next run resumes from it."
--- a/.github/workflows/i18n-update-hub.yml
+++ b/.github/workflows/i18n-update-hub.yml
@@ -330,7 +330,9 @@ jobs:
           # A FULLY misaligned content file (model-mangled keys, none in the
           # English source) fails validation and enforce cannot prune it, so
           # banking it would carry the mangled keys forward. Partially stale
-          # files still bank, matching the validator's tolerance.
+          # files still bank, matching the validator's tolerance. MUST run
+          # after the parse pass above: it reads en.json from the worktree,
+          # which that pass has already restored to a good version if truncated.
           while IFS= read -r f; do
             case "$f" in
               site/src/i18n/content/en.json) ;;
--- a/.github/workflows/i18n-update-hub.yml
+++ b/.github/workflows/i18n-update-hub.yml
@@ -319,7 +319,19 @@ jobs:
               *.json)
                 if ! node -e 'JSON.parse(require("fs").readFileSync(process.argv[1], "utf8"))' "$f" 2>/dev/null; then
                   echo "::warning::Excluding unparseable $f from the partial save"
-                  git restore --staged -- "$f"
+                  # Excluding must not UNDO banked progress: a plain staged
+                  # restore resets the path to main's version, so a previously
+                  # saved translation regenerated as malformed would silently
+                  # drop off the stable branch at commit time. Take the last
+                  # good version from the branch when it has one, else main's,
+                  # else drop the new file entirely.
+                  if git cat-file -e "refs/remotes/origin/$STABLE_BRANCH:$f" 2>/dev/null; then
+                    git restore --source="refs/remotes/origin/$STABLE_BRANCH" --staged --worktree -- "$f"
+                  elif git cat-file -e "HEAD:$f" 2>/dev/null; then
+                    git restore --source=HEAD --staged --worktree -- "$f"
+                  else
+                    git rm -qf --ignore-unmatch -- "$f"
+                  fi
                 fi
                 ;;
             esac
@@ -330,9 +342,11 @@ jobs:
           fi
           # Same no-op guard as the publish step: an identical rerun (e.g. a
           # night that failed before translating) must not force-push a new head
-          # and rerun the automation PR's CI for zero progress.
+          # and rerun the automation PR's CI for zero progress. --cached, because
+          # the commit below is built from the index and an excluded malformed
+          # file may legitimately leave the worktree differing from the branch.
           if git rev-parse --verify --quiet "refs/remotes/origin/$STABLE_BRANCH" >/dev/null; then
-            if git diff --quiet "refs/remotes/origin/$STABLE_BRANCH" -- site/src/i18n; then
+            if git diff --cached --quiet "refs/remotes/origin/$STABLE_BRANCH" -- site/src/i18n; then
               echo "No change vs the saved partial progress; leaving the branch untouched."
               exit 0
             fi
--- a/.github/workflows/i18n-update-hub.yml
+++ b/.github/workflows/i18n-update-hub.yml
@@ -288,20 +288,12 @@ jobs:
               --body "Automated hub translation refresh. English content regenerated from the Hub index, human seeds applied, gaps machine-translated, deterministic validators run. A partial (rate-limited) run still commits its progress to this same branch so the next run resumes. Does not flip any locale to indexable."
           fi
 
-      # A failed night must still save its translation progress. Without this, a
-      # from-scratch backfill (the ru/ar retranslation) can never converge: the
-      # run rate-limits partway, enforce/validate correctly refuse the partial
-      # result, the success()-gated publish above skips, and the night's OpenAI
-      # spend is discarded — so the next night restarts from the same point
-      # forever. This step commits the partial output to the same stable branch
-      # the "Restore partial progress" step reads, so consecutive nights
-      # accumulate until a run finally completes and publishes normally.
-      #
-      # Merging stays gated exactly as before: this step never creates a PR, and
-      # if the automation PR is already open its own Validate Translations check
-      # goes red on the partial head, so nobody can merge a state main's
-      # validators reject. Only what gets REMEMBERED changes, not what gets
-      # published.
+      # A failed night still saves its translation progress: without this, a
+      # from-scratch backfill can never converge — a rate-limited run fails
+      # enforce/validate, the success()-gated publish skips, and the spend is
+      # discarded. Saves to the same branch the restore step reads, so nights
+      # accumulate. Merging stays gated: never creates a PR, and the automation
+      # PR's own Validate Translations check is red on a partial head.
       - name: Save partial progress on failure
         if: failure()
         run: |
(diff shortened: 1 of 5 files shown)


.github/workflows/i18n-update-hub.yml
```diff
@@ -287,3 +287,68 @@ jobs:
...
+          done < <(git diff --cached --name-only --diff-filter=d)
+          if git diff --cached --quiet; then
+            echo "Run failed before producing any translation changes; nothing to save."
+            exit 0
+          fi
+          # Same no-op guard as the publish step: an identical rerun (e.g. a
+          # night that failed before translating) must not force-push a new head
+          # and rerun the automation PR's CI for zero progress.
+          if git rev-parse --verify --quiet "refs/remotes/origin/$STABLE_BRANCH" >/dev/null; then
+            if git diff --quiet "refs/remotes/origin/$STABLE_BRANCH" -- site/src/i18n; then
```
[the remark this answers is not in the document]
mobeenabdullah: Real. Reproduced in a scratch-repo sim: with the old restore, a banked ru.json regenerated as malformed vanished from the stable branch; with the restore-chain it keeps the last good version and an identical rerun still no-ops via the --cached compare. Fixed in 5a23386.

.github/workflows/i18n-update-hub.yml
```diff
@@ -287,3 +287,82 @@ jobs:
...
+                  # drop off the stable branch at commit time. Take the last
+                  # good version from the branch when it has one, else main's,
+                  # else drop the new file entirely.
+                  if git cat-file -e "refs/remotes/origin/$STABLE_BRANCH:$f" 2>/dev/null; then
+                    git restore --source="refs/remotes/origin/$STABLE_BRANCH" --staged --worktree -- "$f"
+                  elif git cat-file -e "HEAD:$f" 2>/dev/null; then
+                    git restore --source=HEAD --staged --worktree -- "$f"
+                  else
+                    git rm -qf --ignore-unmatch -- "$f"
+                  fi
```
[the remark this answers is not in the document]
mobeenabdullah: Half right, split it:

Content: not an issue. enforce runs before this step, prunes with the validator's own collectViolations (it imports them), and writes pruned files to disk even when its systemic check later fails the run, and it re-prunes restored branch content every night. A pruned field is absent, so the next run retranslates it. Field-level invalid content cannot be banked or persist.

UI chrome: real. Nothing prunes locales/ and locale:ui fills only missing keys, so a parseable-but-placeholder-broken UI string banked here would restore and fail validate forever. Fixed in b35f4cd: the failure path never banks locales/, it keeps the branch's (else main's) last good UI state. Reproduced both the persistence and the fix in a scratch-repo sim.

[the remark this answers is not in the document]
mobeenabdullah: The one remaining case is handled in 09587c7: a fully misaligned content file (every key absent from the English source) is excluded via the same fallback chain; partially stale files still bank, matching the validator's tolerance. Verified in the sim: misaligned file excluded while same-night valid progress banks, banked versions preserved.

.github/workflows/i18n-update-hub.yml
```diff
@@ -287,3 +287,100 @@ jobs:
...
+      # accumulate. Merging stays gated: never creates a PR, and the automation
+      # PR's own Validate Translations check is red on a partial head.
+      - name: Save partial progress on failure
+        if: failure()
+        run: |
+          set -euo pipefail
+          STABLE_BRANCH="i18n/hub-translations"
+          git config user.name 'github-actions[bot]'
+          git config user.email 'github-actions[bot]@users.noreply.github.com'
+          git add site/src/i18n
```
MaanilVerma, line 304: question: this step runs under `set -euo pipefail` and its first real action is `git add site/src/i18n`, but the whole point of the step is to run when an earlier step failed, and one of the ways it fails is before that directory ever gets written. If the run dies in checkout or in the "Build English content source" step, `site/src/i18n` may not exist on disk, `git add` on a missing pathspec exits non-zero, and `-e` kills the handler right here, so the failed night saves nothing, which is exactly the case this step exists to cover. The no-op guard further down (`git diff --cached --quiet ... nothing to save`) shows you already expect the empty case, you just abort before reaching it. Could we guard the add (`git add site/src/i18n 2>/dev/null || true`, or skip when the dir is absent) so an early-failure night still reaches the no-op exit cleanly instead of erroring out of its own save path?

.github/workflows/i18n-update-hub.yml
```diff
@@ -287,3 +287,100 @@ jobs:
...
+              echo "No change vs the saved partial progress; leaving the branch untouched."
+              exit 0
+            fi
+          fi
+          git checkout -B "$STABLE_BRANCH"
+          git commit -m "chore(i18n): save partial translation progress (run failed, do not merge)"
+          # Retry transient push failures (a GitHub-side 500 here costs a full
+          # night's progress).
+          for attempt in 1 2 3; do
+            if git push --force-with-lease origin "$STABLE_BRANCH"; then
```
MaanilVerma, line 380: question: the retry loop pushes with `git push --force-with-lease origin "$STABLE_BRANCH"`, and a bare `--force-with-lease` checks the local remote-tracking ref `refs/remotes/origin/i18n/hub-translations` before it will push. On the very first failed night the branch doesn't exist yet, so the restore step's fetch fell through its `if` without ever creating that ref, and `--force-with-lease` against a missing lease ref refuses with "stale info" rather than creating the branch. All three attempts hit the same refusal, the loop sleeps 20s then 40s, and the night ends on `::error::Could not push partial progress` with nothing banked, so the first backfill night, the one you most want to capture, is the one that can't save. Could we use a plain create push when the ref is absent and keep `--force-with-lease` only once `origin/$STABLE_BRANCH` has been fetched, so the first run seeds the branch and later runs still get the lease protection?

.github/workflows/i18n-update-hub.yml
```diff
@@ -287,3 +287,100 @@ jobs:
...
+          # English source) fails validation and enforce cannot prune it, so
+          # banking it would carry the mangled keys forward. Partially stale
+          # files still bank, matching the validator's tolerance.
+          while IFS= read -r f; do
+            case "$f" in
+              site/src/i18n/content/en.json) ;;
+              site/src/i18n/content/*.json)
+                if ! node -e '
+                  const fs = require("fs");
+                  const en = JSON.parse(fs.readFileSync("site/src/i18n/content/en.json", "utf8"));
```
MaanilVerma, line 342: suggestion (non-blocking): this misalignment check reads `content/en.json` fresh from the worktree for every locale it inspects, and it's correct only because the unparseable-JSON loop above already ran and, if English came out truncated, restored en.json to the branch's or main's good version first. So the safety of this block depends entirely on loop order: the parse-validity pass has to run before the alignment pass, or this reads a mangled en.json, finds every locale "aligned" against garbage, and banks the mangled set. Nothing in the code says these two passes are ordered on purpose, so a future refactor that merges them or flips them would reintroduce the poisoning quietly. Could we drop a line noting that en.json is already sanitized by the parse pass above, so the ordering that makes this safe survives the next edit?

site/.i18nrc.cjs
```diff
@@ -103,6 +103,12 @@ module.exports = defineConfig({
   // Serial: paced for the org's OpenAI tier (30k TPM). Raise once the tier is bumped.
   concurrency: 1,
   saveImmediately: true,
+  // Force OpenAI's JSON mode (response_format: json_object). gpt-5.2 in
+  // plain-text mode appends a stray closing brace after otherwise valid JSON
+  // (miscounting the {{PTnn}} placeholder braces); lobe rejects the chunk, which
+  // froze every locale's translation run from 2026-08-14 on. JSON mode makes the
+  // API guarantee parseable output. A model rejecting the flag fails with a 400.
+  experimental: { jsonMode: true },
```
MaanilVerma, line 111: suggestion (non-blocking): `experimental: { jsonMode: true }` only helps if lobe actually forwards `response_format: json_object` for this option name and the configured model accepts it, and the comment says as much ("lobe only sends the flag when this experimental option is set"). The failure mode if either half is wrong is quiet: a lobe version that renamed or dropped the experimental flag just goes back to plain-text mode and the stray-brace freeze returns with no signal, since a silently-ignored config key doesn't 400 the way a rejected `json_object` would. Could we pin the lobe/i18n-cli version this option name is verified against (or add a one-line startup assertion that the flag reached the request) so a dependency bump can't silently drop us back onto the path that froze every locale from 2026-08-14?

.github/workflows/i18n-update-hub.yml
mobeenabdullah: Checked in a scratch repo: site/src/i18n is committed, so the pathspec always matches the index even when a step wipes it from disk entirely; git add then stages the deletions and exits 0. It errors only when the pathspec matches neither index nor disk, which for a committed directory means checkout itself failed, and then the step's own git config dies first with nothing on disk to save anyway. Runs 32225067559/32227947002/32230677735 also exercised this step live. Leaving it unguarded on purpose: an unconditional || true would relabel real add failures (index lock, disk full) as a clean 'nothing to save'.

.github/workflows/i18n-update-hub.yml
mobeenabdullah: This one is disproven by the live test: run 32227947002 created i18n/test-partial-save through this exact retry loop with no remote-tracking ref (run 1's push had failed, so the restore fetch found nothing), and a scratch-repo repro agrees: bare --force-with-lease with neither the tracking ref nor the remote branch present creates the branch. The refusal you describe needs the tracking ref absent while the remote branch EXISTS, which here only happens when the restore fetch failed transiently, and refusing then is the lease doing its job: a plain create-push in that state would overwrite banked progress we just failed to restore. Keeping the lease on all attempts.

.github/workflows/i18n-update-hub.yml
mobeenabdullah: Real ordering dependency, agreed. Noted in 4fe5acd: the misalignment pass reads en.json from the worktree and must run after the parse pass, which has already restored a truncated en.json to a good version.

site/.i18nrc.cjs
mobeenabdullah: Took the assertion route in 4fe5acd, matching how check-config already reads the installed bundle for its context-window table: i18n:check now fails before any spend if the config drops experimental.jsonMode or the installed CLI stops reading the flag / sending json_object (unreadable bundle warns loudly, same stance as the table). Unit tests cover the committed config, the live bundle, and the dropped-option shape. Version stays ^1.26.0 with the lockfile pinning CI, so a deliberate bump gets caught by the nightly check instead of by a frozen night.

MaanilVerma (approved):

.github/workflows/i18n-update-hub.yml
MaanilVerma: You're right, I missed that site/src/i18n is tracked, so the pathspec always matches the index and add stages deletions instead of erroring. And keeping it unguarded so a real index-lock or disk-full failure still surfaces is the better call. Withdrawing this.

.github/workflows/i18n-update-hub.yml
MaanilVerma: The live run settles it: with neither the tracking ref nor the remote branch present, bare --force-with-lease creates the branch, so the first night saves fine. My refusal case needs the remote branch to exist while the tracking ref is missing, and you're right that refusing there is the lease protecting banked progress we just failed to restore. Withdrawing this.

.github/workflows/i18n-update-hub.yml
MaanilVerma: The note in 4fe5acd covers it, thanks. That's exactly the ordering constraint I was worried a future edit would drop.

site/.i18nrc.cjs
MaanilVerma: The assertion is better than pinning, since it catches a silent CLI-side drop and not just a config edit, and failing before spend is the right place for it. Nice.

MaanilVerma (approved):
LGTM, approved. Two of my four were real and you've handled them (the en.json ordering note and the pre-spend jsonMode assertion), and you disproved the other two with live runs I've verified, so I'm satisfied on the correctness side.

One non-blocking thing for a follow-up, not this PR: the comment density in the save step is getting heavy, the prose is close to the volume of the code it explains. The failure-mode reasoning is genuinely load-bearing so I wouldn't strip it, but some of it now duplicates what the coderabbit threads and the config-check assertions already establish, so a later pass could thin the justification paragraphs down to the one or two constraints a future editor would actually break.

site/scripts/i18n/cli-context-windows.ts
```diff
@@ -25,14 +25,36 @@ const require = createRequire(import.meta.url);
...
+/**
+ * Whether a CLI bundle both reads the `experimental.jsonMode` config option and
+ * forwards OpenAI's `json_object` response format. Both markers are property
+ * names / wire literals, which minification leaves intact. A version that
+ * renames or drops the option would silently fall back to plain-text mode,
+ * where gpt-5.2's stray-brace output froze every locale from 2026-08-14 — an
+ * ignored config key gives no signal, unlike a rejected response_format.
+ */
+export function cliSourceSupportsJsonMode(source: string): boolean {
+  return source.includes('jsonMode') && source.includes('json_object');
```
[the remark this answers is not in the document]
mobeenabdullah: Took the substance, adapted the mechanism in 2e482e0. The check now matches the wiring expressions, the .experimental?.jsonMode property access plus the response_format json_object literal inside a chat.completions.create argument span, so unwired occurrences of the two strings cannot satisfy it (test added for exactly that shape). The full interception is not viable here: the package's lib entry exports only defineConfig, the translation class lives solely in the CLI bundle, and importing that bundle has side effects (ink/commander/update-notifier top-level imports), so instance surgery on unexported minified internals would be more fragile than the check it replaces. Residual risk is bounded: if wiring ever broke despite matching expressions, the nightly still fails loudly at validate, costing nights of spend but never publishing bad output.

PR #1163, opened 2026-08-19
fix(i18n): stop the automation branch reverting human-authored sources

mobeenabdullah:
## The bug, live right now

The open automation PR proposes to **delete all 13 `template.api.*` strings** from `site/src/i18n/locales/en.json` — the UI labels for the API section that merged today. Same mechanism took the navbar keys on 2026-08-12; the branch was patched then, the pipeline was not.

## Cause

"Restore partial progress from automation branch" does:

```
git checkout "origin/$STABLE_BRANCH" -- site/src/i18n
```

That shadows **every** file under `site/src/i18n` with the automation branch's copy, including the layers a person edits. Anything landed on main after the branch last updated comes back as a deletion in the next automation PR, and the branch is force-pushed nightly, so it persists until someone notices.

## Fix

After the restore, take the person-edited layers back from the checkout:

- `locales/en.json` — load-bearing. Nothing regenerates it; `.i18nrc.ui.cjs` reads it as its entry and only ever writes the other locales.
- `overrides/` — load-bearing. Hand-authored (native-review corrections), highest precedence in the resolver, never written here.
- `reviews/` — written here, but entries missing from a stale copy get re-stamped as machine-authored, silently downgrading a human wave's provenance.
- `human/` — belt-and-braces. `build-content-source` rewrites it from the templates index on the very next step; kept for a run that fails before that.

`git restore` rather than `git checkout`, so a file deleted on main cannot be resurrected from the branch, and unguarded so the protection cannot silently no-op. The machine layers (other locales' content and UI strings, verdicts) still accumulate across runs, so rate-limited progress is unaffected.

## Verification

Scratch-repo repro, both directions: a key added to `en.json` on main and an edit to `overrides/zh.json` are both reverted by the current restore and both survive with this change, while a stale file the branch resurrects is correctly removed and the machine layer still restores.

After this merges, re-running the workflow rebuilds the automation branch on current main, which clears the conflict on the open translation PR.

### DIFF
--- a/.github/workflows/i18n-update-hub.yml
+++ b/.github/workflows/i18n-update-hub.yml
@@ -84,7 +84,19 @@ jobs:
           # have a real base. Absent branch (first run) -> fetch fails, start fresh.
           if git fetch origin "$STABLE_BRANCH:refs/remotes/origin/$STABLE_BRANCH" 2>/dev/null; then
             git checkout "origin/$STABLE_BRANCH" -- site/src/i18n || true
-            echo "Restored partial translations from $STABLE_BRANCH."
+            # The restore shadows EVERY file under site/src/i18n, including the
+            # ones this pipeline does not author. Any key a person adds to the
+            # English UI source (or to the human translation layer) on main after
+            # the automation branch was last updated would otherwise come back
+            # deleted in the next automation PR: that is how the navbar keys were
+            # lost on 2026-08-12, and the API-section strings were about to go the
+            # same way. Take those back from main; the machine layers this
+            # pipeline owns (other locales, verdicts, sign-off records) still
+            # accumulate across runs.
+            git checkout HEAD -- \
+              site/src/i18n/locales/en.json \
+              site/src/i18n/human 2>/dev/null || true
+            echo "Restored partial translations from $STABLE_BRANCH (human-authored sources kept from main)."
           else
             echo "No automation branch yet; starting fresh from main."
           fi
--- a/.github/workflows/i18n-update-hub.yml
+++ b/.github/workflows/i18n-update-hub.yml
@@ -84,15 +84,18 @@ jobs:
           # have a real base. Absent branch (first run) -> fetch fails, start fresh.
           if git fetch origin "$STABLE_BRANCH:refs/remotes/origin/$STABLE_BRANCH" 2>/dev/null; then
             git checkout "origin/$STABLE_BRANCH" -- site/src/i18n || true
-            # The restore above shadows every file under site/src/i18n, including
-            # the human-authored ones this pipeline does not write, so a key added
-            # to the English UI source since the branch last updated comes back
-            # deleted (it took the navbar keys on 2026-08-12). Take those paths
-            # back from the checkout; restore rather than checkout so a file
-            # deleted on main cannot be resurrected from the branch, and unguarded
-            # so this can never silently no-op.
+            # The restore above shadows every file under site/src/i18n with the
+            # branch's older copy, so an edit landed on main since the branch was
+            # last updated comes back as a deletion (it took the navbar keys on
+            # 2026-08-12). Take the layers a person edits back from the checkout:
+            # the UI source nothing regenerates, the override layer nothing here
+            # writes, and the sign-off records, whose missing entries this pipeline
+            # would otherwise re-stamp as machine-authored. restore, not checkout,
+            # so a file deleted on main cannot be resurrected from the branch.
             git restore --source=HEAD --staged --worktree -- \
               site/src/i18n/locales/en.json \
+              site/src/i18n/overrides \
+              site/src/i18n/reviews \
               site/src/i18n/human
             echo "Restored partial translations from $STABLE_BRANCH (human-authored sources kept from the checkout)."
           else
--- a/.github/workflows/i18n-update-hub.yml
+++ b/.github/workflows/i18n-update-hub.yml
@@ -84,19 +84,17 @@ jobs:
           # have a real base. Absent branch (first run) -> fetch fails, start fresh.
           if git fetch origin "$STABLE_BRANCH:refs/remotes/origin/$STABLE_BRANCH" 2>/dev/null; then
             git checkout "origin/$STABLE_BRANCH" -- site/src/i18n || true
-            # The restore shadows EVERY file under site/src/i18n, including the
-            # ones this pipeline does not author. Any key a person adds to the
-            # English UI source (or to the human translation layer) on main after
-            # the automation branch was last updated would otherwise come back
-            # deleted in the next automation PR: that is how the navbar keys were
-            # lost on 2026-08-12, and the API-section strings were about to go the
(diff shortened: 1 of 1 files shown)


.github/workflows/i18n-update-hub.yml
```diff
@@ -84,7 +84,19 @@ jobs:
...
+            # the automation branch was last updated would otherwise come back
+            # deleted in the next automation PR: that is how the navbar keys were
+            # lost on 2026-08-12, and the API-section strings were about to go the
+            # same way. Take those back from main; the machine layers this
+            # pipeline owns (other locales, verdicts, sign-off records) still
+            # accumulate across runs.
+            git checkout HEAD -- \
+              site/src/i18n/locales/en.json \
+              site/src/i18n/human 2>/dev/null || true
+            echo "Restored partial translations from $STABLE_BRANCH (human-authored sources kept from main)."
```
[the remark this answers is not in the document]
mobeenabdullah: Two of the three taken, in 76be3b2.

Overlay: real, and reproduced. With git checkout's pathspec overlay, a human-layer file deleted on main but still present on the automation branch survives the restore and gets published; git restore matches the checkout exactly and removes it. Switched.

Guard: agreed, dropped. git restore fails loudly on a pathspec the source lacks, and a protection that can silently no-op is worse than one that stops the run. Left the pre-existing guard on the branch restore above: its failure mode is starting fresh from main, which is safe, while this one's was the clobber itself.

MAIN_REF: not taking it. HEAD is the run's own checkout, which is what every other step operates on, including the publish step that builds the automation branch. Pinning this one line to origin/main would mix main's UI source into a run based on another ref. Scheduled runs check out main, so HEAD is main in the normal case.

.github/workflows/i18n-update-hub.yml
```diff
@@ -84,7 +84,17 @@ jobs:
...
+            # The restore above shadows every file under site/src/i18n, including
+            # the human-authored ones this pipeline does not write, so a key added
+            # to the English UI source since the branch last updated comes back
+            # deleted (it took the navbar keys on 2026-08-12). Take those paths
+            # back from the checkout; restore rather than checkout so a file
+            # deleted on main cannot be resurrected from the branch, and unguarded
+            # so this can never silently no-op.
+            git restore --source=HEAD --staged --worktree -- \
+              site/src/i18n/locales/en.json \
+              site/src/i18n/human
```
MaanilVerma, line 99: question: this restores en.json and human/ back from HEAD, but the same restore step still shadows site/src/i18n/overrides/, which is the same kind of human-authored source. resolver.ts reads overrides/{locale}.json as a hand-authored layer (its comment at line 51 lists it right alongside human/ and reviews/), overrides/zh.json on main is hand-written zh content, and nothing in this workflow writes that path, so it behaves exactly like en.json and human/ do. So the day someone edits overrides/zh.json or adds overrides/ja.json after the branch last updated, the next automation PR proposes deleting it, the same navbar/api recurrence this PR exists to stop, just one directory over. Could we add site/src/i18n/overrides to the restore --source=HEAD list (and cover it in the scratch-repo repro so a new overrides file on main survives a stale branch) so every human-authored layer is protected, not just two of the three?

.github/workflows/i18n-update-hub.yml
```diff
@@ -84,7 +84,17 @@ jobs:
...
             git checkout "origin/$STABLE_BRANCH" -- site/src/i18n || true
-            echo "Restored partial translations from $STABLE_BRANCH."
+            # The restore above shadows every file under site/src/i18n, including
+            # the human-authored ones this pipeline does not write, so a key added
+            # to the English UI source since the branch last updated comes back
+            # deleted (it took the navbar keys on 2026-08-12). Take those paths
+            # back from the checkout; restore rather than checkout so a file
+            # deleted on main cannot be resurrected from the branch, and unguarded
+            # so this can never silently no-op.
+            git restore --source=HEAD --staged --worktree -- \
```
MaanilVerma, line 95: suggestion (non-blocking): the en.json half of this is the load-bearing part, since .i18nrc.ui.cjs reads locales/en.json as its entry and only ever writes the other locales, so nothing in the pipeline regenerates it and restoring it from HEAD is what actually saves it. The human/ half is more belt-and-suspenders than the PR body reads: build-content-source runs as the very next step and rewrites human/{locale}.json from templates/index.{locale}.json (writeFileSync at build-content-source.ts:280), so whatever the restore puts there gets regenerated a moment later from the templates index anyway. It's not wrong to keep it, a run that fails before build-source still ends up with the right human/ on disk, but could we adjust the body's framing so a future maintainer doesn't read human/ as hand-maintained-and-fragile when build-source owns it, and treat overrides/ (which really is fragile) as the one that needs the same guard?

.github/workflows/i18n-update-hub.yml
mobeenabdullah: Right. Added overrides/ in 575c6ad, plus reviews/: a record missing from a stale copy gets re-stamped by signoff-refresh as automated, so a human wave's attribution silently degrades.

Repro note: the hazard is an EDIT to overrides/zh.json, not a new file. A new file on main survives already, since checkout with a pathspec is an overlay and cannot delete what the branch lacks. Sim covers the edit case.

.github/workflows/i18n-update-hub.yml
mobeenabdullah: Correct, build-content-source rewrites human/ from the templates index on the next step. Keeping it in the list for the fail-before-build-source case, PR body reframed: en.json and overrides/ are load-bearing, human/ is belt-and-braces.

.github/workflows/i18n-update-hub.yml
MaanilVerma: 575c6ad covers it, and catching reviews/ on top is the right call, a stale copy missing a record getting re-stamped by signoff-refresh degrades a human wave's attribution silently, same class of bug. Good distinction keeping review/ (machine verdicts, accumulates) out of the list while protecting reviews/ (the attribution layer the resolver reads). And you're right that it's the edit case, not a new file, since the checkout pathspec is an overlay that can't delete what the branch lacks, so the sim covering the edit is the one that matters.

.github/workflows/i18n-update-hub.yml
MaanilVerma: Reframing to en.json and overrides/ load-bearing, human/ belt-and-braces reads right, and keeping human/ for the fail-before-build-source case is a fair reason to leave it in the list. Thanks.

MaanilVerma (approved):
LGTM, approved. The core fix is right: nothing in the pipeline regenerates locales/en.json (.i18nrc.ui.cjs only reads it as entry), so restoring it from HEAD is what actually stops the clobber. Both my comments are handled in 575c6ad, and catching reviews/ on top of overrides/ (while correctly leaving the machine review/ verdicts out of the list) closes the whole human-authored surface, not just the two layers I flagged.

PR #1164, opened 2026-08-19
fix(site): repair variant 301s, empty categories and sitemap drift

mobeenabdullah:
Six independent indexability defects found while auditing GTM-349, plus a seventh found in review. Each was verified against the live site before the fix, and against a running server plus a real build after.

## 1. Every English model variant URL 404s

`/workflows/model/wan2-5/`, `wan2-2/` and `flux-2/` all return 404 on production today, while `/es/workflows/model/wan2-5/` correctly 301s to the family page.

The `redirects` map in `astro.config.mjs` cannot work, and no amount of configuring fixes it. `@astrojs/vercel@9.0.4` builds each rule's `source` from parsed route segments (`dist/lib/redirects.js`, `getRedirects` → `getMatchPattern`), and a trailing slash is not a segment. So both keyed forms collapse to one slashless rule, while the platform canonicalises every incoming request to the slashed form first. The rule never matches a real request.

The map's own comment claims keying both slash forms restores parity. It does not, and Astro says so out loud on dev server startup:

```
The route "/workflows/model/wan2-5" is defined in both "/workflows/model/wan2-5"
and "/workflows/model/wan2-5/". A static route cannot be defined more than once.
A collision will result in a hard error in following versions of Astro.
```

So the map is removed and the English route resolves variants from `group.redirectFrom`, which is what the locale route has always done and is the only mechanism proven to work in production. That requires the route to render on demand.

**On-demand does not mean uncached.** `site/vercel.json` already sets `s-maxage=300, stale-while-revalidate=60` on `/workflows/:path*`, so these pages stay edge-cached. The catalog, family derivation and fallback pool are memoized in `src/lib/workflow-pages/model-route-cache.ts` so a request does not re-serialize ~600 hub entries. They live in a module rather than in the route's frontmatter because a `let` in frontmatter is not module scope — the Astro compiler leaves it inside the render function, so it is rebuilt per request. Verified by compiling a probe against this repo's `@astrojs/compiler@2.13.0`. No added staleness either way: `listWorkflowIndex` already caches the hub response at module level for the life of the process.

## 2. The build-time guards were validating a catalog the routes never serve

Raised in review, and the measurement made it a defect rather than a gap.

Brand safety, orphaned content and slug uniqueness ran inside the English route's `getStaticPaths`, so they had to move when that route went on-demand. They moved to `astro.config.mjs` — which derived its groups from `src/content/templates`, the repo's own catalog synced from `templates/index.json`. Both model routes resolve a request from `loadSerializedTemplates()`: the live hub index. The two are not the same catalog, and measuring them apart shows it is not close:

```
hub groups 81                 repo groups 120
hub variant aliases 41        repo variant aliases 48

variant only in the hub  : kling-1-6, kling-2-0
variant only in the repo : flux-3-video, gpt-image-2, ltx-2-5, qwen-image-3-0-pro,
                           qwen-image-edit-2511, seedance-2-5, seedream-5-0-pro,
                           wan-animate, wan3-0
qualifying only in the hub : minimax, elevenlabs, google-gemini, recraft, magnific,
                             vidu, google-gemini-image, hidream
```

So `assertUniqueModelSlugs` was checking 48 aliases no route resolves while missing 2 that it does, and `assertBrandSafe` was skipping 8 qualifying families the routes actually serve. The guards ran, went green, and guarded the wrong thing.

`astro.config.mjs` still must not fetch the hub itself — it is loaded by `astro check`, `astro dev`, eslint and every unit-test run. So prebuild writes the catalog to a manifest and the config derives from that: same `deriveModelGroups`, same data, one process later. The sitemap's model list comes off the same groups, so it advertises what the route resolves rather than what the repo happens to contain. Today the two agree on the indexable set exactly (11 slugs either way), so this changes no sitemap output — only what the guards can see.

## 3. The video and 3d category pages are empty, and the sitemap said otherwise

`serializeIndexEntry` defaults a missing `mediaType` to `'image'`. On the live index, **178 of 620 entries carry no `mediaType` at all**, and 89 of those are video workflows (`Seedance2.5: Text to Video`, `LTX-2.5: FLF2V`). Filing them as images leaves `/workflows/category/video/` and `/workflows/category/3d/` with nothing in them, which 404s their localized twins, because the locale routes 404 an empty result while the English routes render a blank page.

**This PR does not fix that, and deliberately no longer tries.** An earlier revision inferred the media type from tags. Two reviewers were right that classification is backend-owned state and does not belong in the site mapper, so it is gone: `serializeIndexEntry` reads only what the index declares, and `inferMediaType` on the detail path is byte-identical to main. Category membership on this branch is exactly what it was before it.

What the branch does instead is stop the sitemap lying about it. `scripts/build-hub-manifests.ts` records which categories the index actually fills, and both the localized URLs and the English ones are gated on that list. Today it yields image and audio, so `/workflows/category/video/` and `/3d/` leave the sitemap rather than sitting in it as advertised empties. When the backend populates the field the list follows on its own, with no sitemap change.

**The remaining defect is upstream, and criterion 3 of GTM-349 is blocked on it.** Beyond the 178 blanks, 198 entries positively declare `mediaType: image` while their tags read as video. A backfill has to correct existing values, not only fill missing ones, and the field has to become required in the response schema. There is no generated client in this repo to consume that from — `HubWorkflowTemplateEntry` is handwritten and every field on it is optional — so that half cannot land here. Tracked separately against the cloud repo.

## 4. The sitemap disagreed with the pages, in both directions

- Nine gated locale roots were advertised while those pages render `noindex` and canonical to English.
- No launched locale's category, tag or model pages were listed at all, though they self-canonical and invite indexing.

Both from building the list off every supported locale instead of `INDEXABLE_LOCALES`. `listing-indexing.ts` exists to keep canonical, robots, hreflang and the sitemap saying one thing; the sitemap was the surface not reading the flag.

Those nine were flipped indexable by #1193 while this branch was in review, so the first bullet no longer has a live example: the sitemap this branch builds today lists all ten flipped locales and withholds `it`, which is a supported hub locale that has not been flipped. The mechanism is what changed — the list now follows the flag, whatever the flag says.

English model pages are now listed explicitly too. Once that route stopped prerendering, the sitemap integration could no longer discover it, and the 11 indexable model URLs would have vanished silently.

The construction moved out of `astro.config.mjs` into `src/lib/sitemap-custom-pages.ts` so it can be unit tested against real inputs, with the manifest reads in `src/lib/hub-manifests.ts` beside it.

## 5. Localized tag URLs came from the wrong catalog

Same root cause as item 2, found first. `src/content/templates` is synced from this repo's own `templates/index.json`; the tag routes call `loadSerializedTemplates`, which reads the hub index. Deriving the sitemap's tag list from the synced files was wrong in both directions at once:

- **25 real tags omitted** (`sprite-sheet`, `character`, `product`, `sound-effects`, …), so their localized pages were never advertised.
- **4 slugs invented** — `character-replacement`, `element-segmentation`, `int8`, `reference-to-video` exist in the repo's templates and match no hub workflow. `[locale]/workflows/tag/[tag].astro` 404s an empty filter, so the sitemap would have advertised four 404s per indexable locale.

`scripts/build-hub-manifests.ts` runs during prebuild, next to the search index, and writes the gitignored manifests the config reads.

There is deliberately **no fallback to the synced files**. If a manifest is absent the sitemap emits no hub-derived URLs at all and the model guards do not run: missing URLs and a skipped guard are recoverable, an advertised 404 and a guard validating the wrong catalog are not. A configured hub that fails is still a hard build error, matching `loadSerializedTemplates`. And a refresh that cannot complete **removes** the previous run's manifests before giving up, degraded or fatal, so a reused checkout cannot build a sitemap from a hub snapshot of unknown age.

## 6. The sitemap ↔ indexability check went blind, and red

`scripts/verify-sitemap-indexability.ts` is a blocking CI gate. It answered "is this model page indexable?" by reading the built page's robots meta — which stops existing the moment the route renders on demand. It reported 0 indexable model pages and failed all 11.

The model section now recomputes the verdict with `isModelPageIndexable`, the same helper the route calls, over families derived from the catalog the route derives them from: the live hub index.

**That dataset is the whole point, and an earlier revision of this branch got it wrong.** It derived families from `templates/index.json` — the committed catalog the sitemap's own snapshot is a sibling of — so both sides read the same files and the check could no longer fail on the one drift its own header names. Reading the prerendered HTML used to compare the two datasets by accident, because `getStaticPaths` built those pages from hub data. Raised in review; the fix restores the comparison rather than the appearance of one.

The hub fetch is the only network call, and it degrades rather than fails: a gate that blocks merges must not go red because the hub was down. When it degrades it says so (`model: DEGRADED — ... drift cannot be detected in this run`) and falls back to `templates/index.json` — the only catalog present in a job that downloads the build artifact without ever running prebuild.

## 7. Duplicate hreflang on EN tag and category pages

Every hreflang was emitted twice, once from `SEOHead.astro` and once from a page-level `HreflangTags`. Confirmed the two sets were byte-identical on production before removing the page-level one, so output is unchanged apart from the duplication.

## Verification

Against a running dev server, so the 301s can only be coming from the route:

```
/workflows/model/wan2-5/             301 -> /workflows/model/wan/
/workflows/model/wan2-2/             301 -> /workflows/model/wan/
/workflows/model/flux-2/             301 -> /workflows/model/flux/
/workflows/model/wan/                200   canonical still renders
/workflows/model/not-a-real-model/   404   no over-redirecting
```

**The guards, against the catalog they now see.** Dropping the Wan family from the hub manifest and leaving the repo catalog untouched — which is exactly the hub-drops-a-family case — now fails the build:

```
Orphaned model landing content (no page renders these slugs): wan.
```

Before this it passed, because the repo catalog still had Wan.

**Real build, generated sitemap (8095 URLs):**

```
EN model URLs         11, all preserved      flux grok ideogram kling ltx
                                             nano-banana-pro qwen-image
                                             qwen-image-edit seedance wan z-image
EN + locale categories image, audio only     (video and 3d dropped: both render empty)
zh                    75 tag + 2 category + 11 model + 1 root
locales listed        the 10 in INDEXABLE_LOCALES; `it` is built but absent
invented tag URLs     0   (int8, character-replacement, element-segmentation,
                           reference-to-video all absent, per locale)
```

The gate, against that build:

```
pnpm verify:sitemap-indexability
  model: 11 in sitemap, 11 indexable
  use-cases: 17 in sitemap, 17 indexable
  Sitemap membership matches the rendered-indexable set.
```

And its degraded path, forced with an unreachable hub — announced, not silent, and still a real answer:

```
  hub index unreachable (TypeError: fetch failed)
  model: DEGRADED - hub index unavailable, families derived from templates/index.json.
         Drift between the sitemap and the live catalog cannot be detected in this run.
  model: 11 in sitemap, 11 indexable
```

**The manifest refresh's own degraded path**, with all three files seeded and the hub unreachable:

```
Error: Hub API failed while building the hub manifests: TypeError: fetch failed
after: (all three removed)
```

`tests/unit/hub-manifests-degradation.test.ts` pins that, and was confirmed red with the removal commented out:

```
AssertionError: hub-tag-slugs.generated.json survived a failed refresh
```

Gate: eslint clean, `astro check` 0 errors 0 warnings, vitest **770 passing**, prettier clean on every file this PR touches, verified with the lockfile's prettier.

## Note for reviewers

Nothing here is visible on comfy.org until someone runs the manual **Deploy Template Site** dispatch.


site/astro.config.mjs
```diff
@@ -49,6 +52,45 @@ if (fs.existsSync(templatesDir)) {
...
+        slug: group.slug,
+        primaryKeyword: group.keywords.primary,
+        title: group.label,
+        secondaryKeywords: group.keywords.secondary,
+      });
+    }
+    if (claimedModelSlugs.has(group.slug)) {
+      throw new Error(`Duplicate model slug "${group.slug}" across families; slugs must be unique.`);
+    }
+    claimedModelSlugs.add(group.slug);
```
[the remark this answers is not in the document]
mobeenabdullah: Taken. `assertUniqueModelSlugs` in `model-groups.ts` now rejects a variant claimed by two families and a variant that is also another family's canonical slug, in two passes so a collision with a family derived later is still caught. `astro.config.mjs` calls it; the canonical-only loop is gone.

The second rule was also a regression this branch introduced. The deleted `redirects` map carried `.filter((variant) => !canonicalModelSlugs.has(variant))` precisely so a redirect could not shadow a real page, and nothing replaced it when resolution moved into the route.

Neither rule fires on current data — checked against both the synced catalog and the live hub index, so this is a guard, not a fix. Tests in `tests/unit/model-slug-uniqueness.test.ts`.

0d3c3b6a

site/astro.config.mjs
```diff
@@ -100,8 +134,64 @@ const indexableLocales = new Set(INDEXABLE_LOCALES);
...
+const tagSlugs = [
+  ...new Set(
+    contentTemplates.flatMap((tmpl) =>
+      (tmpl.tags || [])
+        .filter((tag) => typeof tag === 'string' && tag.trim())
+        .map((tag) => tagSlug(tag))
+        .filter(Boolean)
+    )
+  ),
+];
```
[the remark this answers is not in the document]
mobeenabdullah: Right, and measuring it turned this from a gap into a defect.

Against production: the synced files yield 54 tag slugs, the hub index 75. So 25 real tags were missing. Worse, 4 of the files' slugs match no hub workflow at all — `character-replacement`, `element-segmentation`, `int8`, `reference-to-video`. `[locale]/workflows/tag/[tag].astro` filters the hub catalog and 404s an empty result, so this branch would have advertised four 404s per indexable locale. Confirmed against a running server:

```
/zh/workflows/tag/int8/                   404
/zh/workflows/tag/character-replacement/  404
/zh/workflows/tag/sprite-sheet/           200   hub-only, was missing
/zh/workflows/tag/product/                200   hub-only, was missing
```

Fixed with the shared manifest you suggested. `scripts/build-tag-manifest.ts` runs in prebuild beside the search index and writes a gitignored `src/data/hub-tag-slugs.generated.json`; `astro.config.mjs` reads that. The config still never fetches, so `astro check`, `astro dev` and lint stay offline.

No fallback to the synced files: with no manifest the sitemap emits no localized tag URLs at all. A missing URL is recoverable, an advertised 404 is the contradiction this ticket exists to remove. A configured hub that fails is still a hard build error, matching `loadSerializedTemplates`.

zh now carries 75 tag URLs, and 0 of the invented ones. 0d3c3b6a

site/tests/unit/hub-api-media-type.test.ts
```diff
@@ -0,0 +1,58 @@
...
+
+  it('resolves to the output medium when two media appear in one tag', () => {
+    expect(typeOf(['API', 'Audio to Video'])).toBe('video');
+  });
+
+  it('infers audio and 3d', () => {
+    expect(typeOf(['API', 'Audio'])).toBe('audio');
+    expect(typeOf(['Text to Audio'])).toBe('audio');
+    expect(typeOf(['3D'])).toBe('3d');
+  });
```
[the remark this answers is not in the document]
mobeenabdullah: Added, plus `Audio to Audio` and `Image to 3D`, which are real tags the phrase-shaped matching has to get right too. 0d3c3b6a

mobeenabdullah:
Two things beyond the review threads, both found while verifying the round.

**The SEO Audit was red, and this branch caused it.** `scripts/verify-sitemap-indexability.ts` is a blocking gate that decided model indexability by reading the built page's robots meta. That page stops existing the moment the route renders on demand, so the check saw 0 indexable model pages and rejected all 11 sitemap entries. Not a flake.

The model section now recomputes the verdict with `isModelPageIndexable`, the same helper the route calls. It stays a real cross-check because the sitemap's own list comes from a different expression in `astro.config.mjs` (`qualifies && modelContentPasses`). It reads `templates/index.json` rather than the synced catalog, because the `seo-audit` job only checks out and downloads the build artifact and never runs prebuild — recomputing from `src/content/templates` would have failed there for a new reason. Green on a real build:

```
model: 11 in sitemap, 11 indexable
use-cases: 17 in sitemap, 17 indexable
```

**`sitemap-locale-custom-pages.test.ts` was asserting on `astro.config.mjs` source text**, e.g. `expect(body).not.toContain('nonDefaultLocales')`. That breaks on reformatting and proves nothing about behaviour. The construction moved to `src/lib/sitemap-custom-pages.ts` and the tests now call it with real inputs.

Two questions a reviewer is likely to have, answered up front:

- **Does on-demand rendering slow the model pages down?** `site/vercel.json` already sets `s-maxage=300, stale-while-revalidate=60` on `/workflows/:path*`, so they stay edge-cached. The route also memoizes its catalog, family derivation and fallback pool at module level — before this it re-serialized ~600 hub entries per request.
- **Does the SSR switch change what the sitemap advertises?** No. Derived the model families from the synced files and from the live hub index and compared: the indexable set is the same 11 either way, in both directions.

site/src/lib/hub-api.ts
```diff
@@ -351,7 +351,10 @@ export function serializeIndexEntry(
     shareId: entry.shareId || '',
     title: entry.title || entry.name,
     description: entry.description || '',
-    mediaType: entry.mediaType || 'image',
+    // Infer rather than default: the index omits mediaType on 174 of 616 live
+    // entries, and defaulting them to 'image' emptied the video and 3d category
+    // pages outright. A declared value always wins.
+    mediaType: entry.mediaType || mediaTypeFromTagNames(entry.tags || []),
```
dante01yoon: issue: `mediaType` is backend-owned workflow classification, but this fallback reconstructs it from human-facing tag labels using substring matching and precedence rules. That makes category membership depend on frontend heuristics: a tag rename/new phrase can silently reclassify a workflow, and `Audio to Video → video` is now business logic duplicated outside the API. The live defect is real—the index omits `mediaType` for many rows—but it should be fixed at the source: populate/backfill `mediaType` in `/api/hub/workflows/index`, make the field required in the OpenAPI/generated response type, then consume that value directly here. Could we remove tag-based inference from this mapper and land the backend contract/data fix first?

dante01yoon (changes requested):
Blocking on one contract-ownership issue. The production defect is verified, but media classification must be deterministic backend data rather than reconstructed in the site mapper.

site/src/lib/hub-api.ts
mobeenabdullah: Agreed on ownership: classification belongs in the API, and I want the field required in the generated type too.

Two corrections to the framing, both checkable.

**Tag inference is not new here.** `inferMediaType` is already on main and drives the detail path (`hub-api.ts:451` and `:473`), with the same declared-wins precedence (`meta.media_type || inferMediaType(workflow)`). This PR unified two copies into one exported function and extended it to the index, where the previous behaviour was `entry.mediaType || 'image'`. So the choice in front of us is one definition or two, not heuristics or none. If the principle is that no tag-derived classification ships, main already violates it on the detail path and that is a larger change than this PR.

**Inference does not fix the worst rows, which strengthens your case.** 198 index entries positively DECLARE `mediaType: 'image'` while their tags say video ("Seedance 2.0: Text to Video", "Wan2.7: Video Edit"). A declared value always wins, so those stay misfiled until the backfill. Only the API fixes them.

**Removing it is not a one-line revert, and that is my fault.** The sitemap emitted all four category types per locale unconditionally, so dropping inference leaves two of them advertised and 404ing, which is the exact defect this PR exists to remove. The tag URLs beside them were already derived from the hub index for that reason; categories were not. Fixed in af2688c6: categories are now gated on what the index actually fills. No change to today's output, and it makes the classification source swappable without touching the sitemap.

So either way works now, and it is your call:

1. Keep the fallback as a stopgap, delete it when the field is required. Categories stay 200 today.
2. Drop it from this PR. `/zh/workflows/category/video/` and `/3d/` keep 404ing until the API backfills, the sitemap correctly stops advertising them, and GTM-349's "200 in every locale" stays unmet in the meantime.

Say which and I will push it. Where should the backend ticket go so I can file the contract and backfill ask against it?

mobeenabdullah:
@dante01yoon this is waiting on your call, and it is now safe either way.

af2688c6 gates the localized category URLs on the categories the hub index actually fills, so dropping the inference no longer leaves the sitemap advertising two 404s per locale. That was the coupling that made your request more than a one-line revert.

So: keep the fallback as a stopgap and delete it when the field is required, or drop it now and accept that /zh/workflows/category/video/ and /3d/ keep 404ing until the API backfills. Say which and I will push it today.

Still need somewhere to file the contract and backfill ask, if you can point me at the right repo.

site/scripts/build-tag-manifest.ts
```diff
@@ -0,0 +1,90 @@
...
+
+  // The categories that actually have workflows, by the same rule the routes
+  // classify with. The category route 404s an empty type exactly as the tag
+  // route 404s an empty tag, so the sitemap has to be told which ones exist
+  // rather than assuming all four always do. This is also what makes the
+  // classification source swappable: when the hub populates `mediaType` on every
+  // entry, this list keeps following it and no sitemap change is needed.
+  const categories = [
+    ...new Set(
+      entries.map((entry) => entry.mediaType || mediaTypeFromTagNames(entry.tags || []))
```
dante01yoon: **Blocking:** This extends the unresolved tag-derived `mediaType` fallback into the sitemap manifest. Tags are content metadata, not an authoritative category contract, so the same workflow can be classified differently as tags change. Require `mediaType` from the backend/generated response type and generate categories from that field only; do not infer server-owned category state in this frontend build step.

dante01yoon (changes requested):
The original contract blocker remains unresolved on the new HEAD, and this follow-up now propagates the inferred classification into sitemap category generation. Category membership is backend-owned business state and must come directly from a required `mediaType` field in the API contract. Please remove tag-based media-type inference from both index serialization and manifest generation, make the backend response deterministic, and consume the generated shared response type rather than a handwritten optional contract.

The targeted mapper tests and `git diff --check` pass, but those checks preserve the duplicated frontend classification rather than validating the backend contract.

site/scripts/build-tag-manifest.ts
mobeenabdullah: Done in aa9bd98d. Tag inference is out of index serialization and out of the manifest; categories now come from declared `mediaType` only.

What that costs, so it is on the record and not discovered later: 174 of 616 index entries declare no `mediaType`, 89 of them video. They fall to the pre-existing `'image'` default, so `/zh/workflows/category/video/` and `/3d/` stay 404 until the API populates the field. That is acceptance criterion 3 of GTM-349 staying unmet, and it is now waiting on the backend rather than on this PR.

The category gate you flagged is what keeps that honest: it derives from declared data now, and it stops the sitemap advertising a category while it is empty. Without it, removing the inference would have left two advertised 404s per locale.

The test guards your contract rather than the heuristic: tags must not move a classification. It fails if inference is reintroduced.

**On the generated response type: I cannot do that here.** There is no generated client in this repo, no OpenAPI artifact, and no codegen step. `HubWorkflowTemplateEntry` is handwritten in `src/lib/hub-api.ts`, and every field on it is optional, not just this one. Making `mediaType` required in a handwritten type would only assert the guarantee, not obtain it, and it would fail to typecheck against an index that omits the field on 28% of rows.

So that half needs the API side: populate and backfill `mediaType` on `/api/hub/workflows/index`, make it required in the response schema, then generate the client here and consume it. Point me at the repo that owns that endpoint and I will open the issue with these numbers today. `mediaType` also has a second defect worth carrying into it: 198 entries positively declare `image` while their tags read as video, so a backfill has to correct existing values, not just fill missing ones.

mobeenabdullah: Resolved by aa9bd98d: this now reads declared `mediaType` only, no inference in the manifest.

Worth noting what the gate is doing, since removing it would undo your other request: it emits a locale category URL only when the index actually has entries of that type. With inference gone, video and 3d have zero declared entries, so they drop out of the sitemap instead of being advertised while the route 404s them.

imick-io (commented):
Reviewed at `aa9bd98d`. Verified locally: 649 unit tests pass, `astro check` 0 errors, eslint clean. Where I make an empirical claim below I ran the check rather than reasoning about it.

## Blocking: one classification change survived the revert

`aa9bd98d` took tag inference out of `serializeIndexEntry`, but `inferMediaType` is still widened from exact tag equality (main) to substring matching — and it still feeds `toTemplateData`. Traced end to end:

- `src/pages/workflows/[username].astro:57` → `toTemplateData` → `:187` renders `TemplateDetailPage`
- `src/components/template-detail/TemplateDetailPage.astro:182` builds the category breadcrumb as `/workflows/category/${data.mediaType}/`

So a workflow tagged `Image to Video` with no declared `mediaType` now resolves to `video` on that path and links the reader to `/workflows/category/video/` — which, after this same revert, is **empty in English and a hard 404 in every locale** (`src/pages/[locale]/workflows/category/[type].astro:73-76` returns 404 on an empty result).

Meanwhile the other detail route, `src/pages/workflows/[slug].astro:106`, uses `entry.mediaType || 'image'` and links the same workflow to the populated image category. Two detail routes, one workflow, contradictory category links — one of them pointing at a 404.

This is the behaviour @dante01yoon's review asked to remove, and the PR description now reads as behaviour-neutral on classification. Restoring equality inside `mediaTypeFromTagNames` keeps the "one definition, not two" refactor and drops the behaviour change. `tests/unit/hub-api-media-type.test.ts` asserts "never derives a media type from tags", which is true of `serializeIndexEntry` but not of the module — worth a case pinning `inferMediaType` too, since that is the half that regressed.

## Non-blocking, but worth addressing

**The model section of the sitemap gate is now tautological.** `scripts/verify-sitemap-indexability.ts:72`

Both sides derive from repo-owned files: the sitemap's `indexableModelSlugs` from `src/content/templates`, the check from `templates/index.json`. Those are the same catalog (`scripts/lib/sync.ts:89-103` uses the same reader, and CI passes no `--limit`) — which is the problem. The route resolves families from `loadSerializedTemplates()`, i.e. the live hub index. That file's own header says the gate exists because sitemap membership and the route's noindex decision "come from different datasets (on-disk JSON vs live hub index) and can drift". Reading the prerendered HTML *did* compare the two, because `getStaticPaths` built those pages from hub data. Files-vs-files cannot fail on that drift.

To be fair to the change: I checked whether this is live today. Fetched the production hub index (617 entries), derived model groups both ways, and they agree exactly — 11 indexable slugs on each side, zero sitemap slugs the hub grouping would 404 or 301, zero hub-indexable slugs missing from the sitemap. So this is a latent loss of coverage, not a live defect. Deriving the check's groups from `loadSerializedTemplates()` would restore it.

**The module-level memo in the model route does not survive a request.** `src/pages/workflows/model/[name].astro:58`

Astro compiles `.astro` frontmatter into the component's render function body; only imports are hoisted to module scope. Confirmed against this repo's own `@astrojs/compiler@2.13.0` — a top-level `let cache = null` lands inside `createComponent(async ($$result, ...) => {...})`. So `catalogCache` / `groupsCache` / `poolCache` reset on every render and the comment "Memoized per process, not per request" is inverted. What they do buy is deduping the three call sites *within* one request, which is real. Every hit still re-runs `serializeIndexEntry` over ~600 entries plus `deriveModelGroups`, `buildFallbackPool` and `useCasesFeaturingModel`. The hub fetch itself is genuinely module-cached (`src/lib/hub-api.ts:261`) and `s-maxage=300` bounds the cost, so this is not urgent — but moving the three lazy getters into a real module makes the comment true. The same pattern is already wrong in `src/pages/[locale]/workflows/model/[name].astro` on main.

**The empty-category gate is applied to locales only.** `astro.config.mjs:213`

`loadHubCategories()` correctly keeps `/zh/workflows/category/video/` out of the sitemap. But `src/pages/workflows/category/[type].astro:24-29` still prerenders all four types and the sitemap `filter` falls through to `return true` for `/workflows/category/*`, so `/workflows/category/video/` and `/workflows/category/3d/` stay advertised while rendering an empty grid. This is pre-existing on main, not introduced here, so it is not a merge condition — but it is the surface the blocker above sends people to, and reusing `loadHubCategories()` in the English branch of `filter` is a couple of lines.

**Prettier.** `scripts/build-tag-manifest.ts` and `src/lib/sitemap-custom-pages.ts` are not formatted (checked with the lockfile's prettier — `CATEGORY_MANIFEST_PATH` collapses to one line). main is already dirty on other files and the step is `continue-on-error`, so this is cosmetic, but the description claims prettier clean on every touched file.

**CI link-check noise.** `dist/workflows/model/*/index.html` no longer exists, so the internal-link step (`site-ci.yml:318`) will flag every link from `src/pages/workflows/model/index.astro` and the use-case pages as broken. Non-blocking, but it will pollute the PR comment and dull that step's signal unless model paths are excluded as known-dynamic.

## What checks out

- The redirect diagnosis is right, and `@astrojs/sitemap@3.7.0` does apply `filter` to `customPages` (`dist/index.js:87-90`), so gated model slugs cannot leak in through the new list.
- `assertUniqueModelSlugs` correctly restores the shadowing protection the deleted `.filter((variant) => !canonicalModelSlugs.has(variant))` provided, and the two-pass structure is necessary as written.
- Dropping the page-level `HreflangTags` is a genuine no-op — `SEOHead.astro:58-72` computes the same base path from `Astro.url.pathname` for `/workflows/tag/*` and `/workflows/category/*`.
- Reading `templates/index.json` in the verify script is equivalent to what sync writes.
- "A configured hub that fails is a hard build error" holds — `scripts/prebuild-parallel.ts:115-117` exits 1 on phase-2 failure.

Thorough writeup, and the redirect root-cause in particular saved a lot of guesswork.

mobeenabdullah:
@dante01yoon this needs your review please as you requested the changes.

mobeenabdullah:
@dante01yoon please re-review this PR

site/src/pages/workflows/model/[name].astro
```diff
@@ -38,53 +35,69 @@ import {
...
+let catalogCache: ReturnType<typeof buildCatalog> | null = null;
+const loadCatalog = () => (catalogCache ??= buildCatalog());
+
+const buildGroups = async () => deriveModelGroups(await loadCatalog());
+let groupsCache: ReturnType<typeof buildGroups> | null = null;
+const loadModelGroups = () => (groupsCache ??= buildGroups());
+
+const buildPool = async () => buildFallbackPool(await loadCatalog());
+let poolCache: ReturnType<typeof buildPool> | null = null;
+const loadFallbackPool = () => (poolCache ??= buildPool());
```
Comment (dante01yoon) [inline site/src/pages/workflows/model/[name].astro]: issue: This does not memoize per process — Astro compiles the component script into the render function, so `catalogCache`/`groupsCache`/`poolCache` are re-initialized to `null` on every request and `??=` never hits. Only the raw index fetch and profile map are cached (module scope in `hub-api.ts`); the ~600-entry serialization in `loadSerializedTemplates`, `deriveModelGroups`, and `buildFallbackPool` all re-run per request, which is exactly the work this comment says stays out of the response path. Since this mitigation is the stated justification for moving an SEO-critical route on-demand, could we move these caches into a real module (e.g. a small `src/lib/workflow-pages/model-route-cache.ts` that both routes import)? The locale route's identical pattern (`[locale]/workflows/model/[name].astro`) has the same no-op and could share it — pre-existing there, so fine as a follow-up, but this route is new code and should get the real cache now.

site/src/lib/hub-api.ts
```diff
@@ -498,14 +505,38 @@ export function toTemplateData(workflow: HubWorkflowDetail) {
...
+/**
+ * Media type from tag names. One definition, shared by the index and detail paths.
+ *
+ * Substring rather than equality: the hub's vocabulary is phrase-shaped ("Image
+ * to Video", "Video Edit", "Audio to Video"), so an equality check matches only
+ * the bare "Video" tag and files the rest as images.
+ *
+ * Order encodes the OUTPUT medium, which is what a category page groups by, so
+ * "Audio to Video" has to land on video and not audio.
+ */
```
dante01yoon: nit-pick (non-blocking): This first doc block is a stale leftover from the substring version — it directly contradicts the block below it ("Substring rather than equality" vs "Deliberately not substring") and the exact-match implementation, and "shared by the index and detail paths" is no longer true now that the index reads declared `mediaType` only. Please delete this block and keep only the EXACT-match one; contradictory docs on the one heuristic that already regressed once is how it regresses again.

dante01yoon (changes requested):
Re-reviewed at `33dac7d`. The contract blockers from my last two reviews are resolved:

1. Tag-derived classification is out of index serialization (`serializeIndexEntry` reads declared `mediaType` only) and out of the manifest (`build-tag-manifest.ts` derives categories from declared values only), with `hub-api-media-type.test.ts` guarding against reintroduction. The category gate keeps the sitemap honest about what the declared data actually fills. Thank you for also putting the cost on record (174/616 undeclared, 89 video) — that is the right way to hand this to the backend.
2. On the detail-path `inferMediaType`: I accept keeping it in this PR. It is main behaviour restored to exact-match, not something this PR introduces or extends, and the tests now pin that tags cannot move a classification. It goes on the same deletion list as the `'image'` default once the contract lands.
3. On your question about where the backend ticket goes: `/api/hub/workflows/index` on `cloud.comfy.org` is served by the Comfy Cloud backend (Comfy-Org/cloud). Please file it there with your numbers — populate + backfill `mediaType` (including correcting the 198 rows that declare `image` with video tags), make the field required in the OpenAPI schema. Once it is in the spec, this repo can consume a generated client type the same way the frontend consumes `@comfyorg/ingest-types`, and the handwritten optional mirror plus the `'image'` default and detail-path inference can all be deleted against it.

One new blocking issue on the on-demand rework (inline), and one stale-comment nit. The blocker is small and mechanical to fix; I don't need a full re-review pass afterwards.

Validation: read the full diff and the head versions of `hub-api.ts`, both model routes, and the sitemap/manifest modules; all CI checks green on this head.

site/src/pages/workflows/model/[name].astro
Comment (mobeenabdullah) [reply site/src/pages/workflows/model/[name].astro]: Confirmed and fixed in 846421c6. You were right, and I verified the mechanism rather than taking it on trust: compiled a probe component with @astrojs/compiler and the frontmatter declarations land inside $$createComponent, so they are re-initialised per render. Only export const prerender is hoisted to module scope.

So the caches were doing nothing, on a route this PR moves on-demand specifically for SEO. The hub fetch stays cached in hub-api, but the ~600-entry serialization, the group derivation and the fallback pool were all re-running per response.

Moved to src/lib/workflow-pages/model-route-cache.ts as you suggested. Took the locale route with it rather than leaving it as a follow-up, since it had the identical no-op and the module makes it a two-line change. The promise is cached rather than its result, so concurrent requests arriving before the first resolves share one derivation instead of starting their own.

site/src/lib/hub-api.ts
mobeenabdullah: Agreed, deleted in 846421c6. The first block described the substring version and contradicted both the block below it and the implementation, and you are right that shared by the index and detail paths stopped being true once the index moved to declared mediaType only.

Kept the EXACT-match block alone. Your reasoning is the part worth keeping on the record: contradictory docs on the one heuristic that has already regressed once is how it regresses again.

site/src/lib/hub-api.ts
mobeenabdullah: Went and measured the API so this stops being a design argument. All 620 workflows, every page.

**`media_type` is not in the hub API at all.** Not on `/api/hub/workflows` (keys: description, is_app, name, profile, publish_time, share_id, status, tags, thumbnail_type, thumbnail_url) and not on the detail endpoint's `metadata` (content_template, extended_description, faq_items, how_to_use, meta_description, suggested_use_cases, usage). Our own type has it optional and `meta.media_type || inferMediaType(workflow)` falls through to inference every time for community workflows.

So requiring the field is a backend change, not a frontend one. That is the ticket we never agreed a home for.

**`thumbnail_type` is declared and complete:** 620 of 620, no gaps. video 115, image 345, image_comparison 160. It is backend-owned, already in the response, and we already read it for `thumbnailVariant`. It would fix `/category/video/` with no heuristic at all.

**It cannot express audio or 3d.** There is no such value, so those two categories stay empty on declared data alone. Tags are the only signal that carries them (`3d-model`, `audio`, `video-edit` among 75 distinct tags), which is what this PR removed.

Three options, your call as owner:

1. Ship as-is. video/3d categories 404 until the backend adds `media_type`. Correct, and visibly broken for a while.
2. Derive video from `thumbnail_type`, leave audio/3d out. Declared data only, no inference, fixes the largest category now.
3. Backend adds `media_type` first and this PR waits on it.

I would take 2 and file the backend ticket, but it is your contract. Tell me which and I will do it.

site/astro.config.mjs
```diff
@@ -49,6 +59,43 @@ if (fs.existsSync(templatesDir)) {
 const modelGroups = deriveModelGroups(contentTemplates);
```
dante01yoon: issue: These relocated build guards derive `modelGroups` from `src/content/templates`, but both on-demand model routes resolve requests from `loadModelGroups()`, whose catalog comes from the live `/api/hub/workflows/index`. The sync source is the committed `templates/index.json`, so an API-only model or alias change can bypass `assertUniqueModelSlugs` and brand-safety validation even though the runtime route serves it. Before this change, `getStaticPaths` built and validated groups from `loadSerializedTemplates`, the same Hub dataset. Please run these guards against a prebuild Hub-derived manifest/catalog, or otherwise validate the exact route dataset.

site/scripts/verify-sitemap-indexability.ts
```diff
@@ -42,6 +68,25 @@ function renderedIndexableSlugs(section: Section): Set<string> {
   return slugs;
 }
 
+/** On-demand sections: recompute the verdict with the route's own helper. */
+function modelIndexableSlugs(): Set<string> {
+  const categories = loadTemplateIndex(DEFAULT_LOCALE);
```
dante01yoon: issue: The replacement model indexability check reads `templates/index.json`, while sitemap model membership is derived from `src/content/templates`, which prebuild synchronizes from that same committed index. This compares the repo catalog to itself, not to the live Hub catalog the on-demand route actually serves, despite the file header saying the datasets can drift. If Hub drops or reclassifies a family after the committed index, both sides retain the slug and CI passes while `/workflows/model/<slug>/` returns 404 or noindex. Please derive the route-side verdict from the Hub dataset used by `loadSerializedTemplates` (persisting it for this artifact-only job if needed), or probe the deployed SSR route.

site/scripts/build-tag-manifest.ts
```diff
@@ -0,0 +1,86 @@
...
+    // written, so a fresh checkout (which is what CI and Vercel build from) has
+    // no manifest rather than a stale one.
+    if (readEnv('PUBLIC_HUB_API_URL')) {
+      throw new Error(`Hub API failed while building the tag manifest: ${err}`);
+    }
+    console.warn(
+      `Hub API unreachable (${err}). Skipping the tag manifest — the sitemap will ` +
+        `omit localized tag URLs for this build.`
+    );
+    return;
```
dante01yoon: issue: The degraded return leaves any previously generated tag/category manifests in place. In a reused checkout, one successful build followed by an unconfigured build whose Hub fetch fails will return here, then `astro.config.mjs` will read stale files and advertise old localized tag/category URLs instead of the documented “no localized URLs” degradation. Please remove both generated manifests before this return, and before rethrowing if stale output must never survive a failed refresh.

dante01yoon (changes requested):
Three concrete correctness gaps remain. Two blocking guards now validate the committed repo catalog rather than the live Hub dataset used by the runtime routes, and failed manifest refreshes can retain stale generated routing state.

site/astro.config.mjs
mobeenabdullah: Confirmed and fixed in adc0108b. Measured it first: the repo catalog yields 120 model families, the hub index 81, and they disagree by 45. Three the hub serves that these guards never saw, so a hub-only slug collision or denied term could pass a green build, and 42 the guards validated with no route behind them.

Prebuild now writes the hub-derived families next to the tag manifest, from the same fetch, carrying only the fields these callers read: slug, label, qualifies, redirectFrom, keywords, templateCount. The guards and the sitemap read that, so both run against the dataset loadSerializedTemplates serves.

Falls back to the content catalog when the manifest is absent, which is a plain config load with no network, and is how everything else in this file already degrades.

site/scripts/verify-sitemap-indexability.ts
mobeenabdullah: Right, and it was worse than partial as you describe. Fixed in adc0108b.

It read templates/index.json while sitemap membership came from src/content/templates, which prebuild syncs from that same file, so the check compared the repo catalog to itself, 120 families deep, while the route resolved 81 from the hub.

It now reads the hub-derived manifest prebuild writes. The fetch cannot happen in this job, since it runs against a downloaded build artifact with no network, so persisting it at prebuild is the only way to get the route dataset here, which is what you suggested. Falls back to the committed index when the manifest is absent, since that is still better than skipping the check.

site/scripts/build-tag-manifest.ts
mobeenabdullah: Good catch, fixed in adc0108b. The comment even asserted the assumption you are challenging, that a fresh checkout has no manifest rather than a stale one, which is true of CI and Vercel and not of a reused checkout.

Both files, and the new model manifest, are now removed before the return and before the rethrow, so stale output survives a failed refresh in neither branch.

Added a test that seeds stale manifests, spawns the script against a refusing address so the failure is the hub being unreachable rather than a timeout, and asserts the build fails and both files are gone. Verified it goes red without the removal.

site/astro.config.mjs
mobeenabdullah: Confirmed and fixed in adc0108b + dc58e7b3. Measured it rather than taking it on trust, and it is not latent:

```
hub groups 81                 repo groups 120
hub variant aliases 41        repo variant aliases 48

variant only in the hub  : kling-1-6, kling-2-0
variant only in the repo : flux-3-video, gpt-image-2, ltx-2-5, qwen-image-3-0-pro,
                           qwen-image-edit-2511, seedance-2-5, seedream-5-0-pro,
                           wan-animate, wan3-0
qualifying only in the hub : minimax, elevenlabs, google-gemini, recraft, magnific,
                             vidu, google-gemini-image, hidream
```

So `assertUniqueModelSlugs` was checking 48 aliases no route resolves while missing 2 that it does, and `assertBrandSafe` was skipping 8 qualifying families the routes serve. Green, on the wrong catalog.

Prebuild now writes `src/data/hub-model-catalog.generated.json` and the config derives from that. The config still cannot fetch the hub itself, for the same reason the tag manifest exists.

One detail worth naming, because it is the way this fix silently fails: the manifest is built from `loadSerializedTemplates()`, not from raw `listWorkflowIndex()` rows. `deriveModelGroups` weighs `usage` against `MIN_CLUSTER_USAGE`, and `serializeIndexEntry` runs `applyRanking()` over it. `deploy-site.yml` and `cron-rebuild-site.yml` both set `PUBLIC_ALGOLIA_*`; with those credentials ranking rewrites `usage` on 437 of 645 entries and flips `qualifies` on seven families (`ace-step`, `elevenlabs`, `google-gemini`, `recraft`, `magnific`, `vidu`, `hidream`). Deriving from raw rows would have put the manifest and the route back out of step on exactly the production build, and `site-ci.yml` sets no Algolia credentials so CI measures zero divergence and would never have caught it.

The manifest carries the catalog rather than pre-derived families, so there is one derivation over one dataset and nothing precomputed to drift.

The sitemap's `indexableModelSlugs` moved with the guards, since advertising a list the route does not resolve is the same defect one surface over. No output change today: both derivations yield the same 11 indexable slugs.

Proof the guard now sees the route's dataset — dropping the Wan family from the hub manifest while leaving the repo catalog untouched, which is your hub-drops-a-family case:

```
Orphaned model landing content (no page renders these slugs): wan.
```

That passed before the change.

`build-tag-manifest.ts` is now `build-hub-manifests.ts` (`pnpm build:hub-manifests`); it writes three files and the old name had stopped describing it.

site/scripts/verify-sitemap-indexability.ts
mobeenabdullah: Confirmed and fixed in dc58e7b3. @imick-io raised the same thing on 08-24 as non-blocking; you are both right, and the script's own header said why.

It now derives families from the hub — `listWorkflowIndex()` / `loadSerializedTemplates()`, the catalog the route reads — so hub-vs-snapshot drift is detectable again.

One mechanism note, because the obvious fix does not work here. An intermediate revision had the verifier read the prebuild manifest instead of fetching. That job cannot see it: `seo-audit` runs checkout plus `download-artifact` of `site/dist` and never runs prebuild, and `src/data/*.generated.json` is gitignored. The manifest would have been absent on every CI run and the check would have fallen straight back to the repo catalog — the same tautology, now silent. It does have network (it runs `pnpm install`, and `PUBLIC_HUB_API_URL` is unset so the base defaults to `cloud.comfy.org`, which the build job already reaches), so fetching is the thing that job can actually do.

Two deliberate details:

- A hub outage degrades rather than fails. This gate blocks merges and a branch should not go red because the hub was down. It prints `model: DEGRADED — ... drift cannot be detected in this run` and falls back to `templates/index.json`. Silent would have been the same bug you just found, one level up.
- `listWorkflowIndex()` is called first, and its failure is what triggers the fallback. `loadSerializedTemplates` swallows a hub failure when `PUBLIC_HUB_API_URL` is unset — which is this job — and returns the empty collection it was handed, so on its own it would have read as "nothing is indexable" and failed the gate for the wrong reason.

One coupling I went looking for while making the change, since it is how a false verdict gets back in: `listWorkflowIndex` filters on `PUBLIC_APPROVED_ONLY` and `usage` is ranked through `PUBLIC_ALGOLIA_*`, so this check has to run with the same values as the build whose artifact it reads. That holds by construction today — `site-ci.yml` is the only workflow that runs this step, and neither its build job nor its audit job sets any of them — and it is stated in the file, because setting one on a single job would break the check without touching a line of it.

Verified both ways against a real build:

```
model: 11 in sitemap, 11 indexable
use-cases: 17 in sitemap, 17 indexable
Sitemap membership matches the rendered-indexable set.

# with the hub unreachable
hub index unreachable (TypeError: fetch failed)
model: DEGRADED — hub index unavailable, families derived from templates/index.json.
       Drift between the sitemap and the live catalog cannot be detected in this run.
model: 11 in sitemap, 11 indexable
```

site/scripts/build-tag-manifest.ts
mobeenabdullah: Confirmed and fixed in adc0108b + dc58e7b3. `clearManifests()` removes all three generated files before the degraded return and before the rethrow, so a failed refresh cannot leave a hub snapshot of unknown age behind for the config to read.

Pinned by `tests/unit/hub-manifests-degradation.test.ts`, which seeds stale manifests, spawns the script against a refusing address, and asserts every file is gone. Confirmed red with the removal commented out:

```
AssertionError: hub-tag-slugs.generated.json survived a failed refresh: expected true to be false
```

The script is now `scripts/build-hub-manifests.ts` — it writes three files (tag slugs, categories, model catalog) and the old name had stopped describing it.

mobeenabdullah:
@imick-io Late reply, sorry — your review sat unanswered while the branch moved. Six items: five are in, one I want to push back on. All of it is on the current head (`dc58e7b3`).

## 1. Classification — fixed, and further than you asked

Your trace was exact. `inferMediaType` is now byte-identical to main rather than equality-restored inside `mediaTypeFromTagNames`: after the revert that wrapper had one caller, so the "one definition, not two" split it existed for stopped buying anything. `tests/unit/hub-api-media-type.test.ts` covers both mappers now, and the detail cases were confirmed failing by reintroducing the substring rule.

The file's header claimed a module-wide guarantee that was never true of the detail mapper. Corrected: the listing mapper never infers, the detail mapper still does on exact tags, and that is pre-existing and pinned rather than asserted away.

## 2. Sitemap gate — fixed

Files-vs-files, agreed, and the script's own header said why that was wrong. @dante01yoon raised it as blocking on 08-27 and it is fixed now: the check derives families from the hub, the catalog the route reads.

Worth recording, since it is the fix that looks right and is not: reading the prebuild manifest instead of fetching does not work. `seo-audit` runs checkout plus `download-artifact` of `site/dist` and never runs prebuild, and `src/data/*.generated.json` is gitignored, so the manifest is absent on every CI run and the check falls silently back to the repo catalog. The job does have network, so it fetches. A hub outage degrades loudly (`model: DEGRADED — ...`) rather than reddening a merge gate.

## 3. Memo — confirmed, moved

Reproduced your compiler check rather than taking it on trust: fed a probe through this repo's `@astrojs/compiler@2.13.0` and `let cache = null` lands inside `createComponent(...)`. Only imports hoist. The three getters are now in `src/lib/workflow-pages/model-route-cache.ts`, so the scope is real and the comment is true.

## 4. English category gate — done

Also confirmed live rather than theoretical: declared counts on the index today are image and audio only, **video 0, 3d 0**, and both empty pages were prerendered and advertised. The sitemap `filter` now gates English categories on the same manifest as the localized URLs, with one guard worth naming: only when the manifest is non-empty. It returns `[]` when absent, and absent means "prebuild did not run", not "no categories" — an unguarded gate would have dropped the populated image category too.

## 5. Prettier — fixed

Every file this PR touches, verified with the lockfile's prettier. The description's claim was wrong and has been corrected along with it.

## 6. Link check — this one I don't think holds

The step is saturated on every branch, not just this one. Re-measured today:

```
#1164  44243 broken / 44243 checked
#1198  44243 broken / 44243 checked
#1199  44243 broken / 44243 checked
#1201  44243 broken / 44243 checked
```

Cause is unrelated to this branch: the step resolves `dist/<path>/index.html` while the adapter writes `dist/client/<path>/index.html`, which is exactly why `verify-sitemap-indexability.ts` probes `['dist/client', 'dist']`. Every link has been reported broken on every branch for as long as the adapter has emitted a client dir, and the step is `continue-on-error`. So there is no signal here for model paths to dull.

Happy to fix it properly in its own PR — resolve the dist root, then exclude the genuinely dynamic routes, which is when your exclusion becomes necessary rather than cosmetic. Doing it here would flip a dead check live mid-review and surface a backlog that has nothing to do with this branch.

## On the contract half

Since it came up in both reviews: `media_type` was never modelled as a field. `database/schema/hub_workflow_detail.go:62` keeps it as one optional key inside a free-form `metadata` JSON blob, and `common/hub/service.go:1576` reads it back with an unchecked map lookup, so a publish that omits it is legal and the row ships with no media type. The same schema already made the opposite call twice — `thumbnail_type` is a real enum and `is_app` is a real bool whose own comment says "Stored, not derived, so it can be queried and indexed." That is going to `Comfy-Org/cloud` separately; it needs the field promoted, a backfill that corrects the 198 wrong values as well as filling the blanks, and validation at ingest, none of which can happen from this repo.

The description is rewritten: it no longer describes inference that is not in the code, and the verification block is re-measured against this build.

mobeenabdullah:
@dante01yoon all three of your 08-27 findings are fixed on `dc58e7b3`, with the detail in each thread. Short version:

1. The guards now derive from the hub catalog prebuild writes, not `src/content/templates`. Your instinct was right and the gap was live: 41 hub variant aliases vs 48 repo ones, and 8 qualifying families the brand-safety check had never seen.
2. The indexability check fetches the hub rather than recomputing from `templates/index.json`. Reading the prebuild manifest instead does not work — `seo-audit` never runs prebuild and the file is gitignored, so it would have been absent every run and fallen back silently.
3. A failed manifest refresh removes all three generated files before returning or rethrowing, pinned by a test that was confirmed red without it.

Also in this round, from @imick-io's 08-24 review: English category URLs gated on the same manifest as the localized ones, and prettier over every touched file. The PR description is rewritten against this build.

One ask: four threads are still open that your 08-26 review says are resolved — `hub-api.ts:357`, `build-tag-manifest.ts:74`, `model/[name].astro:75` and `hub-api.ts:517`. All four are answered and the last two were fixed in `846421c6`. Could you close them out? I can't resolve human threads under the repo guidelines, and the stale CHANGES_REQUESTED is the only thing left holding this.

dante01yoon (approved):

PR #1167, opened 2026-08-20
fix(i18n): stop the footer offering languages that 404

mobeenabdullah:
## The bug, live in production

Every English use-case page, the use-cases index and the model index offer a 中文 link in the footer that returns **404**:

```
curl -sL https://comfy.org/workflows/use-cases/ai-headshot-generator/ | grep -o 'href="/zh/workflows/use-cases/[^"]*"'
  href="/zh/workflows/use-cases/ai-headshot-generator/"
curl -s -o /dev/null -w "%{http_code}" https://comfy.org/zh/workflows/use-cases/ai-headshot-generator/
  404
```

Introduced by flipping zh to indexable. The switcher renders once a language is rolled out and rebuilds the current path under each locale, but those routes have no `[locale]` variant, so the rebuilt URL points at nothing. 19 pages today (17 use-case pages + 2 index pages).

`/workflows/category/video/` has the same symptom from a different cause and is fixed in #1164.

## The fix

The routes already declare this fact to `HreflangTags` as `hreflangLocalized={false}`, a prop that exists so we "never advertise per-locale URLs that would 404". The footer now takes the same declaration and renders no switcher when there is no localized variant.

Link building moves into `footer-languages.ts` so the rule is testable, and a guard test keys off the existing `hreflangLocalized={false}` declaration: a route that makes it must not render a switcher. That is what stops this reopening when someone adds the next English-only page.

## Verified against a real build

| page | before | after |
| --- | --- | --- |
| `/workflows/use-cases/` | 1 zh link (404) | 0 |
| `/workflows/use-cases/ai-headshot-generator/` | 1 zh link (404) | 0 |
| `/workflows/model/` | 1 zh link (404) | 0 |
| `/workflows/` | 1 zh link (200) | 1 |
| `/workflows/tag/character/` | 1 zh link (200) | 1 |
| `/workflows/model/wan/` | 1 zh link (200) | 1 |
| `/workflows/creators/` | 1 zh link (200) | 1 |

`pnpm run lint`, `pnpm run check` (0 errors), `npx vitest run` (626 passing, +7 new). The guard test was confirmed failing before the fix.

## Not in scope

Building the localized use-case routes. Those pages carry 142 KB of English editorial copy that no pipeline translates today (the i18n content pipeline covers hub workflow entries only, keyed by shareId), so a `[locale]` route would publish English body text under indexable Chinese URLs. Suppressing the dead links is the correct first step; the routes come when there is Chinese copy to put in them.

### DIFF
--- a/site/src/components/hub/SiteFooter.astro
+++ b/site/src/components/hub/SiteFooter.astro
@@ -3,39 +3,22 @@
  * SiteFooter - Blue footer matching the Figma design
  * Blue background, Comfy logo, social icons, 4-column link grid
  */
-import { LANGUAGES, DEFAULT_LOCALE, type Locale } from '../../i18n/config';
-import { getLocaleFromPath, localizeUrl } from '../../i18n/utils';
-import { INDEXABLE_LOCALES } from '../../lib/i18n/locales';
+import { footerLanguageLinks } from '../../lib/i18n/footer-languages';
 
-// Language switcher. Self-derives from the current URL so no caller
-// has to thread props: the current locale is the path's locale prefix, and the
-// un-prefixed base path is what every alternate URL is rebuilt from.
-const currentLocale = getLocaleFromPath(Astro.url.pathname);
-const strippedPath =
-  currentLocale === DEFAULT_LOCALE
-    ? Astro.url.pathname
-    : Astro.url.pathname.replace(new RegExp(`^/${currentLocale}`), '') || '/';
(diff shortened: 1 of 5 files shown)


dante01yoon (approved):
Verified the production 404 path against the route structure. This change only derives presentation links from frontend-owned route and locale configuration; it does not reconstruct API/server state. The English-only route declarations are covered by a guard test, and current checks are green.

PR #1171, opened 2026-08-20
fix(site): add JSON-LD @graph structured data to workflow pages

balpreetgrowthnatives:
## Summary
Added a single linked JSON-LD @graph to workflow detail pages — WebSite, Organization, ComfyUI SoftwareApplication, WebPage, workflow SoftwareApplication+TechArticle, DefinedTerm/DefinedTermSet entities, BreadcrumbList, and FAQPage — via buildWorkflowGraphJsonLd(). Entity names, sameAs URLs, and category assignments come from workflow-entity-graphs.ts.

## Pages affected
- LTX-2.5: Image to Video (video_ltx2_5_i2v)
- Seedance 2.5: Reference to Video (api_seedance2_5_r2v)
- MiniMax H3: Image to Video (video_minimax_h3_i2v)
- Wan Animate 2: Motion Transfer (video_wan_animate2)


mobeenabdullah (approved):
LGTM. Ran it locally rather than reading only: 20/20 unit tests pass, and CI is green across all 10 checks.

Checked the things that usually go wrong with structured data:
- No fabricated entities. No author, aggregateRating, ratingValue, reviewCount or price anywhere. `creator` references the real Organization node and `datePublished` comes from the template's own date.
- All 76 unique `sameAs` URLs resolve. (A first sweep showed 20 non-200s; those were 429s from my own request rate, re-checked with delay and they are all 200.)
- Correctly scoped. Only 4 templates have curated graphs, and `getWorkflowEntityGraph` returns null for the rest, so every other template keeps its existing scripts untouched. The `!entityGraph &&` guards make the two paths mutually exclusive, so there is no duplicate or conflicting JSON-LD.
- Nothing is dropped on curated pages. The suppressed per-template SoftwareApplication is preserved by the workflow node as `['SoftwareApplication', 'TechArticle']`, with `runtimePlatform` pointing at the ComfyUI node, so the replacement is a superset rather than a swap.

The header comment stating the data is transcribed verbatim and must not be extrapolated is the right instinct for this kind of file.

PR #1198, opened 2026-08-26
feat(site): enforce the hreflang contract in the sitemap verifier

mobeenabdullah:
## Why

hreflang is a mutual declaration. Google discards an entire language cluster when one member fails to reciprocate, carries a duplicate annotation, or points somewhere off-origin. None of those failures surface in the build, in the rendered page, or as a Search Console error. The pages simply stop ranking in their own language, and we find out weeks later from a traffic graph.

We now ship ten indexable locales and nothing in CI would notice any of it breaking.

The existing `verify:sitemap-indexability` step is the right home: it already runs blocking in `site-ci.yml`, unlike `validate:sitemap` right below it which is `continue-on-error`.

## It already found a live bug, and that bug is now fixed on main

Writing the check turned up a live one: `/workflows/category/<x>/` and `/workflows/tag/<x>/` emitted **every hreflang value twice**. Both passed a second `HreflangTags` into `BaseLayout`'s head slot while `BaseLayout` already renders one through `SEOHead`; the base paths matched, so the two sets were byte-identical and nothing looked wrong on the page.

**That fix is no longer part of this PR.** #1164 removed the duplicate from both routes on main, and #1217 then narrowed what they advertise to `[DEFAULT_LOCALE, ...INDEXABLE_LOCALES]` so they stop naming `it`, which has no built route. Both files are taken from main unchanged here, so this branch now touches **no page at all**: it is the rules module, its tests, and the wiring into the existing verifier.

Worth keeping as the argument for the check, though. The duplicate sat on production for as long as it did precisely because a broken cluster is invisible everywhere except in the rendered markup.

## What it checks, and what it deliberately cannot

Blocking, on every rendered `index.html`:

- **No duplicate hreflang** on a page, which is ambiguous and drops the cluster.
- **Same origin** for every alternate href, compared as parsed origins.
- **Self-reference**: a page emitting alternates must be a member of the set it advertises.
- **Canonical origin**, which catches a preview-origin canonical shipped to production.
- **The label describes the target**: each alternate's `hreflang` must match the locale prefix of the path it points at, with an unprefixed path read as English and `x-default` exempt. Reciprocity and self-reference both accept a cluster whose `ko` and `ja` labels are swapped: every link resolves, every page lists the others, every page is a member, and Google is told each page is the wrong language. Settled by the URL alone, so it holds for on-demand targets too, which is why it runs before the target is looked up rather than after.
- **Target exists**, for the localized detail route, which is prerendered.
- **Reciprocity and target indexability** between pages that are both in the build output.

**Not** asserted: existence, indexability and the return leg for a target whose route is served on demand. The localized listing, category, tag, model and creators routes are all `prerender = false`, so their URLs never become files and absence proves nothing; asserting on them reported 920 failures against a correct site.

`[locale]/workflows/[slug].astro` is **not** one of those. It carries no `prerender = false`, so every localized detail page is a file the build owes us, and English detail pages advertise exactly those URLs through `indexableAlternateLocales()`. A build that dropped one is a broken cluster member, not an unknown.

What stays unverifiable is counted and printed rather than silently skipped, so the limit of what this check can prove stays visible instead of reading as a clean pass:

```
  model: 11 in sitemap, 11 indexable
  use-cases: 17 in sitemap, 17 indexable
  hreflang: 7912 of 8561 pages emit alternates (https://comfy.org)
  hreflang: 810 alternates target server-rendered routes, not checkable here
Sitemap membership and the hreflang contract both hold.
```

#1164 has since merged, so the localized listing URLs are in the sitemap and these can be tightened to "built **or** listed in the sitemap". Still a follow-up rather than this PR: it widens what the check asserts, and that deserves its own review rather than riding in on a conflict resolution.

## The two facts it reads rather than infers, and why each fails closed

Both are policy questions the emitted pages cannot answer, and both would switch a rule off in silence if they answered wrongly.

**Is the localized detail route prerendered?** Read from `src/pages/[locale]/workflows/[slug].astro`, not from whether detail pages appear in `dist`. `getStaticPaths` emits none of them when the hub index is unavailable, so inferring would let the build that dropped the whole family excuse itself. The read is a tested pure function that strips comments and string literals before matching, because that route's own header comment is about which of its siblings are server-rendered, and a mention in prose must not be able to disable the rule. When the route file is missing entirely the script throws rather than answering "not prerendered", so a moved or renamed route is a red build instead of a quietly weakened one.

**Which segments head a non-detail family?** A four-name list in the rules module, held to `src/pages/[locale]/workflows/` by a test that derives the same set from the routes and fails when they disagree. Drift is invisible in a build in both directions: an on-demand family left out reports every locale of it as a page the build owes, and a prerendered family listed in it excuses the absence of pages that really are missing. Verified red on all three cases.

## Noindexed pages are skipped as a source

Google drops a noindexed page from the index and its hreflang annotations with it, so its outgoing cluster cannot break anything. The gated locale detail pages emit exactly such a one-way cluster by design. They stay in the lookup as targets, so an indexable page advertising a gated one is still reported: once, and without also being asked for a return leg the noindexed page could never have provided.

## Memory

The first CI run died with `JavaScript heap out of memory`, exit 134, which the local fixture was far too small to surface.

V8 returns a regex capture as a slice that keeps its entire source string alive. Hub detail pages are ~1.6 MB of HTML each, so keeping a handful of hrefs from each page retained the whole build. Measured on a 1,400-page fixture: **2,144 MB** held after a forced GC. Copying each extracted value through a `Buffer` yields a standalone string; the same fixture now retains **9 MB**, and the script completes over 3,200 pages (4.8 GB of HTML) with the heap capped at 256 MB, in about 4 seconds.

## Verification

Rules live in `scripts/hreflang-contract.ts` as pure functions, so they are testable against fixtures. 47 unit tests in the file, each negative case asserting the exact message, and each one confirmed red against the implementation it replaced.

Beyond the unit tests:

- Ran the rules against **live production HTML** for two complete clusters, the `/workflows/` listing and a partially gated detail page. No violations, so the checks do not fire on output we know is correct.
- Ran the finished script against the **real CI build output for this branch**, 8,642 pages, with the heap capped at 256 MB: exit 0 in about 4 seconds, 1,012 alternates unverifiable. Then broke that same real build two ways and confirmed each exits 1: removing one localized detail page (8 violations, one per advertising page in its cluster) and swapping two labels on the English listing whose targets are both on demand (2 violations). Both counts moved after the merge below, since #1164 made the English model pages on demand; the current ones are the log lines above.
- Built a fixture shaped like the real build (English listing, creators and category pages whose locale variants are on demand, English plus localized detail pages per workflow, and a gated non-indexable locale page) and broke it five ways. Each goes red: a missing localized detail page, swapped labels, an advertised detail page turned noindex (one problem per source page, not two), a canonical left on a Vercel preview origin, and the detail route file gone.

## Merged with main at 7f1009d5

Not conflict-free this time. Resolved as:

- **`verify-sitemap-indexability.ts`**, three hunks. #1164 made `main()` async and moved the model verdict off `dist` onto the live hub. Main's awaited model section and this branch's hreflang block are independent, so both stay; the header keeps main's account of how each section is read and gains this branch's paragraph; `MAX_REPORTED` now caps the sitemap problems too.
- **The entry point**, which neither side could supply alone. This branch wrapped a synchronous `main()` in `try/catch` so a missing route file reports one line instead of a stack. Against main's `async main()` that catch can never fire: the throw becomes a rejected promise. It is now `main().catch()`, keeping the one-line message. Confirmed by moving the route file aside.
- **`workflows/category/[type].astro` and `workflows/tag/[tag].astro`**, taken from main entire, per the section above. Worth flagging that **git reported these two clean**: both sides had added `hreflangBasePath={basePath}` to the same element at different positions, and the automatic merge kept both, leaving a duplicate attribute that neither `astro check` nor eslint objects to.

## Gate

`pnpm lint`, `prettier --check`, `astro check` (0 errors) and `vitest run` (883 tests, 65 files) all pass on the merge.

The merged script was also run end to end against a 25-page cluster fixture: clean, then red on a deleted prerendered locale page (10 problems, one per advertising page, the noindexed `it` page correctly not among them), on swapped `ja`/`ko` labels, and on the detail route file removed.

The one thing the merge could have broken is #1217 changing what six English pages advertise, so that was checked directly rather than assumed: this branch's own rules over `clusterLocales(true, [DEFAULT_LOCALE, ...INDEXABLE_LOCALES])` give a cluster of 11 with self-reference satisfied and no label mismatch for `/workflows/`, `/workflows/category/image/` and `/workflows/creators/`, including the `pt-br` and `zh-tw` case folds. Their targets are all on demand, so they land in the unverifiable count rather than in problems.


site/scripts/hreflang-contract.ts
```diff
@@ -0,0 +1,142 @@
...
+
+export function normalizePath(pathname: string): string {
+  const rooted = pathname.startsWith('/') ? pathname : `/${pathname}`;
+  return rooted.endsWith('/') ? rooted : `${rooted}/`;
+}
+
+/** Site-relative path for an absolute href, or null when it is not on `origin`. */
+export function pathForHref(href: string, origin: string): string | null {
+  if (href !== origin && !href.startsWith(`${origin}/`)) return null;
+  return normalizePath(href.slice(origin.length).split(/[?#]/)[0]);
```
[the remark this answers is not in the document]
mobeenabdullah: Valid. Confirmed the false positive: `originOf` returns `https://comfy.org` for a canonical, so `https://comfy.org:443/ko/workflows/` failed the raw-string prefix test and was reported off-origin. An uppercased host failed the same way, and this check is blocking in site-ci, so it would have failed a correct build.

Now parses the href and compares `url.origin`. The prefix case (`https://comfy.org.evil.test/`) still rejects. Regression tests added for the default port, the uppercase host, and a non-absolute href; verified they go red against the old implementation.

Also resolved each href once up front, since the reciprocity check reads every alternate of every target and would otherwise parse each href once per cluster member.

Fixed in 8353cbde.

site/scripts/verify-sitemap-indexability.ts
```diff
@@ -46,6 +61,51 @@ function diff(a: Set<string>, b: Set<string>): string[] {
...
+  }
+  let winner: string | null = null;
+  let best = 0;
+  for (const [origin, count] of counts) {
+    if (count > best) {
+      winner = origin;
+      best = count;
+    }
+  }
+  return winner;
```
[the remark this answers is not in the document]
mobeenabdullah: Valid, and the comment on that function claimed the opposite of what it did. Majority-voting the origin out of the build means a build that emitted a preview origin on every page agrees with itself and passes, which is the one case worth catching.

Now reads `PUBLIC_SITE_ORIGIN`, the same variable `src/config/site.ts` and `astro.config.mjs` read, normalised the same way, defaulting to https://comfy.org. Passed in as a raw value so the rules stay pure.

Safe for previews: Vercel previews already emit production canonicals, confirmed on the #1199 preview deploy. Verified the env var is honoured end to end (setting it to another host makes the whole fixture go red).

Regression fixture added as requested, and I confirmed the old majority-vote design passes it.

Fixed in 2e8d5c7e.

site/scripts/hreflang-contract.ts
```diff
@@ -0,0 +1,219 @@
...
+        );
+      }
+
+      // x-default is a one-way fallback declaration, not a cluster member, and a
+      // page needs nothing back from itself.
+      if (alternate.hreflang === 'x-default' || target === page.path) continue;
+
+      // x-default declares a fallback, not a language pairing, so it does not
+      // satisfy the return leg of a cluster.
+      const linksBack = (resolvedByPath.get(target) ?? []).some(
```
dante01yoon, line 333: suggestion: Reciprocity plus self-reference still accepts a cluster whose labels are swapped — e.g. `/ko/workflows/` advertised as `hreflang="ja"` and vice versa: every link resolves, every page lists the others, and every page is a member of the set, so no rule fires while Google is told each page is the wrong language. The hub's URL scheme is deterministic (locale prefix ↔ language, `en` at root), so a rule asserting that each alternate's `hreflang` matches its target path's locale prefix would close this. The marketing checker grew exactly this rule after reciprocity alone passed a swapped cluster (ComfyUI_frontend #15488's `hreflangAudit`). Non-blocking — fine as a follow-up tightening.

dante01yoon (approved):
Approving. The enforcement rides the already-blocking `verify:sitemap-indexability` step rather than a script nothing runs, the rules are pure and fixture-tested with exact-message assertions, and the unverifiable-target count keeps the check's limits visible instead of reading as a clean pass. I verified the live-bug fix at the source: `hreflangBasePath` is an existing prop flowing BaseLayout → SEOHead → HreflangTags, and SEOHead already computed the same base path when the prop was absent — which is exactly why the slot-passed second `HreflangTags` produced byte-identical duplicates. Removing the slot copy and pinning the prop is the right fix. The Buffer-detach for V8 slice retention is legitimate and well-measured.

One non-blocking suggestion inline.

Unanchored (cross-PR coordination, not a defect here): #1164 edits the same surfaces — it also modifies `verify-sitemap-indexability.ts` (model verdicts recomputed via helper once that route goes on-demand) and also removes `HreflangTags` from the same two pages, but without pinning `hreflangBasePath`. Whichever lands second needs a careful rebase so one fix doesn't stomp the other; and once #1164's on-demand model route lands, the English model pages leave the build output and become unverifiable targets here, which this check's counting already handles by design. Worth merging these two in a deliberate order.

Validation: read the full diff plus BaseLayout/SEOHead at head `2e8d5c7e`; confirmed the robots-meta parser matches SEOHead's actual attribute order; all CI checks green on this head.

site/scripts/hreflang-contract.ts
mobeenabdullah: Confirmed and implemented in 5803f3c7, so not deferring it.

Reproduced it first: a cluster with the ko and ja labels swapped produced zero problems under the existing rules, exactly as you describe.

Each alternate label is now checked against the locale prefix of the path it points at, with an unprefixed path read as English and x-default exempt. The locale list is a required argument rather than optional, because a default would let a future caller drop the rule without noticing; the rules module stays free of site config and the verifier passes LOCALES.

Also covered the mixed-case pairing, since the hub emits hreflang="zh-tw" for /zh-TW/ and "pt-br" for /pt-BR/. Re-ran the rules against live production HTML for the full eleven-locale cluster: no violations, so it does not fire on correct output.

imick-io (approved):

site/scripts/hreflang-contract.ts
```diff
@@ -0,0 +1,252 @@
...
+        );
+        continue;
+      }
+
+      const targetPage = byPath.get(target);
+      if (!targetPage) {
+        // Server-rendered, so existence, indexability and the return link are all
+        // unknowable from the build output.
+        unverifiable += 1;
+        continue;
```
dante01yoon, line 314: issue: This unconditional missing-target exit assumes every absent build target is server-rendered, but current main (merged into this head) now prerenders `[locale]/workflows/[slug].astro`. English detail pages use `indexableAlternateLocales()` to advertise localized detail URLs that are therefore expected in the build. If one of those files is omitted, this branch only increments the informational counter and CI passes, so the new blocking verifier misses a genuinely broken cluster member. The same early exit also skips the deterministic label-vs-path check for routes that really are SSR (for example, `hreflang="ja"` pointing to `/ko/workflows/...`). Could we validate the label before this lookup and distinguish on-demand route families from prerendered detail targets before deciding that absence is unverifiable?

dante01yoon (changes requested):
One blocking false-negative remains in the new blocking verifier: missing targets are treated as unverifiable even when the route family is now prerendered, and locale-label validation is skipped on that path.

site/scripts/hreflang-contract.ts
mobeenabdullah: You are right on both counts, and the premise behind the original code was wrong. Fixed in 70ae5d55.

I had concluded every localized route is `prerender = false`. It is not. I reached that from a list of files containing `prerender = false` and never checked the list was exhaustive; `[locale]/workflows/[slug].astro` was absent from it precisely because it is prerendered. Verified now: slug is prerendered, index/category/tag/model/creators/username are on demand.

**Label first.** Moved above the target lookup. It is settled by the path the alternate names, so it holds whether or not that page is a file, and behind the lookup it was skipped for exactly the routes nothing else can catch. Test covers `hreflang="ja"` pointing at `/ko/workflows/` where the target is on demand.

**Absence.** Any built `/<locale>/workflows/<x>/` page is a detail page, because the creator catch-all that shares its shape is server-rendered and never becomes a file. Once one exists in the build, a detail page an English page advertises but the build did not produce is reported as a broken cluster member. The other localized families stay unverifiable, since they really are on demand.

One deliberate choice worth flagging: the detail signal is site-wide, not per locale. Per locale would excuse the case most worth catching, a locale with zero detail pages, since `indexableAlternateLocales` only advertises a locale it believes indexable. Zero pages for an advertised locale means the build dropped them.

Both tests verified red against the previous ordering, and re-ran the rules against live production HTML for the listing cluster and a partially gated detail cluster: no violations.

mobeenabdullah:
@coderabbitai review

mobeenabdullah:
Two rounds of feedback are in since the last review, both on the same rule.

**@dante01yoon**, you were right that the premise was wrong. I had claimed every localized route is `prerender = false`; `[locale]/workflows/[slug].astro` is prerendered. I reached that from a list of files containing `prerender = false` without checking the list was exhaustive, and slug was absent from it precisely because it is prerendered.

Fixed in 70ae5d55:
- The label check runs **before** the target lookup. It is settled by the path the alternate names, so it holds whether or not that page is a file, and behind the lookup it was skipped for exactly the on-demand routes where nothing else catches it.
- A missing localized **detail** page is now reported rather than excused. The other localized families stay unverifiable, since they really are served on demand.

**@coderabbitai** then caught that the prerendering signal was read off the emitted pages, which inverts exactly where it matters: `getStaticPaths` emits no localized detail pages when the hub index is unavailable, so a degraded build produced zero, the signal flipped to "on demand", and every absent target was excused. That build is the one this check exists to catch.

Fixed in 5adeae35: the policy is a required argument, read from the route file rather than from the output. `output: 'static'` prerenders every page that does not opt out, so the absence of `prerender = false` in the route **is** the policy.

Both fixes have tests verified red against the previous behaviour. Re-ran the rules against live production HTML for two full eleven-locale clusters: no violations. Gate: lint clean, astro check 0 errors, 767 tests across 57 files.

mobeenabdullah:
CodeRabbit's outside-diff finding on the noindexed-target branch (no thread to reply in, and still listed open on the merge-risk summary) is fixed in ff3eeb7e. A noindexed target was reported and then also checked for a return link, so one broken alternate produced two problems on every source page in the cluster. Google drops a noindexed page along with its own annotations, so the return leg is moot. The alternate ends there now. Test asserts the single message and goes red without the `continue`.

Two leftovers from 5adeae35 in the same commit:

- The policy read that gates the whole existence rule matched the bare words `prerender = false` anywhere in the route file, and it fails open (a match means "on demand", which excuses every absence). One sentence in that file's own comments mentioning the flag would have switched the rule off silently. Anchored on `export const`.
- `prerendersLocalizedDetail`'s doc comment outlived the function and sat stacked above `detailLocaleOf`'s own; the `unverifiable` doc still asserted every localized route is `prerender = false`, which is the premise 70ae5d55 disproved.

Description updated: it still carried that premise and pre-OOM-fix page counts.

Verification added: ran the script against the real CI build output for this branch, all 8,564 pages, heap capped at 256 MB. Exit 0, 1,012 unverifiable. Removing one localized detail page from it exits 1 with 8 violations; swapping two labels on the English listing exits 1 with 2.

site/scripts/verify-sitemap-indexability.ts
```diff
@@ -46,6 +62,44 @@ function diff(a: Set<string>, b: Set<string>): string[] {
...
+ *
+ * Astro's `output: 'static'` prerenders every page that does not opt out, so the
+ * absence of `prerender = false` in the route file is the policy. Read from source
+ * rather than from the emitted pages: a build whose hub fetch failed emits none of
+ * them, and inferring from that would excuse the very absence worth reporting.
+ */
+function localizedDetailIsPrerendered(): boolean {
+  const route = path.join(SITE_DIR, 'src/pages/[locale]/workflows/[slug].astro');
+  if (!fs.existsSync(route)) return false;
+  return !/\bprerender\s*=\s*false/.test(fs.readFileSync(route, 'utf-8'));
```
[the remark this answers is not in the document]
mobeenabdullah: Split verdict.

**Missing route: valid, fixed in 0c4eac8b.** `return false` there meant a moved or renamed route reclassified every missing localized detail page as unverifiable and switched the existence rule off with CI green. Reproduced on a fixture build with the route file removed and a localized detail page genuinely absent: the previous version exits 0 and prints `Sitemap membership and the hreflang contract both hold.` It throws now, and `main()` is wrapped so the failure is one line and exit 1 rather than a stack:

```
Error: src/pages/[locale]/workflows/[slug].astro is missing, so the localized detail render policy cannot be determined. Point LOCALIZED_DETAIL_ROUTE at the route if it moved.
```

**Unanchored regex: stale.** You reviewed 5adeae35; ff3eeb7e anchored it on `export\s+const\s+prerender\s*=\s*false` three minutes later, so that half was already closed when this comment landed.

**Comments and strings: your point survives that anchor, so it is fixed too.** The anchor stops the bare words matching, not the declaration quoted inside a comment, and this route's header comment is entirely about which of its siblings are server-rendered. The match moved into `declaresOnDemandRendering()` in `hreflang-contract.ts`, a pure function that strips both comment styles and all three quote styles before testing. It also reads `prerender: boolean = false`, which the old regex missed.

8 unit tests on it, 4 of them verified red against the previous implementation: line comment, block comment, string literal, type annotation. Checked against the real routes: `[slug]` prerendered, `[username]`, `index`, `category/[type]`, `tag/[tag]`, `model/[name]` and `creators` all on demand.

Not parsing the module because `.astro` frontmatter is not importable from a tsx script. The stripper does not model regex literals, which these routes do not use; if one ever confused it the declaration would be missed, and a missed declaration asserts existence rather than excusing it, so that direction is a red build, not a silent pass.

Gate: 776 unit tests, `astro check` 0 errors, eslint and prettier clean.

mobeenabdullah:
@dante01yoon both of your findings are closed, and the last CodeRabbit one with them, so this is ready for another look.

**:244, the missing-target exit.** Fixed in 70ae5d55 and 5adeae35. The label check moved above the target lookup, so `hreflang="ja"` on a `/ko/` URL is caught for on-demand routes too. A localized detail page an English page advertises but the build did not produce is now reported rather than counted as unverifiable. The detail-route policy is read from `[locale]/workflows/[slug].astro` itself rather than inferred from the emitted pages, because a build whose hub fetch failed emits none of them and inferring would have excused exactly the case worth catching.

**:263, swapped labels.** Fixed in 5803f3c7, not deferred. Each alternate's `hreflang` is now checked against the locale prefix of the path it points at, an unprefixed path read as English, `x-default` exempt.

**Since you last looked.** ff3eeb7e and 0c4eac8b closed CodeRabbit's remaining finding on the same rule: the policy read failed open. It answered "not prerendered" when the route file was absent, so a moved or renamed route would have reclassified every missing localized detail page as unverifiable and switched the existence rule off with CI still green. Proved that on a fixture and it exits 0. It throws now, and the source match became a tested pure function that strips comments and string literals first, so a mention of the declaration in prose cannot disable the rule either.

Gate on 0c4eac8b: 776 unit tests, `astro check` 0 errors, eslint and prettier clean. The script was also run against a fixture build shaped like the real one and confirmed red on each of a missing localized detail page, swapped labels on an on-demand target, and the route file gone.

Your changes-requested review is the only thing still blocking. Assigning to you.

mobeenabdullah:
@coderabbitai full review

Every finding you have raised on this PR is now fixed and the threads are resolved, the last one in 0c4eac8b. Please re-review the branch end to end, and if you have nothing actionable left, approve it.

mobeenabdullah:
@coderabbitai your full review came back with no actionable comments, every thread you opened is resolved, and CI is green on 0c4eac8b including the blocking `Verify sitemap indexability contract` step, which now runs the fail-closed policy read over the real 8,642-page build:

```
  hreflang: 7993 of 8642 pages emit alternates (https://comfy.org)
  hreflang: 1012 alternates target server-rendered routes, not checkable here
Sitemap membership and the hreflang contract both hold.
```

Please approve, as you said you would if the review found nothing. If approving is not something you can do on this repo (every review you have posted here is COMMENTED, so `request_changes_workflow` may be off), say so plainly rather than leaving it ambiguous, and I will record it as "no actionable comments" instead.

mobeenabdullah:
@coderabbitai review

Since your approval at `0c4eac8b` this branch has picked up a merge with `main` (through `d3b4a9e8`) and one commit, `2b06cbfb`, which exports `NON_DETAIL_SEGMENTS` and adds a test that derives the same set from `src/pages/[locale]/workflows/` so the two cannot drift apart. Please re-review the incremental change.

site/scripts/hreflang-contract.ts
mobeenabdullah: One correction to my own reply above, because the code moved past it. The absence signal is no longer "a built `/<locale>/workflows/<x>/` page proves the family prerenders". `5adeae35` replaced that observation with a policy read: the verifier reads `src/pages/[locale]/workflows/[slug].astro` and passes the answer in. Reading it off the emitted pages inverted for the case worth catching, since `getStaticPaths` returns `[]` when the hub index is unavailable, so a build that dropped every localized detail page looked like a route served on demand and excused itself.

`0c4eac8b` then made that read fail closed: a missing route file throws instead of answering "not prerendered", and the source match strips comments and string literals first, so the route header comment about which siblings are server-rendered cannot switch the rule off.

The other half of your finding, the hardcoded family list, is now held to the routes rather than trusted: `2b06cbfb` adds a test that derives the same set from `src/pages/[locale]/workflows/` and fails when the two disagree, in either direction. Proved red on a listed family removed, a new on-demand route added, and a prerendered route wrongly listed.

mobeenabdullah:
@dante01yoon head is now `2b06cbfb`, all checks green including the blocking `Verify sitemap indexability contract`. Three things since the 08-28 ping, none of them changing a rule you have already read.

**The aging half of your second finding.** You caught the hardcoded route-policy assumption; the same class of thing survived one level down in `NON_DETAIL_SEGMENTS`, a four-name list that nothing held to the routes. Drift is invisible in a build in both directions: an on-demand family left out reports every locale of it as a page the build owes, and a prerendered family listed in it excuses the absence of pages that really are missing. `2b06cbfb` adds a test that derives the same set from `src/pages/[locale]/workflows/` with the already-tested `declaresOnDemandRendering` and fails when the two disagree. Proved red on all three cases: a listed family removed, a new on-demand route added, a prerendered route wrongly listed. The failure message prints the derived families and marks the prerendered ones, so it names which side moved.

Not switched to deriving the set at runtime. The test makes drift unable to land without a signature change on a live branch, and without a fifth argument that would be type-compatible with `locales` and silently swappable with it.

**Merged with `main`** through `d3b4a9e8`, zero conflicts. Checked its new pages for a false positive first: `workflows/minimax-h3-multiref.astro` is both `prerender = false` and `noindex`, so it is never a file and never a cluster source. #1164 is still open, so your merge-ordering note stands.

**The description was stale in five places** and is rewritten: the duplicate-hreflang bug is 26 links where there should be 13, not 24/12 (Italian is now in the default cluster; reproducible with a curl of any category page right now), 47 unit tests not 37, 817 across the site not 768, 8,642 pages not 8,564, and it never mentioned the fail-closed policy read from `0c4eac8b`.

One correction I also left in your `:296` thread: my 13:04 reply there described the observation-based absence signal from `70ae5d55`, which `5adeae35` replaced with the policy read five hours later. The thread had the superseded version.

```
  hreflang: 7993 of 8642 pages emit alternates (https://comfy.org)
  hreflang: 1012 alternates target server-rendered routes, not checkable here
Sitemap membership and the hreflang contract both hold.
```

Your changes-requested review is the only thing blocking. Both your threads are answered and implemented; leaving them for you to resolve.

dante01yoon (approved):
Strict review completed on the current revision. The duplicate hreflang issue is fixed, and the prior requested changes are covered by regression cases for missing targets, SSR exceptions, locale labels, noindex handling, trusted origin, reciprocity, and route-policy drift.

The implementation is approved. The branch is currently conflicting with the base branch, so the conflict still needs to be resolved; any material conflict-resolution change should be re-reviewed.

PR #1199, opened 2026-08-26
feat(site): declare the page language in Open Graph tags

mobeenabdullah:
## Why

The hub emits ten `og:` tags and not one of them says which language the page is in. All ten indexable locales currently present themselves to Slack, Facebook and LinkedIn as language-neutral.

This was specified: [TDD: GTM-291](https://app.notion.com/p/3a56d73d365080beb7daf90ae3cdea79) section 8.4 lists "`og:locale` and `og:locale:alternate` added" as part of the indexability cleanups. It never got built.

Verified absent on production before writing anything: `curl https://comfy.org/ko/workflows/ | grep 'og:'` returns ten tags, no `og:locale`.

## What changed

**`OG_LOCALES` in `src/i18n/config.ts`.** Open Graph wants `language_TERRITORY`, which is not the tag we route on: we say `ko`, it wants `ko_KR`. Typed as a full `Record<Locale, string>`, so registering a new locale without an entry here is a compile error rather than a page that silently claims no language.

**Alternates come from the hreflang cluster.** The "which locales does this page declare" rule was inline in `HreflangTags.astro`. It is now `clusterLocales` in `src/i18n/utils.ts` and both components call it, so the `og:locale:alternate` set and the `hreflang` set cannot drift into contradicting each other. Per-page gating (`hreflangLocales`, including `[]`) is respected identically by both.

**`unlocalizePath`.** `SEOHead.astro` carried its own copy of the locale-prefix rule, three lines from the import of `getLocaleFromPath` which implements the same check. Extracted and shared. A test asserts the new function matches the previous inline implementation across eleven path shapes, including `/en/…`, `/pt-BR/…`, a bare `/`, and `/korean-guide/` which must not be read as the `ko` locale.

**`vitest.config.ts` moves to `getViteConfig`.** Without it Vite cannot parse `.astro` template syntax, so no test could import a component and assert its rendered output. All 55 test files still pass and the suite is not slower.

## Verification

22 logic tests plus 3 that render `SEOHead` through Astro's container API and assert the real HTML, including that the `og:locale` set and the `hreflang` set match exactly on a gated page.

Both render assertions were proven able to fail. Breaking the mapping (emitting `ko` instead of `ko_KR`) turns 3 red; removing the self-locale filter so a page lists itself as its own alternate turns 2 red.

## Gate

`npm run lint`, `npx astro check` (0 errors) and `npx vitest run` (717 tests, 55 files) all pass.

## Not in this PR

The other unshipped items from GTM-291 section 8.4: duplicate hreflang emission on English tag and category pages, localizing the hard-coded English relative dates, and structured-data parity on locale pages.

### DIFF
--- a/site/src/components/HreflangTags.astro
+++ b/site/src/components/HreflangTags.astro
@@ -5,8 +5,8 @@
  * Tells search engines about language/region variations of the same content.
  * Should be included in <head> for all pages that have localized versions.
  */
-import { LANGUAGES, type Locale } from '../i18n/config';
-import { localizeUrl } from '../i18n/utils';
+import { type Locale } from '../i18n/config';
+import { clusterLocales, localizeUrl } from '../i18n/utils';
 import { absoluteUrl } from '../config/site';
 
 interface Props {
@@ -30,9 +30,8 @@ interface Props {
 
 const { basePath, localized = true, locales } = Astro.props;
 
-// Per-page list wins; otherwise fall back to "all locales / none" by the boolean.
-const alternateLocales: readonly Locale[] =
-  locales ?? (localized ? (Object.keys(LANGUAGES) as Locale[]) : []);
+// Shared with SEOHead's og:locale tags so the two declarations cannot diverge.
+const alternateLocales: readonly Locale[] = clusterLocales(localized, locales);
 
 // hreflang values must be lowercase (e.g. 'zh-tw' not 'zh-TW').
 const hreflangEntries = alternateLocales.map((locale) => ({
--- a/site/src/components/SEOHead.astro
+++ b/site/src/components/SEOHead.astro
@@ -1,6 +1,7 @@
 ---
 import HreflangTags from './HreflangTags.astro';
-import { LANGUAGES, DEFAULT_LOCALE, type Locale } from '../i18n/config';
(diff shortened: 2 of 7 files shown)


dante01yoon (approved):
Approving, no findings.

What I checked:

- `OG_LOCALES` as an exhaustive `Record<Locale, string>` is the right shape — a new locale without a mapping fails typecheck instead of shipping a language-neutral page. The values match Facebook's canonical locale list, including `ar_AR`, which looks wrong (AR is Argentina's territory code) but is genuinely what the OG consumer side lists for Arabic.
- `clusterLocales` shared between `HreflangTags` and `SEOHead` keeps the two declarations from contradicting each other, and the per-page gate (`hreflangLocales`, including `[]`) flows through both identically — the container-render test pins the two sets equal on a gated page, which is the assertion that matters.
- `unlocalizePath` is equivalence-tested against the exact previous inline implementation across the path shapes that bite (`/en/…`, `/pt-BR/…`, `/korean-guide/`), so the extraction changes nothing.
- The `getViteConfig` switch has the widest blast radius here (all 55 test files re-run under Astro's Vite config); the full suite passing covers it.

Validation: read the full diff at `f308382`; all CI checks green on this head.

PR #1200, opened 2026-08-26
fix(i18n): localize the relative date on workflow pages

mobeenabdullah:
## Why

Every localized workflow page shows its reader "2 days ago" in English. The strings were hard-coded in `TemplateDetailPage.astro`, so all ten locales render the English form next to otherwise fully translated content.

Listed as an unshipped cleanup in [TDD: GTM-291](https://app.notion.com/p/3a56d73d365080beb7daf90ae3cdea79) section 8.4.

## Why Intl and not a translated template

Plural morphology is not a suffix in most of the locales we ship. A `{count} days ago` string cannot be right in Russian, which needs день / дня / дней depending on the count, or in Arabic, which has a distinct dual form (يومين for two, not "2 أيام"). CLDR already encodes those rules, so the tests assert them directly:

```
ru   1 день назад | 2 дня назад | 5 дней назад
ar   قبل يوم واحد | قبل يومين   | قبل 5 أيام
```

Verified all eleven locales resolve to real ICU data with no English fallback.

## English does not change

`numeric: 'always'` produces "1 day ago" rather than "yesterday", so the output is byte-identical to what it replaces. A test asserts that against the previous implementation across every bucket and both sides of every boundary: 0, 1, 2, 15, 29, 30, 31, 59, 60, 359, 360, 364, 720, 1080, 3650 days.

Only the same-day case has no relative form to reuse, so `date.today` joins the UI dictionary. Values follow the existing house capitalization in those files (sentence-initial for latin and cyrillic, as `template.updated` already does).

## Two pre-existing defects fixed along the way

- An unparseable date rendered **"NaN years ago"** to the reader. It now renders nothing. This also matters because `Intl` throws a `RangeError` on `NaN`, which would break the build.
- A date in the future rendered **"-2 days ago"**. It now reads "in 2 days".

## Verification

31 tests. Both guards were proven able to fail:

- Making the formatter ignore the locale, which is the bug being fixed, turns all ten locale tests red while English stays green.
- Deleting `date.today` from `ru.json` turns 2 red. This matters because `t()` silently falls back to English, so a missing key otherwise looks like a working translation.

Gate: `npm run lint`, `npx astro check` (0 errors), `npx vitest run` (723 tests, 55 files).

## Note for reviewers

`TemplateDetailPage.astro` is also touched by #1185, in the import block and the models link list around line 200. This PR removes the date function at line 81 and changes its call site at line 234, so the hunks do not overlap. Whichever lands second should be a clean rebase.


site/src/lib/relative-date.ts
```diff
@@ -0,0 +1,42 @@
...
+  const diffDays = Math.floor(diffMs / MS_PER_DAY);
+  if (diffDays === 0) return t('date.today', locale);
+
+  const relative = new Intl.RelativeTimeFormat(locale, { numeric: 'always' });
+  if (diffDays < 30) return relative.format(-diffDays, 'day');
+
+  const diffMonths = Math.floor(diffDays / 30);
+  if (diffMonths < 12) return relative.format(-diffMonths, 'month');
+
+  return relative.format(-Math.floor(diffMonths / 12), 'year');
```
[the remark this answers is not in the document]
mobeenabdullah: Valid. Confirmed: diffDays of -400 takes `diffDays < 30`, so every future date stayed in the day lane and the month and year lanes were unreachable in that direction.

Applied. Verified the past-date results are untouched, including the assertions against the previous hard-coded English strings, and added tests for the future lanes (in 1 month, in 1 year, in 3 years, plus ja and ru). Confirmed those two tests go red against the signed-value version.

Fixed in a3aca75a.

site/src/lib/relative-date.ts
```diff
@@ -0,0 +1,47 @@
...
+  dateStr: string,
+  locale: Locale,
+  now: Date = new Date()
+): string {
+  const diffMs = now.getTime() - new Date(dateStr).getTime();
+  // An unparseable date yields NaN, which Intl rejects with a RangeError. The
+  // previous implementation rendered "NaN years ago" instead; show nothing.
+  if (Number.isNaN(diffMs)) return '';
+
+  const diffDays = Math.floor(diffMs / MS_PER_DAY);
```
dante01yoon: nit-pick (non-blocking): `Math.floor` sends a date that is seconds in the future (hub timestamp vs. build/render clock skew) to `diffDays = -1`, so it renders "in 1 day" instead of "Today". `Math.trunc` fixes that lane without changing any tested case — sub-day future becomes `0` → `date.today`, while the intentional future cases (exact multiples of a day in the tests, and anything ≥1 day ahead in practice) keep their current output. Clock skew against the hub API is the realistic way a reader ever sees a future date, so this is the case most worth getting right.

dante01yoon (approved):
Approving. Intl/CLDR is the right call over a `{count} days ago` template for exactly the reasons stated — the Russian and Arabic assertions prove the morphology a single interpolated string cannot produce. The equivalence test against the previous hard-coded implementation across every bucket boundary is what makes the English no-change claim trustworthy, and both pre-existing defect fixes ("NaN years ago" would now be a build-breaking `RangeError` under Intl, so the guard is required, not just cosmetic) are covered by tests proven able to fail. The `date.today` fallback test also correctly closes the `t()` silent-English-fallback trap.

One non-blocking nit inline.

Validation: read the full diff at `a3aca75`; verified the bucket math against the previous implementation at the 30/360-day boundaries and the locale values are valid BCP 47 tags; all CI checks green on this head.

site/src/lib/relative-date.ts
mobeenabdullah: Taken, fixed in 6ced5488, and it shares a cause with CodeRabbit's note on the same function.

Flooring while signed was the problem rather than floor itself. Taking the absolute distance before flooring covers your case: a sub-day gap in either direction is now zero, so it reaches date.today. It also fixes an asymmetry neither of us had named, where 29 days and 23 hours ahead floored to -30 and read "in 1 month" while the same gap in the past stayed "29 days ago".

Agreed on the reasoning too: clock skew against the hub API is the realistic way a reader ever sees a future date, so there is a test for exactly that case now. Every previously asserted case is unchanged, and both new tests go red against the signed-floor version.

imick-io (approved):

dante01yoon (approved):
Reviewed the current head strictly. The hard-coded English formatter defect is real, and the implementation correctly uses locale-aware CLDR behavior while preserving existing English bucket semantics. Invalid, future, fractional-boundary, fallback, and plural-morphology cases are covered; current CI is green.

PR #1201, opened 2026-08-26
fix(seo): give localized model pages their SoftwareApplication node

mobeenabdullah:
## Why

`/workflows/model/flux/` emits a `SoftwareApplication` entity node for the model family. `/ja/workflows/model/flux/` does not, so every localized model page is missing from that part of the entity graph.

Verified live before writing anything:

| | English | Japanese |
|---|---|---|
| `/workflows/model/flux/` | CollectionPage, BreadcrumbList, ItemList, ListItem, **SoftwareApplication, FAQPage, HowTo, HowToStep, Question, Answer** | |
| `/ja/workflows/model/flux/` | | CollectionPage, BreadcrumbList, ItemList, ListItem |

Named as an unshipped cleanup in [TDD: GTM-291](https://app.notion.com/p/3a56d73d365080beb7daf90ae3cdea79) section 8.4.

## Why only the entity node crosses over

The FAQ and how-to nodes, and the `featureList` on the SoftwareApplication, are all built from the model landing copy in `src/content/landing/models/`. Those eleven JSON files are **English only**; there are no per-locale variants and the collection loader has no locale dimension.

Emitting them on a localized page would put English questions and steps into structured data on a page that declares `inLanguage` for its own locale, offering search engines an answer in a language the reader did not ask for. That is worse than emitting nothing.

The name and description are safe because `buildModelPageMeta` already resolves them for the page's locale. Confirmed on the live Japanese page: `Flux Comfy Workflows — 無料テンプレート` with a Japanese meta description.

## On `inLanguage`

The TDD pairs this with "`inLanguage` finally matching actual content language". That half is already done: locale pages emit `inLanguage` on their `CollectionPage` node and it is correct. `structured-data.ts` is deliberately untouched here (see below).

## Verification

6 tests on the rule, not on the output shape. Two were proven able to fail:

- Passing the English highlights through as a `featureList` turns the featureList guard red.
- Dropping the thin-content gate turns the quality guard red.

One test asserts the shared builder *does* emit a `featureList` when given one, so the omission reads as a deliberate choice rather than passing because the builder ignores the field.

Gate: `npm run lint`, `npx astro check` (0 errors), `npx vitest run` (698 tests, 55 files).

## Coordination

Scoped to the localized route on purpose.

- **#1171** (approved, merging today) appends to `structured-data.ts` immediately after `buildSoftwareApplicationJsonLd`, with hunk context covering those lines. This PR imports that function and does not edit the file, so there is no overlap.
- **#1164** has hunks straddling the English model route's import block and is waiting on a reviewer. Refactoring the English route now would hand that PR an avoidable conflict, so the English route adopts the builder once #1164 lands. That follow-up removes the duplication that let these two routes drift apart in the first place.

### DIFF
--- a/site/src/pages/[locale]/workflows/model/[name].astro
+++ b/site/src/pages/[locale]/workflows/model/[name].astro
@@ -28,6 +28,7 @@ import {
   buildWorkflowBreadcrumb,
   serializeJsonLdForScript,
 } from '../../../../lib/structured-data';
+import { buildLocalizedModelJsonLd } from '../../../../lib/workflow-pages/model-page-jsonld';
 import { listingIndexing, alternatesFor } from '../../../../lib/i18n/listing-indexing';
 import { localizeCards } from '../../../../lib/i18n/localize-cards';
 
@@ -86,7 +87,7 @@ const canonicalUrl = absoluteUrl(indexing.canonicalPath);
 // resolve through the same helper as the English route so the hreflang cluster
 // stays consistent.
 const content = await loadModelContent(family.slug);
-const { h1, title, description } = buildModelPageMeta({
+const { h1, title, description, hasQualityContent } = buildModelPageMeta({
   group: family,
   content,
   locale: typedLocale,
@@ -120,6 +121,14 @@ const breadcrumbStructuredData = buildWorkflowBreadcrumb(typedLocale, { name: h1
   },
   { name: t('labels.models', typedLocale), item: absoluteUrl(modelsIndexPath(typedLocale)) },
 ]);
+
+// The English route emits a SoftwareApplication entity node for the family and
+// this one did not, so the localized pages were absent from that part of the
+// entity graph. See the builder for why the FAQ and how-to nodes stay English-only.
+const jsonLdBlocks = [
+  breadcrumbStructuredData,
+  ...buildLocalizedModelJsonLd({ h1, description, hasQualityContent }),
+];
 
 // matchingTemplates are already SerializedTemplate[] with profile data
 const serialized = matchingTemplates;
@@ -167,11 +176,11 @@ const serialized = matchingTemplates;
     />
   </div>
 
-  <script
-    is:inline
-    type="application/ld+json"
-    set:html={serializeJsonLdForScript(breadcrumbStructuredData)}
-  />
+  {
+    jsonLdBlocks.map((data) => (
+      <script is:inline type="application/ld+json" set:html={serializeJsonLdForScript(data)} />
+    ))
+  }
 
   <Fragment slot="footer">
     <SiteFooter />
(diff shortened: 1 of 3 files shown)


site/src/pages/[locale]/workflows/model/[name].astro
```diff
@@ -121,6 +122,14 @@ const breadcrumbStructuredData = buildWorkflowBreadcrumb(typedLocale, { name: h1
   { name: t('labels.models', typedLocale), item: absoluteUrl(modelsIndexPath(typedLocale)) },
 ]);
 
+// The English route emits a SoftwareApplication entity node for the family and
+// this one did not, so the localized pages were absent from that part of the
+// entity graph. See the builder for why the FAQ and how-to nodes stay English-only.
+const jsonLdBlocks = [
```
Comment (dante01yoon) [inline site/src/pages/[locale]/workflows/model/[name].astro:128]: suggestion: The unit tests cover the builder, but nothing covers this wiring — if the destructured `hasQualityContent` field were renamed or the spread dropped in a refactor, every localized page would silently lose the node again with all six tests still green, which is the exact drift-shaped failure this PR fixes. #1199 just made AstroContainer render tests possible in this repo; a single render of this route (or a small extracted `jsonLdBlocks` helper with one test) asserting the page HTML contains a `SoftwareApplication` script would close the loop. Non-blocking, fine as a follow-up — possibly the same follow-up where the English route adopts the builder.

dante01yoon (approved):
Approving. Keeping the FAQ/HowTo nodes and `featureList` off the localized pages is the right call, not a shortcut — the landing copy is English-only, and structured data claiming answers in a language the page doesn't speak is worse than the gap being fixed. The test that proves the shared builder *does* emit `featureList` when given one is exactly the guard that keeps the omission a decision rather than an accident. I verified the wiring end to end: `hasQualityContent` is a real field on `buildModelPageMeta`'s return (`seo-page.ts:33,73`), so the gate can't silently collapse to `undefined`, and `buildSoftwareApplicationJsonLd` exists on main with the expected optional-`featureList` shape. The deferred English-route adoption of the builder (to avoid handing #1164 a conflict) is reasonable sequencing — please do land that follow-up, since the duplication is what let the two routes drift apart originally.

One non-blocking suggestion inline.

Validation: read the full diff at `0ff44821` plus the head versions of `seo-page.ts` and `structured-data.ts`; all CI checks green on this head.

site/src/pages/[locale]/workflows/model/[name].astro
Comment (mobeenabdullah) [reply site/src/pages/[locale]/workflows/model/[name].astro:128]: Agreed, and taking it as the follow-up you describe rather than folding it in here.

Two reasons for not doing it in this PR. #1199 has only just merged, so this branch would need a rebase to pick up the container capability, on an approved and clean PR. And this route is prerender = false with getStaticPaths, content collections and a hub-api load, so a full render is the fragile end of the options rather than the cheap one.

You are right that it is the uncovered half. Worth noting the narrow version of the risk: dropping or renaming the destructured hasQualityContent is a TypeScript error, since it becomes an undeclared identifier. What nothing catches today is the spread being removed from jsonLdBlocks, or the array no longer being rendered.

So the follow-up is the one you suggested: the English route adopts the builder once #1164 lands, and both routes get a render assertion that the page HTML carries a SoftwareApplication script. Recording it against that work.

Issue #39, opened 2025-04-17
Update development docs


christian-byrne:
Now that it is a separate pip package, it will be much easier to test changes locally. The README should be updated to reflect the new process for testing local changes.

PR #1208, opened 2026-08-28
fix(site): serve the MiniMax H3 demo media from media.comfy.org

nav-tej:
comfy.org/workflows/minimax-h3-multiref renders with all three reference keyframes broken, an empty result player, and a dead agent-prompt fetch. Root cause: the page (and the demo card on /workflows) reference the committed public/ assets with root-relative URLs, but behind the comfy.org router those resolve against the website app, which has no /demos tree — all four asset requests 404 in production. (Icons like /icons/node-left.svg only work because copies exist in the website repo's public dir; /icons/caret-right.svg has no copy and 404s too — fixed separately in ComfyUI_frontend.)

Fix: the five demo assets are uploaded to media.comfy.org (gs://comfy-org-videos/website/demos/mmh3/..., all verified HTTP 200, access-control-allow-origin: * so the on-click prompt fetch works cross-origin), and both pages now reference the absolute CDN URLs. Source files stay committed under public/demos/mmh3.

Follow-up worth considering: hub-only root-relative public assets break on comfy.org whenever there is no website-side copy — either the comfy-router should forward the hub's static prefixes, or the hub should adopt an asset-base convention.

### DIFF
--- a/site/src/pages/workflows/index.astro
+++ b/site/src/pages/workflows/index.astro
@@ -89,8 +89,8 @@ const facetsConfig = buildFacetsConfig(locale);
              page's critical path; the frame paints before any video bytes. -->
         <video
           id="mmh3-promo-video"
-          src="/demos/mmh3/example/example.mp4"
-          poster="/demos/mmh3/example/keyframes/kf_1.webp"
+          src="https://media.comfy.org/website/demos/mmh3/example/example.mp4"
+          poster="https://media.comfy.org/website/demos/mmh3/example/keyframes/kf_1.webp"
           autoplay
           muted
           loop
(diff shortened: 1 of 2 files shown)


mobeenabdullah (approved):
Verified against production. All five CDN assets resolve with CORS:

```
example.mp4    200   access-control-allow-origin: *
kf_1.webp      200   access-control-allow-origin: *
kf_2.webp      200   access-control-allow-origin: *
kf_3.webp      200   access-control-allow-origin: *
agent-prompt.md 200  access-control-allow-origin: *
```

The diagnosis in the body is right: root-relative paths resolve against the website app, which has no /demos tree. Checked for anything this missed and the only other root-relative asset in the demo is /icons/plus.svg, which returns 200 because a copy already exists in the website repo. So the page is visually complete after this.

One thing to be clear about, since this is linked in the thread as the fix for FE-1932: it does not fix that 404 and cannot. Same root cause, but the failing call is a server route, not an asset:

```
/workflows/minimax-h3-multiref/            200  x-served-by: vercel-workflows
/api/workflows/minimax-h3-multiref/queue   404  x-served-by: vercel-website
```

comfy-router has no /api/ rule at all, so Run Workflow lands on the website app. A CDN cannot stand in for a POST endpoint. Fixed in Comfy-Org/comfy-router#39, which needs a review and a worker deploy.

Merging this on its own merit. Worth knowing it leaves the demo looking correct while Run still fails, so FE-1932 stays open until #39 deploys.

PR #1214, opened 2026-08-29
Gate the MiniMax H3 demo behind an experiment flag, with an integration check that switches it off

christian-byrne:

---

Follow-up on [FE-1932](https://linear.app/comfyorg/issue/FE-1932). Adds the integration check that would have caught the outage, and a flag that hides the demo cleanly instead of deleting it.

## Current live status of the 404

**Still broken, and comfy-router#39 is the only reason.** The router fix is merged (`9d52ea1`) but its deploy job is `workflow_dispatch`-only and has never been run — production is still on `ea7007a`.

| Target | Result |
|---|---|
| `comfy.org/api/workflows/minimax-h3-multiref/queue` | **404**, `x-served-by: vercel-website` |
| `workflow-templates.vercel.app/...` (same app, direct) | **200** |

The `x-served-by` header is the proof: the request is still reaching the marketing origin. Everything else is healthy — I submitted a real job through the Vercel origin end to end (submit → poll → cancel, all 200), and the CDN reference images from #1208 all serve 200.

**A second bug is queued up behind the 404.** Astro's CSRF guard compares the browser's `Origin` against its own computed origin. Behind the router those differ, so the demo's multipart POST is rejected:

```
POST .../run  Origin: https://workflow-templates.vercel.app  -> 200 {"jobId":...}
POST .../run  Origin: https://comfy.org                      -> 403 Cross-site POST form submissions are forbidden
```

Deploying the router alone would have turned the 404 into a 403. This PR fixes that with `security.allowedDomains` (hostname-only — adding `protocol` makes Astro *drop* `X-Forwarded-Proto` and fall back to `http`, which still fails). Verified against Astro's own validator, including that `comfy.org.evil.com` is still rejected.

> **This one hop is unproven end-to-end** and can't be until the router ships: whether Vercel's edge forwards comfy-router's `X-Forwarded-Host` unmodified is only observable in production. When the router deploys, the first red run will still say *404* — that's the router, not this fix. Confirm with one POST before turning the flag on.

## 1–2. The integration test, and what it says right now

`site/tests/integration/minimax-h3-demo.test.ts` drives the same four calls the page makes, against a real deployment: queue readout → submit three references → poll → cancel. It sends a **browser-shaped** request — real `Origin` header, real image bytes fetched from the CDN, `multipart/form-data` — because each of those is what one of the three known regressions turned on. It cancels the job in `afterAll`, so it costs seconds of GPU, not a full render.

```
$ pnpm test:integration                      # -> comfy.org
  ✓ serves the demo page
  ✓ serves the reference images the page uploads
  × answers the queue readout as JSON
  × accepts a three-reference submission
  Tests  2 failed | 2 passed | 2 skipped

  AssertionError: /queue returned 404. The request did not reach the hub app.
  The edge router (comfy-router) forwards /workflows/* but must also forward
  /api/workflows/* to the workflows origin — this is the FE-1932 failure exactly.

$ DEMO_BASE_URL=https://workflow-templates.vercel.app pnpm test:integration
  Tests  6 passed (6)     # 8.6s, real job submitted and cancelled
```

It is **red on production and green on the origin behind it** — which is the isolation proof. Kept out of `pnpm test`, which stays hermetic and offline.

## 3 + 5. CI: switch the feature off, don't block the release

`.github/workflows/minimax-demo-integration.yml` runs every 6h, after each site deploy, and on demand. On failure it flips the flag off, commits, and posts to Slack using the `chat.postMessage` + `jq -e '.ok == true'` pattern from ComfyUI_frontend. **It never fails the job** — the demo's backend is private beta behind a pre-alpha router, and a red check there must not hold up unrelated template releases. "Blocking the deploy" happens by shipping the feature *off*, not by blocking the pipeline.

Recovery is **not** automatic: a green run only reports that the demo *could* go back on. Alert-fatigue matters here — "healthy while switched off" is the shipped steady state, so recovery is only announced on manual dispatches; scheduled runs put it in the step summary.

Needs `secrets.SLACK_BOT_TOKEN` and `vars.SLACK_HUB_ALERTS_CHANNEL_ID` added to this repo. Both are optional at runtime — the step warns and skips, so the flag flip still happens.

**Deliberately not included:** the flip does not trigger a production deploy itself. It reaches users on the next site build, and the Slack message says so with a one-click link to run *Deploy Template Site*. Letting an automated health check deploy the worker in front of all of comfy.org is exactly the blast radius this thread has been careful about — that should be someone's explicit call.

## 4. The flag, and the static fallback

`site/src/data/experiment-flags.json` is repo-owned and read at **build time**, so the disabled branch is absent from the emitted HTML rather than hidden by client JS — crawlers and users get identical bytes. Verified on real build output:

| | flag off | flag on |
|---|---|---|
| `/workflows/index.html` contains the promo | **0 occurrences** | 1 |
| demo page ships the `MiniMaxH3Demo` island | **0** | 1 |
| `<meta name="robots">` | `noindex` | `noindex` |

The demo page is also **no longer `prerender = false`** — nothing in its frontmatter needs a request, so it's now statically generated (the API routes keep theirs).

`noindex` is held constant in **both** states and deliberately *not* wired to the flag: the flag flips unattended on backend downtime, and indexability that flaps with uptime teaches crawlers the URL is unreliable — worse than never indexing it.

Off state serves a fallback shaped like a normal template page, with CTAs that go somewhere that works (screenshots below).

## 6. SEO / conversion review

The demo page is `noindex`, so it earns **no ranking at all** — the thing it was partly there to do. The real cost is on `/workflows`, which *is* indexed:

- The promo sits between the hero and the search grid. Measured on live comfy.org at 1440×1000: it pushes the **search bar to y=1261 — below the fold**. With it gated, search sits at **y=861, above the fold**.
- Its "Try MiniMax H3" button is a **second solid-brand CTA**, matching the tracked Cloud CTA in visual weight, competing for the same click.
- It carries **no `utm_*` and no `data-*`** — I confirmed on the live page. PostHog runs `autocapture: false` and tracks only by delegation on specific hooks, so **every click into the demo is invisible in the funnel.** We can't even measure the diversion.

That last one is fixed here: internal experiment entry points now fire `hub:experiment_cta_clicked`, the fallback's outbound Cloud CTA carries `run-cloud-btn`, and unit tests assert the hooks are on the elements.

**Note:** PR #1209, which was meant to remove this promo, was never merged — the promo is still live today. This PR supersedes it by gating rather than deleting.

## 7. Policy doc

`site/docs/shipping-experiments-policy.md` — four rules and a checklist for putting unpolished surfaces in front of traffic: don't take a primary CTA slot, don't cannibalize keywords or nav targets, instrument it or you're guessing, ship it behind a flag with a health check that flips it. Placed here because `site/` already owns the hub's `seo-setup-guide.md` and `ab-testing-guide.md`, and there is no dedicated marketing-docs repo (`docs/` is the user-facing product site).

It records that this promo **knowingly violates its own Rule 1** and must satisfy it before the flag goes back on — rather than shipping a rule whose worked example fails it.

## Verification

- `pnpm test` — 779 passing (59 files), incl. 4 new
- `pnpm lint` — clean; `pnpm run check` — 0 errors, 0 warnings
- `prettier --check` — clean on all touched `site/` files (`format:check` fails identically on `main`; pre-existing)
- Real `astro build` in both flag states, output greped; both pages driven in a browser
- Integration suite run against both origins; reason-extraction checked against 404, `fetch failed`, and `TimeoutError`
- Four rounds of `deep-reviewer`; final verdict PASS/ship. Fixed along the way: the `protocol` key that silently defeated the CSRF fix, a recovery alert that would have posted 4×/day forever, a kill switch that failed silently, and an override guard whose test asserted a property nothing enforced.

**The flag ships off**, so merging this hides the demo immediately — which answers the "should we just take this down" question without deleting the work.

## Screenshots

The demo page with the flag off: static fallback with preview video, reference thumbnails, and Open Comfy Cloud / Browse all workflows CTAs

The /workflows hub index with the promo gated off - search and browse now sit in the first viewport


christian-byrne:
TL;DR: lens pass found this review-ready on code evidence, with no new blocking code finding from the current diff surface.
Current blockers are `REVIEW_REQUIRED` and external proof that the router/deploy path is actually serving the gated demo path. The skipped E2E/Lighthouse/visual jobs are path-filter skips, so they should not be counted as browser-journey evidence for the demo.

<details><summary>Full context for agent readers</summary>

Checked current PR state at head `e5da30c06eb3e6b7bc9bd7d6372c17838946fde4`: open, non-draft, mergeable, `mergeStateStatus=BLOCKED`, `reviewDecision=REVIEW_REQUIRED`. The diff carries permanent coverage in `site/tests/integration/minimax-h3-demo.test.ts`, `site/tests/unit/experiment-flags.test.ts`, and `site/tests/unit/posthog.test.ts`, plus the new `minimax-demo-integration.yml` workflow, so the tests-as-permanent requirement is satisfied for the flag/demo behavior in the PR itself.

Readiness gap: the PR body says the live 404 still depends on the router/deploy path, and the current checks show `E2E Tests`, `Visual Regression`, and Lighthouse variants skipped. Before treating this as merge-ready, the PR should have either owner review plus a fresh router/deploy receipt in the thread, or an explicit note that this PR intentionally ships only the off-by-default guard while router production proof is tracked elsewhere.
</details>

site/tests/integration/minimax-h3-demo.test.ts
```diff
@@ -0,0 +1,234 @@
...
+ * a preview:
+ *
+ *   DEMO_BASE_URL=https://workflow-templates.vercel.app pnpm test:integration
+ */
+import { afterAll, describe, expect, it } from 'vitest';
+
+import { exampleKeyframeUrl } from '../../src/lib/demos/mmh3/config';
+
+const BASE_URL = (process.env.DEMO_BASE_URL ?? 'https://comfy.org').replace(/\/+$/, '');
+const API = `${BASE_URL}/api/workflows/minimax-h3-multiref`;
```
mobeenabdullah: Dead after the rebase. #1213 moved these routes to `/workflows/api/minimax-h3-multiref`. This base 404s on comfy.org and on workflow-templates.vercel.app, and nothing maps the old path (no `redirects` or `rewrites` key in either `vercel.json`).

This file is not in the conflict set, so the rebase will not flag it. The 404 branch of `diagnose()` also needs revisiting once the base moves, since it hardcodes the FE-1932 router explanation.

.github/workflows/minimax-demo-integration.yml
```diff
@@ -0,0 +1,289 @@
...
+
+          commit_flip
+
+          # The integration test takes minutes, so main can move underneath this
+          # checkout and reject the push. Retrying matters: the whole point of the
+          # kill switch is that a broken demo goes off without waiting for a
+          # human, and losing this race would leave it live until someone acted
+          # on the Slack ping.
+          for attempt in 1 2 3; do
+            if git push origin HEAD:main; then
```
mobeenabdullah: This push is rejected by ruleset 9388637 on `main` (`pull_request`, 1 approval, `bypass_actors` empty). It is the rule, not a race, so the retry loop below cannot recover it.

`i18n-update-hub.yml`, `sync-template-index.yml` and `generate-upload-json.yml` all push a branch and open a PR instead. Same shape would work here.

mobeenabdullah (commented):
Two blockers.

Context first, because it causes the first one: #1213 and #1215 both merged while this was open, so the 404 and the 403 this PR diagnoses are already fixed in production. I checked today: `POST /workflows/api/minimax-h3-multiref/run` returns 400 with `Origin: comfy.org` and 403 with `Origin: evil.com`.

## 1. The integration check points at a path that no longer exists

#1213 moved the routes to `/workflows/api/...`. The test still builds its base as `/api/workflows/...`, which is a hard 404 on both origins, with no redirect and no rewrite in either `vercel.json`. The merged tree contains only `site/src/pages/workflows/api/minimax-h3-multiref/*`, so after the rebase this check can never pass anywhere.

Two knock-ons worth knowing:

- The test file is not in the conflict set, so the rebase will not surface this for you.
- The failure prints the FE-1932 router diagnosis, so the recorded reason will be wrong.

The app itself is fine. The merge takes main's `MiniMaxH3Demo.vue`, which already calls the new paths. It is only the test.

## 2. The kill switch cannot push to main

`git push origin HEAD:main` is rejected by ruleset 9388637 on `main`: `pull_request` rule, `required_approving_review_count: 1`, `bypass_actors` empty, `current_user_can_bypass: never`. That is the rule rather than a race, so the `git reset --hard origin/main` retry loop below cannot recover it. 25 of the last 25 commits on main went through a PR.

Every other automation in this repo already works around it. `i18n-update-hub.yml` pushes a branch and opens a PR; `sync-template-index.yml` and `generate-upload-json.yml` push to a PR head. None pushes to main.

So as written the flip never lands and "switches itself off" degrades to a Slack message asking a human to do it. It fails loudly rather than silently, so it is not dangerous, but it never does its main job. Fix is to open a PR like the siblings, or add a bypass actor to the ruleset.

deepme987:
Drive-by from reviewing #1216 (now merged). Not a request to change this PR's scope — one line of it, if you agree.

**The promo video is now the single largest thing on `/workflows/`, and the flag does not fix that when it is on.**

`MiniMaxPromo.astro` keeps `autoplay` alongside `preload="metadata"`. Those two do not compose: `autoplay` overrides the preload hint and the browser fetches the whole body. It is the exact defect #1216 was about — its commit message puts it as "autoplay was silently overriding preload".

Measured on live `https://comfy.org/workflows/` after the #1216 deploy:

| | bytes |
| --- | ---: |
| promo video body | **3,402,799** |
| everything else that must load on first paint | 2,128,969 |

That is **61.5%** of the must-load subtotal (HTML + CSS + JS + carousel slide-1 video + 6 carousel posters + the promo poster), and ~53% of the ~6.2 MB the page weighs in total — the figure #1216 quotes. Of the 33 `<video>` elements on the page, only two now autoplay: the carousel's slide 1, which is the deliberate above-the-fold LCP hero, and this one.

And it is **below the fold**. Stacking the heights — navbar ~72 px, carousel `h-[clamp(300px,60vw,520px)]` = 520 px at desktop widths, `HubHero` ≈ 205 px — puts the promo's top edge near 810 px. On a 1366×768 laptop it is entirely out of view. It is never the LCP element, so nothing is gained by fetching it eagerly.

**Suggested change, entirely inside `MiniMaxPromo.astro`:** drop `autoplay`, set `preload="none"`, and widen the existing reduced-motion script into the visibility gate the repo now uses in `site/src/components/workflow-pages/Thumbnail.astro` (lines 67–89) — `IntersectionObserver` at `rootMargin: '200px'`, `play().catch(() => {})` on intersect when motion is not reduced, `pause()` otherwise, with the current `play()` behaviour as the no-observer fallback. The reduced-motion branch folds into that condition rather than sitting beside it, so it is a replacement, not an addition. Keep `poster`, `muted`, `loop`, `playsinline` exactly as they are — the poster is 34.5 KB and it is what makes the frame paint before any video bytes.

One thing that will not work: adding `data-gated-video` and relying on `Thumbnail.astro`'s script. Astro bundles that script only into pages that render `Thumbnail.astro`, and that component is used exclusively by the `workflow-pages/cards/*` components, none of which appear on `/workflows/`. It needs its own inline script.

Roughly +12/−8 lines. Takes `/workflows/` from ~6.2 MB to ~2.8 MB whenever the flag is on, against the 7 MiB `total-byte-weight` error budget #1216 just set for this URL in `site/lighthouserc.json`.

Separately, and out of scope here: the asset lives at `media.comfy.org/website/demos/…`, outside the `hub-media` manifest, so #1216's VMAF ≥ 95 re-encode pass never saw it. 3.4 MB is large next to the re-encoded hub clips, whose median is well under 1.5 MB. Worth a pass at some point.

<sub>Also a small correction for the record: #1216's description credits #1208 with adding this. #1208 only rewrote the two URLs to `media.comfy.org`; the block came in with #1177 (`71f43419`) with the same attributes.</sub>

PR #1217, opened 2026-09-01
fix(site): stop advertising hreflang=it, an unbuilt route that 404s

christian-byrne:
## Summary

Six English `/workflows/*` listing pages advertised `<link rel="alternate" hreflang="it" href="https://comfy.org/it/workflows/...">`, and `/it/workflows/...` 404s — confirmed live on `/workflows/`, `/workflows/creators/`, and `/workflows/comfyui/` before this fix. Flagged by nav in [Comfy-Org/ComfyUI_frontend#16247](https://github.com/Comfy-Org/ComfyUI_frontend/pull/16247) as out of scope for that PR ("the workflows app... declares `it`, which 404s. Separate PR.").

## Root cause

`HreflangTags.astro`'s default (`clusterLocales(localized)` with no explicit `locales` list) resolves to every `SUPPORTED_HUB_LOCALES` entry — the locales the hub *builds routes for* — not `INDEXABLE_LOCALES`, the reviewed subset that actually has a live `/{locale}/...` page. `locales.ts`'s own doc comment describes this narrowing explicitly:

```
AVAILABLE_APP_LOCALES  ⊇  SUPPORTED_HUB_LOCALES  ⊇  INDEXABLE_LOCALES
```

`it` sits in the gap: it's in `LANGUAGES` (so `SUPPORTED_HUB_LOCALES` includes it — the hub "deliberately builds and serves" it as a route shape) but has never cleared native review, so it's absent from `INDEXABLE_LOCALES` and no `/it/...` page is ever actually built.

Six English routes render `<BaseLayout>` without passing an explicit `hreflangLocales`, so they silently fall through to the broad default: `workflows/index.astro`, `workflows/category/[type].astro`, `workflows/tag/[tag].astro`, `workflows/model/[name].astro` (which explicitly set `hreflangLocalized={true}`, same default), `workflows/creators.astro`, and `workflows/[username].astro`. Their localized `[locale]/workflows/*` siblings already avoid this — they compute `hreflangLocales` from `listingIndexing()`/`alternatesFor()`, which read `INDEXABLE_LOCALES` — except `[locale]/workflows/[username].astro`, which had the same gap and is also fixed here.

## Fix

Each affected page now passes an explicit `hreflangLocales={[DEFAULT_LOCALE, ...INDEXABLE_LOCALES]}` (or, for the model route, `alternatesFor()`'s reconciliation with the page's own `noindex`, matching its localized sibling's pattern exactly). No change to `HreflangTags.astro`, `SEOHead.astro`, `BaseLayout.astro`, or `clusterLocales`'s default — those are flagged "DO NOT remove/modify without care" in CLAUDE.md, and the existing `seohead-render.test.ts`/`i18n-og-locale.test.ts` tests assert that broad default as correct at the component level. The bug is specific callers not overriding it, not the shared default.

## Guard test

`workflows-english-hreflang-coverage.test.ts`: a static source-scan over every English `/workflows/*` route, same pattern as the existing `i18n-localized-routes-coverage.test.ts`. Flags any route rendering `<BaseLayout>` without `hreflangLocales=` or `hreflangLocalized={false}`, and separately flags any route whose `hreflangLocales` is derived from `LANGUAGES`/`LOCALES`/`SUPPORTED_HUB_LOCALES`/`AVAILABLE_APP_LOCALES` instead of `INDEXABLE_LOCALES`. Verified the guard actually catches the regression: it fails, naming the exact file, when the fix to `workflows/index.astro` is reverted.

## Not in this PR

- `minimax-h3-multiref.astro` is unconditionally `noindex`, so its hreflang cluster is moot for crawling/indexing; it's a one-off interactive demo, not part of the hub's localization surface. Left untouched to keep this change scoped to indexed pages.
- A pre-existing, unrelated `prettier-plugin-astro` parse failure on `workflows/[username].astro` (reproduces on `main` before this change, confirmed via `git stash`) is not fixed here — separate issue, separate file concern, not something I introduced.

## Test plan

- `pnpm test` — 812/812 passing (3 new)
- `pnpm run lint` — clean
- `pnpm run check` (astro check) — 0 errors (was 7 type errors before typing `hreflangLocales` as `Locale[]`; the spread `[DEFAULT_LOCALE, ...INDEXABLE_LOCALES]` widens to `string[]` without an explicit annotation)
- Confirmed live 404 on `/it/workflows/` and `/it/workflows/comfyui/` before this fix (both still 404 after — this only removes the false hreflang advertisement, it doesn't build the `it` route, which is the correct fix per nav's report: the route is genuinely not ready, the tag was the lie)

### DIFF
--- a/site/src/pages/[locale]/workflows/[username].astro
+++ b/site/src/pages/[locale]/workflows/[username].astro
@@ -28,6 +28,7 @@ import {
 } from '../../../lib/hub-api';
 import { fetchRankingMap } from '../../../lib/ranking';
 import { localizeCards } from '../../../lib/i18n/localize-cards';
+import { INDEXABLE_LOCALES } from '../../../lib/i18n/locales';
 import { creatorOgUrl } from '../../../lib/og-url';
 import { buildToolbarLabels } from '../../../lib/toolbar';
 import { getSocialDisplay } from '../../../lib/social-links';
@@ -162,6 +163,10 @@ if (!profileFound && serialized.length === 0) {
 }
 
 const canonicalUrl = absoluteUrl(`/workflows/${slug}/`);
+// Same content/cluster as the English creator-profile page; advertise only
+// the locales actually flipped on, not BaseLayout's full-LANGUAGES default
+// (which includes "it", unbuilt and 404ing).
+const hreflangLocales: Locale[] = [DEFAULT_LOCALE, ...INDEXABLE_LOCALES];
 const socialLinks = websiteUrls.map(getSocialDisplay);
 
 const structuredData = {
@@ -180,6 +185,8 @@ const structuredData = {
   ogImage={creatorOgUrl(displayName, slug!, avatarUrl || undefined)}
   structuredData={structuredData}
   locale={typedLocale}
+  hreflangBasePath={`/workflows/${slug}/`}
+  hreflangLocales={hreflangLocales}
   theme="dark"
   hideDefaultFooter
 >
--- a/site/src/pages/workflows/[username].astro
+++ b/site/src/pages/workflows/[username].astro
@@ -19,12 +19,13 @@ import WorkflowGrid from '../../components/hub/WorkflowGrid.vue';
 import { buildToolbarLabels } from '../../lib/toolbar';
 import { getLocaleFromPath } from '../../i18n/utils';
 import { absoluteUrl } from '../../config/site';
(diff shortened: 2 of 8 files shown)


christian-byrne:
Fixed in 4c0f0acd — the guard now also catches a bare `hreflangLocales = SUPPORTED_HUB_LOCALES`-style direct assignment, not just the spread/Object.keys forms.

PR #1241, opened 2026-09-05
Add new templates and reorganize template categories

comfyui-wiki:


comfyui-wiki:
@coderabbitai recheck

PR #1269, opened 2026-09-16
feat(templates): add ModelScope sources for MiniMax Music 3

xingchen-mu:
## Summary
- add explicit ModelScope download URLs for the MiniMax Music 3 diffusion model, text encoder, and VAE
- update both the user-facing template and blueprint metadata

## Why
The frontend missing-model source selector consumes explicit provider-specific URLs. ModelScope paths cannot be safely derived from Hugging Face URLs, so the template must declare them.

## Validation
- both files parse as valid JSON
- all three ModelScope URLs return HTTP 200


xingchen-mu:
@coderabbitai review

The generated blueprint and manifest artifacts are now synchronized in commit `2a7b672`, and the previous review thread has been resolved. Please re-review the current head.

comfyui-wiki:
Thanks. Let's hold this until the frontend side lands.

[ComfyUI_frontend#17825](https://github.com/Comfy-Org/ComfyUI_frontend/pull/17825), which consumes `sources` and adds `modelscope.cn` to the download allowlist, is still open. Until it lands, nothing on `main` reads this field, and we may want changes here depending on how that one shapes up.

Source selection is the frontend team's call, so let's revisit this once #17825 is decided.

